Issue #157, opened 2020-01-29
Developer Admin UI shall have always a "Are you sure ?" confirmation


de-jcup:
As long as we have no dedicated WebUI (see #11 ) we use the Developer Admin UI for operations. But using this developer tool at production... at least a "Are you sure ?" confirmation for any action shall appear before executed. 

There shall be also the possibility to define a system property `sechub.developertools.admin.confim` (like done for `sechub.developertools.admin.environment` ). If this property is `false` than no confirmation shall be shown! (To help developers to reduce this step on non-productive environments were it doesn't mateter)

Classes of interest:
- `DialogUI` 
- `AbstractUIAction`

de-jcup:
@Jeeppler please use/create a feature branch as described at https://github.com/Daimler/sechub/wiki/Collaboration

de-jcup:
Hmm.. may idea is too too much generic and we should simply go over each action and add an confirmation explicit there - reduces "magic effect" to developers looking at the action and wondering why the confirmation always appears...

Also in some cases it just not make sense - e.g. an action like "list users" needs no confirmation.

@Jeeppler 
Please just do following
- look into every action where we do an operation which does change system behaviour or updates data (so changes...)
- in every of those "changing" actions add a confirmation + provide cancel 
- there shall be no special system property any longer preventing the operation for developers - so we have exact same behaviour as for everybody else

PR #200, opened 2020-03-03
Feature 157 developer admin ui are you sure

Jeeppler:
Added extra confirmation dialogs to all update, create and delete actions in the developer UI, which did not previously had it.


de-jcup:
closes #157

PR #229, opened 2020-05-14
Feature 226 scheduler triggers only when not overloaded

de-jcup:


sechub-shared-kernel/src/main/java/com/daimler/sechub/sharedkernel/util/SimpleByteUtil.java
```diff
@@ -0,0 +1,22 @@
+package com.daimler.sechub.sharedkernel.util;
+
+import java.text.CharacterIterator;
+import java.text.StringCharacterIterator;
+
+public class SimpleByteUtil {
+
+    public static String humanReadableBytesLength(long bytes) {
+        long absB = bytes == Long.MIN_VALUE ? Long.MAX_VALUE : Math.abs(bytes);
```
de-jcup, line 9: will be renamed

de-jcup:
Comments with spdx headers must be added as well

Issue #256, opened 2020-06-16
SecHub client codescan default whitelist


de-jcup:
To avoid having binary content, we will provide a default whitelist containing all wellknown/accepted file endings

So we do:

- define directories to scan code
- apply default filter for includes (*.java, *.python , *.cpp ...)
- apply custom include filters
- apply custom exclude filters

We adopt the ant-pattern style for our filter format (see https://ant.apache.org/manual/dirtasks.html)

---
<sup>Albert Tregnaghi <albert.tregnaghi@daimler.com>, Daimler TSS GmbH, [imprint](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)</sup>

de-jcup:
Suggestion for wellknown/accepted file endings:

- *.java, *.groovy, *.build
- *.cpp

de-jcup:
We should also implement a "--ignore-filters" so every file can be uploaded in case of a missing file association.

PR #262, opened 2020-06-18
Client: implementation of default exclude patterns #227 #260

### LINKED ISSUE #227: Implement reasonable default exclude patterns
Code Scan: Enable lazy usage of client:
If no excludes are defined in the sechub.json config file, then use reasonable defaults.
( see https://daimler.github.io/sechub/latest/sechub-client.html#example-upload-folders-with-excludes )

Benefits:
- save bandwidth and disk space on server
- speed up scan

Open discussion:
Which exclude patterns should be added (source code scan)?
```
- Archives: **/*.7z **/*.bz2 **/*.gz **/*.rar **/*.tar **/*.tgz **/*.zip
- Binary image files: **/*.bin **/*.cue **/*.iso **/*.mdf **/*.vcd
- Build artifacts: **/*.o
- Documents: **/*.doc **/*.docx **/*.eps **/*.odc **/*.odp **/*.odt **/*.rtf **/*.pdf **/*.pps **/*.ppt **/*.pptx **/*.ps **/*.svg **/*.xls **/*.xlsx
- Executables: **/*.com **/*.exe
- Fonts: *.fnt *.otf *.ttf
- Images: **/*.bmp **/*.gif **/*.ico **/*.jpeg **/*.jpg **/*.png **/*.svg **/*.tif **/*.tiff **/*.xcf
- Libraries: **/*.a **/*.dll **/*.jar **/*.so
- OS packages: **/*.apk **/*.app **/*.cab **/*.deb **/*.rpm **/*.dmg
- Sound/Videos: **/*.aif **/*.avi **/*.flac **/*.flv **/*.gifv **/*.ogg **/*.ogv **/*.mid **/*.mkv **/*.mov **/*.mp2 **/*.mp3 **/*.mp4 **/*.mpa **/*.mpeg **/*.mpg **/*.vob **/*.wav **/*.wma **/*.wmv **/*.webm
- Text files: **/*.md **/*.adoc **/*.asciidoc **/*.log **/*.plantuml **/*.puml **/*.tex **/*.txt **/*.latex **/*.vcf
```

### LINKED ISSUE #260: Client - Exclude pattern **/*.xyz does not work
An exclude pattern like **/*.xyz should match *.xyz files in all subdirectories,
Bug: But **/*.txt does not exclude e.g. "/x/y/z/V/b.txt" so these files would be added to the zip file even if they should'nt.

---

sven-dmlr:
- implemented default exclude directory patterns
- fixed documentation's json example code (bracket was missing), example for directories, because files will be covered with #256
- re-implementation of func Filepathmatch via regexp matching which is more simple and also fixes #260
- added tests for above

closes #227 
closes #260 

---

### DIFF
--- a/sechub-cli/src/daimler.com/sechub/cli/constants.go
+++ b/sechub-cli/src/daimler.com/sechub/cli/constants.go
@@ -1,9 +1,17 @@
 // SPDX-License-Identifier: MIT
+
 package cli
 
 // DefaultSecHubConfigFile represents the name of the sechub configuration file
 // used per default when no other set
 const DefaultSecHubConfigFile = "sechub.json"
+
+// DefaultZipExcludeDirPatterns - Define directory patterns to exclude from zip file:
+// code in directories named "test" is not considered to end up in the binary
+// also ignore .git directory
+var DefaultZipExcludeDirPatterns = []string{"**/test/**", "**/.git/**"}
+
+// Definition as var because a constant needs a fix array size.
 
 /* ---------------------------------- */
 /* -------- Exit codes -------------- */
--- a/sechub-cli/src/daimler.com/sechub/util/filepathmatcher.go
+++ b/sechub-cli/src/daimler.com/sechub/util/filepathmatcher.go
@@ -3,91 +3,38 @@ package util
 
 import (
 	"fmt"
-	"path/filepath"
+	"regexp"
 	"strings"
 )
 
-//
-// This method provides ANT like selectors.
+// Filepathmatch - This method provides ANT like selectors.
+// See https://ant.apache.org/manual/dirtasks.html
 // For example: "**/a*.txt" will accept
-//               - "/home/tester/xyz/a1234.txt
-//               - "/root/a1b.txt
+//               - "/home/tester/xyz/a1234.txt"
+//               - "a1b.txt"
 //
 func Filepathmatch(path string, pattern string) (result bool) {
-	/* ------------------------------ */
-	/* Explanation - how it works ... */
-	/* ------------------------------ */
-	// - we separate pattern into different path patterns by path wildcards into elements
-	// - then the last element is inspected to be a file wildcard
-	// - file wildcards are resolved /tried by golang filepath.Match method
-	// - if this is okay we do our special path wildcard matching
 
+	// Let's turn the ant style pattern into a regexp:
+	doublestarPatterns := strings.Split(pattern, "**/")
+	//fmt.Printf("\nbefore: %q path: %q", pattern, path)
+	for i, subElement := range doublestarPatterns {
+		// esacpe . with backslash, * to .*
+		doublestarPatterns[i] = strings.Replace(strings.Replace(subElement, ".", "\\.", -1), "*", ".*", -1)
+	}
+	regexpPattern := "^" + strings.Join(doublestarPatterns, ".*/") + "$"
 
-	/* separate patterns containing path wildcards */
-	subPattern := strings.Split(pattern, "**/")
-//	fmt.Printf("\n%s - seperated into %q", pattern, subPattern)
-	length := len(subPattern)
-	potentialFilePattern := subPattern[length-1]
-
-	/* ------------------------ */
-	/* handle file name pattern */
-	/* ------------------------ */
-	substring := path
-	if strings.Index(potentialFilePattern, "*") != -1 {
-		/* found a file pattern in last entry */
-		/* change length - so slice not changed but last entry no longer handled later */
-		length--
-
-		/* check if file patter is applicable */
-		lastPathIndex := strings.LastIndex(path, "/")
-		fileName := path
-		indexForCut := lastPathIndex
-		if lastPathIndex != -1 {
-			fileName = path[lastPathIndex:]
-		} else {
-			indexForCut = 0
-		}
-		/* we just use the golang integrated filename matcher:*/
-		matches, err := filepath.Match(potentialFilePattern, fileName)
-		if err != nil {
-			fmt.Println(err)
-		}
-		if !matches {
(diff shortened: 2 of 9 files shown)


sechub-cli/src/daimler.com/sechub/cli/constants.go
```diff
@@ -1,10 +1,18 @@
...
 package cli
 
 // DefaultSecHubConfigFile represents the name of the sechub configuration file
 // used per default when no other set
 const DefaultSecHubConfigFile = "sechub.json"
 
+// DefaultZipExcludeDirPatterns - Define directory patterns to exclude from zip file:
+// code in directories named "test" is not considered to end up in the binary
+// also ignore .git directory
+var DefaultZipExcludeDirPatterns = []string{"**/test/**", "**/.git/**"}
```
Jeeppler, line 12: Why not use `const` here?

sechub-cli/src/daimler.com/sechub/util/filepathmatcher.go
```diff
@@ -3,91 +3,38 @@ package util
...
-		matches, err := filepath.Match(potentialFilePattern, fileName)
-		if err != nil {
-			fmt.Println(err)
-		}
-		if !matches {
-			return false
-		}
+	// Let's turn the ant style pattern into a regexp:
+	doublestarPatterns := strings.Split(pattern, "**/")
+	//fmt.Printf("\nbefore: %q path: %q", pattern, path)
```
Jeeppler, line 20: I think that `fmt.Printf` can be removed.

sechub-cli/src/daimler.com/sechub/util/filepathmatcher.go
```diff
@@ -3,91 +3,38 @@ package util
...
-			/* found but not starting with it*/
-			/* when last pattern was not an asterisk ...*/
-			return false
-		}
-		substring = substring[index+len(currentSearch):]
-		/* after first entry we have always path asterisks between - even
-		 * when we got no longer "" between (happens only on first entry)
-		 */
-		lastOneWasAsterisk=true
+	//fmt.Printf("\n after: %q path: %q\n", regexpPattern, path)
```
Jeeppler, line 33: I think the `fmt.Printf` is obsolete.

sechub-cli/src/daimler.com/sechub/cli/constants.go
sven-dmlr: I have tried that already. `const` for []string ist not allowed.
go-lint says: _"const initializer []string literal is not a constant. No quick fixes available"_

This type is needed or things like
`append(json.CodeScan.Excludes, DefaultZipExcludeDirPatterns...)`
will not work. (both need to be of type []string)

sechub-cli/src/daimler.com/sechub/util/filepathmatcher.go
sven-dmlr: Yes - it could be removed since it's only a comment.

But - I found similar in the existing code and for the development it was very helpful and convenient, just to uncomment to get the values while testing.
So I just continued that "tradition" because I find it helpful.
I think, it won't harm if we keep the line. :-)

sechub-cli/src/daimler.com/sechub/util/filepathmatcher.go
sven-dmlr: same as for above...

sechub-cli/src/daimler.com/sechub/util/filepathmatcher.go
de-jcup: Please don't keep dead code inside your git repo. Let us create a "go snippet wiki page" or something to hold such snippets and remove it from the code base. When a commented line is inside code when commit - okay, this can happen. But it should be accidently and removed after recognition.

sechub-cli/src/daimler.com/sechub/util/filepathmatcher.go
de-jcup: Sames as above from my side. Please remove...

sechub-cli/src/daimler.com/sechub/util/filepathmatcher_test.go
```diff
@@ -43,6 +44,18 @@ func Test_When_double_asterisk_on_start_any_path_is_accepted_when_filename_witho
 	AssertFalse(Filepathmatch("/x/y/z/V/a.txt", "**/b.txt"), t)
 }
 
+func Test_When_double_asterisk_on_start_and_filename_with_asterisk(t *testing.T) {
+	AssertFalse(Filepathmatch("/x/y/z/V/a.txtx", "**/*.txt"), t)
+	AssertTrue(Filepathmatch("/x/y/z/V/b.txt", "**/*.txt"), t)
+}
+
+func Test_When_double_asterisk_also_matching_files_in_current_working_directory(t *testing.T) {
```
de-jcup, line 52: Nice! I was just starting to write a comment to add some tests about pathes but you did it already. Good work 👍

sechub-cli/src/daimler.com/sechub/util/filepathmatcher.go
```diff
@@ -3,91 +3,38 @@ package util
...
-
-
-	/* separate patterns containing path wildcards */
-	subPattern := strings.Split(pattern, "**/")
-//	fmt.Printf("\n%s - seperated into %q", pattern, subPattern)
-	length := len(subPattern)
-	potentialFilePattern := subPattern[length-1]
 
-	/* ------------------------ */
-	/* handle file name pattern */
```
de-jcup, line 33: @sven-dmlr : Thank you for removing my old code - your solution is shorter, better and works also in deep 👍

PR #269, opened 2020-06-25
Feature 206 sechub analyzer cli

Jeeppler:
SecHub Analyzer CLI is command line tool, which searches for SecHub markers in files. The markers are:

- Start: `NOSECHUB`
- End: `END-NOSECHUB`

The Analyzer CLI markers are programming language agnostic. 

---
<sup>Jeremias Eppler <jeremias.eppler@daimler.com>, Daimler TSS GmbH, [imprint](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)</sup>


Jeeppler:
Implements: #206

sechub-analyzer-cli/src/main/java/com/daimler/sechub/analyzer/core/FileAnalyzer.java
```diff
@@ -0,0 +1,91 @@
...
+            // iterate through every line in the file
+            while ((line = bufferedReader.readLine()) != null) {
+                lineNumber++;
+
+                // search for the NOSECHUB marker
+                int noSecHub = line.indexOf(NOSECHUB);
+
+                if (noSecHub > -1) {
+                    // search for NOSECHB_END marker
+                    int endNoSecHub = line.indexOf(NOSECHUB_END);
```
de-jcup, line 61: Hmm. could you please define
`noSecHub` as `noSecHubIndex` and 
`endNoSecHub` as `endNoSecHubIndex` ?

So it's clear at reading time that `noSecHub` is no a boolean...

sechub-analyzer-cli/src/main/java/com/daimler/sechub/analyzer/core/FileAnalyzer.java
```diff
@@ -0,0 +1,91 @@
...
+
+import com.daimler.analyzer.model.Marker;
+import com.daimler.analyzer.model.MarkerPair;
+import com.daimler.analyzer.model.MarkerType;
+
+/**
+ * Searches through a file looking for SecHub markers
+ */
+public class FileAnalyzer {
+    private static final String NOSECHUB = "NOSECHUB";
```
de-jcup, line 19: What If there is a text containing just a 
"NOSECHUB"  somewhere but not being a comment - e.g. a "// if this is not a real finding add NOSECHUB there"
in combination with a real finding?

For example:


```java

public void potentialEvilCommentOut(){
     // if this is not a real finding add NOSECHUB there!
      doProbablyStuffWhichCouldLeadToFalsePositive();
}
public void myEvilMethod(){
     doRealEvilStuffLikeSQLWithParamStringConcatenation();
}

public void myNotSoEvilMethod(){
     // NOSECHUB
     doingNotEvilStuffButFalsePositive();
     // NOSECHUB-END
}
```

In the case before and your current implementation potentialEvilCommentOut() method would lead to start - next start will be ignored and end from myNotSoEvilMethod() used. So code inside myEvilMethod() will be inside start-end block!

I would prefer a solution where we do explicit check for comment
"# NOSECHUB"
"// NOSECHUB"
"/* NOSECHUB\w" (I think here checking only start + whitespace is enough)
"/*     NOSECHUB\w"
"/*     NOSECHUB\w"
etc.

I think a fast approach is to check for index as done before and if there is a match use (one time compiled) regular expression patterns to remove whitespaces.

de-jcup: please add real source code files like NoSechHubExample1.java to our /src/test/resources folder - so it can be loaded and you can code really. Than play around with multiple blocks inside same file and some odd situations as mentioned before and load this in your tests (like done already but with text files - so you can use dedicated editors, etc. )

sechub-analyzer-cli/src/main/java/com/daimler/sechub/analyzer/core/Processor.java
```diff
@@ -0,0 +1,117 @@
...
+     * 
+     * @param rootFiles
+     * @return all files with markers
+     */
+    protected Map<String, List<MarkerPair>> analyzeFiles(List<File> rootFiles) {
+        Map<String, List<MarkerPair>> result = new HashMap<>();
+
+        Set<File> files = new HashSet<>();
+        
+        // create a list of all files
```
de-jcup, line 67: I prefer block comments for comments - most times I use //  either for parts to remove, or when being a "sub comment"

sechub-analyzer-cli/src/test/java/com/daimler/sechub/analyzer/core/ProcessorTest.java
```diff
@@ -0,0 +1,273 @@
...
+
+/**
+ * Integration Tests
+ * 
+ * Tests the Processor and FileAnalyzer classes
+ */
+public class ProcessorTest {
+    
+    final String path = "src/test/resources/";
+    private Processor processor = null;
```
de-jcup, line 32: avoid setting fields to null - is always done in java automatically and reduces readability

sechub-analyzer-cli/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+Please report issues at https://github.com/daimler/sechub
+~~~
+
+##### Example
+
+~~~
+$ java -jar analyzer-0.0.0.jar -p example/
+{
+  "findings" : {
```
de-jcup, line 75: I would prefer "noSecHubMarkers"
instead of "findings". So we can reuse general class "Markers" for other parts and its also more clear what is meant by the json content.

sechub-analyzer-cli/src/main/java/com/daimler/analyzer/model/AnalyzerResult.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+public class AnalyzerResult {
+
+    private Map<String, List<MarkerPair>> findings;
+
+    public AnalyzerResult(Map<String, List<MarkerPair>> findings) {
+        this.findings = findings;
+    }
+    
+    public Map<String, List<MarkerPair>> getFindings() {
```
de-jcup, line 18: Please rename to `getNoSecHubMarkers()`

de-jcup: As long as we do not provide this as an API it's okay to provide this map directly over the method (I have done this as well at some sechub locations). But we must be aware when this is used at different locations this can cause trouble - reason: The map is not immutable. but directly returned, so callers can change the content!

sechub-analyzer-cli/src/main/java/com/daimler/sechub/analyzer/cli/Analyzer.java
```diff
@@ -0,0 +1,137 @@
+package com.daimler.sechub.analyzer.cli;
+
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.util.List;
+
+import org.apache.commons.cli.CommandLine;
```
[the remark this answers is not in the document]
Jeeppler: I actually did not use Log4j 1.x, but used instead Log4j 2. Log4j 2 separates the implementation from the API, therefore it would be possible to use the Log4j 2 API with a different implementation. 

Log4j 2 incorporated the good parts from SLF4J. Furthermore, it made it possible to log objects instead of only strings:

https://stackoverflow.com/questions/41498021/is-it-worth-to-use-slf4j-with-log4j2

sechub-analyzer-cli/src/test/java/com/daimler/sechub/analyzer/core/CommentCheckerCommentInLineTest.java
```diff
@@ -0,0 +1,389 @@
...
+import java.util.List;
+
+@RunWith(Parameterized.class)
+public class CommentCheckerCommentInLineTest {
+    private static final String NOSECHUB = "NOSECHUB";
+    private static final String NOSECHUB_END = "END-NOSECHUB";
+    
+    private CommentChecker commentChecker;
+    
+    @Parameters
```
de-jcup, line 24: Very nice!

I am always was to lazy to use parameters - but here it makes absolute sense. Nice work!

de-jcup:

---

PR #376, opened 2020-09-25
Feature 324 simple java client library

de-jcup:


Jeeppler (commented):
Really nice. Especially, having OpenAPI v3 support and separating the SecHub report model.

PR #393, opened 2020-10-02
Update sechub.json for sechub itself

### LINKED ISSUE #340: Update sechub.json for sechub itself
## Situation
Inside our root folder we have got a `sechub.json` config file .

We want following sub project NOT to be scanned:

- sechub-developertools
- sechub-doc
- sechub-integrationtest
- sechub-other
- sechub-test
- sechub-testframework
- sechub-shared-kernel-testframework
- sechub-scan-testframework

also test source folders must be also out of scope.

## Problem
We have defined folders explicit to scan. Now some additional sub projects where added (e.g. sechub-pds, sechub-analyzer-cli, ...) and those folders will be not scanned at the moment

## Solution
The best way would be to use default excludes of client (so no test folders should be there).
+ a way to get all (*) but to exclude formerly mentioned folders and their children.

sven-dmlr:
- change from explicitely naming directories
- to instead cover all and define excludes

closes #340

---

### DIFF
--- a/sechub.json
+++ b/sechub.json
@@ -3,25 +3,25 @@
 
      "codeScan": {
         "fileSystem": {
-            "folders": ["sechub-adapter/src/main/java",
-                        "sechub-adapter-checkmarx/src/main/java",
-                        "sechub-adapter-nessus/src/main/java",
-                        "sechub-adapter-netsparker/src/main/java",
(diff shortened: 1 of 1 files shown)


de-jcup:

---

PR #395, opened 2020-10-05
Feature 387 actuator prometheus

### LINKED ISSUE #387: Spring actuator for prometheus
We want to integrate spring actuator for prometheus.

Preconditions
- is configurable. : spring boot properties are able to define a dedicated port for actuators (for security) or we define sechub properties and do
  it inside code
- shutdown hook will be disabled

4w31ss3:
Activates prometheus actuator for more sophisticated monitoring.

Closes #387
---

### DIFF
--- a/sechub-server-core/build.gradle
+++ b/sechub-server-core/build.gradle
@@ -5,6 +5,7 @@ dependencies {
     compile(library.flyway)
     
     implementation 'org.springframework.boot:spring-boot-starter-actuator'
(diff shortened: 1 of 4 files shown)


de-jcup (commented):
@4w31ss3 :
In future: please do always update library.gradle file and reuse the library variable instead of directly defining the dependency inside build.gradle files

For an example look at: compile(library.flyway)

Reason: So we got always the same version in all subprojects - easier to maintain.
Having prometheus only in server core and no reuse at the moment we can let it as is.


---

Issue #143, opened 2020-01-17
Add security integration tests for REST API (role access checks)


de-jcup:
## Situation
Some REST API parts are already tested by integration tests to be only accessible by dedicated role.
But not all.

## Solution
There should be a simple test for an existing scenario (e.g. scenario3 where users have acess to project) where just url calls are checked. 

urlsAccessibleAnonymous - must not be checked
urlsAccessibleByAdmins - must be checked that valid user has no access, and also no role owner
urlsAccessibleByOwner - must be checked that validUser having access to project is still not able to access this url

Just simple test that a rest call to a url denies the access.

### Idea
Maybe this could be solved generic - We use already Reflections library to check all REST parts are documented. In same way we could create a list of URLs for every documented part. Most parts are directly identifiable by URL name pattern, but some could need a special treatment.

This would automatically test access security for every new (and well documented) REST API part.

PR #401, opened 2020-10-12
#143: testing all routes for sufficient assigned roles

4w31ss3:
Closes https://github.com/Daimler/sechub/issues/143

---

### DIFF
--- a/sechub-test/src/test/java/com/daimler/sechub/test/RoutesTest.java
+++ b/sechub-test/src/test/java/com/daimler/sechub/test/RoutesTest.java
@@ -1,0 +1,174 @@
+package com.daimler.sechub.test;
+
+import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
+import java.util.stream.Collectors;
+import javax.annotation.security.RolesAllowed;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.springframework.beans.factory.config.BeanDefinition;
+import org.springframework.context.annotation.ClassPathScanningCandidateComponentProvider;
+import org.springframework.core.annotation.AnnotationUtils;
+import org.springframework.core.type.filter.AnnotationTypeFilter;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+import com.daimler.sechub.sharedkernel.RoleConstants;
+
+import static org.junit.Assert.*;
+
+@SuppressWarnings("rawtypes")
+public class RoutesTest {
+	
+	private Map<Class, Map<String, ArrayList<String>>> classRoutesRolesMap = new HashMap<Class, Map<String, ArrayList<String>>>();
+	
+	@Before
+	public void before() {
+		List<Class> allClasses = getAllClassesForAnnotationType(RestController.class);
+		
+		for (Class clazz : allClasses) {
+			getClassUserRolePaths(clazz);
+		}		
+	}
+	
+	@Test
+	public void routes_assigned_to_right_roles() {
+			
+		for (Entry<Class, Map<String, ArrayList<String>>> entry : classRoutesRolesMap.entrySet()) {
+			
+			String clazzName = entry.getKey().getName();
+			Map<String, ArrayList<String>> routesEntries = entry.getValue();
+			
+			for (Entry<String, ArrayList<String>> routesEntry : routesEntries.entrySet()) {
+				String route = routesEntry.getKey();
+				String[] roles = routesEntry.getValue().toArray(new String[routesEntry.getValue().size()]);
+				
+				if (!(route.contains("anonymous") || route.contains("actuator"))) {
+					assertTrue(clazzName + " with route " + route + " should contain roles.", roles.length > 0);
+				}
+				
+				if (route.contains("anonymous") || route.contains("actuator")) {
+					assertTrue(clazzName + " with route " + route + " should not contain roles.", roles.length == 0);
+				}
+				
+				if (route.contains("/api/admin")) {
+					assertTrue(clazzName + " with route " + route + " should only contain one role, ROLE_SUPERADMIN.", roles.length == 1);
+					assertTrue(clazzName + " with route " + route + " should contain ROLE_SUPERADMIN.", Arrays.asList(roles).contains(RoleConstants.ROLE_SUPERADMIN));
+				}
+				
+				if (route.contains("/api/project") || route.contains("/api/job")) {
+					assertTrue(clazzName + " with route " + route + " should contain ROLE_USER.", Arrays.asList(roles).contains(RoleConstants.ROLE_USER));
+					assertTrue(clazzName + " with route " + route + " should contain ROLE_SUPERADMIN.", Arrays.asList(roles).contains(RoleConstants.ROLE_SUPERADMIN));
+				}
+			}	
+		}
+	}
+	
+	private List<Class> getAllClassesForAnnotationType(Class clazz) {
+		ClassPathScanningCandidateComponentProvider scanner = new ClassPathScanningCandidateComponentProvider(false);
+		scanner.addIncludeFilter(new AnnotationTypeFilter(clazz));
+		
+		Set<BeanDefinition> allRestControllerBeans = scanner.findCandidateComponents("com.daimler.sechub");
+		
+		return allRestControllerBeans.stream()
+				.map( bean -> {
+					try {
+						return Class.forName(bean.getBeanClassName());
+					} catch (Exception e) {
+						e.printStackTrace();
+					}
+					return null;
+				})
+				.collect(Collectors.toList());
+	}
+	
+	private void getClassUserRolePaths(Class clazz) {
+		RequestMapping classMappings = AnnotationUtils.findAnnotation(clazz, RequestMapping.class);
+		
+		ArrayList<String> userRoles = new ArrayList<String>();
+		List<String> resultPaths = new ArrayList<String>();
+		
+		String basePath = "";
+		
+		if (classMappings != null) {
+			String[] paths = classMappings.path();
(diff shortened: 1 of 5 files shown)


sechub-test/src/test/java/com/daimler/sechub/test/RoutesTest.java
```diff
@@ -0,0 +1,174 @@
...
+	
+	@Test
+	public void routes_assigned_to_right_roles() {
+			
+		for (Entry<Class, Map<String, ArrayList<String>>> entry : classRoutesRolesMap.entrySet()) {
+			
+			String clazzName = entry.getKey().getName();
+			Map<String, ArrayList<String>> routesEntries = entry.getValue();
+			
+			for (Entry<String, ArrayList<String>> routesEntry : routesEntries.entrySet()) {
```
de-jcup, line 49: I know, checkstyle does prefer entries when it comes to sets.

But I am more a "fan" of 
```
for (String route: routesEntries.keySet()){
  ...
}
```
Reason: readability - maybe entries are some nanos faster, but the readability is much better

sechub-test/src/test/java/com/daimler/sechub/test/RoutesTest.java
```diff
@@ -0,0 +1,174 @@
...
+				String route = routesEntry.getKey();
+				String[] roles = routesEntry.getValue().toArray(new String[routesEntry.getValue().size()]);
+				
+				if (!(route.contains("anonymous") || route.contains("actuator"))) {
```
de-jcup, line 53: What about "/anonymous/" and "/actuator/" ?

If there is anybody designing a rest part containing those words in another context... yes I know a little bit fictional, but we should be aware

sechub-test/src/test/java/com/daimler/sechub/test/RoutesTest.java
```diff
@@ -0,0 +1,174 @@
...
+					assertTrue(clazzName + " with route " + route + " should contain roles.", roles.length > 0);
+				}
+				
+				if (route.contains("anonymous") || route.contains("actuator")) {
```
de-jcup, line 57: same as above

sechub-test/src/test/java/com/daimler/sechub/test/RoutesTest.java
```diff
@@ -0,0 +1,174 @@
...
+					assertTrue(clazzName + " with route " + route + " should not contain roles.", roles.length == 0);
+				}
+				
+				if (route.contains("/api/admin")) {
+					assertTrue(clazzName + " with route " + route + " should only contain one role, ROLE_SUPERADMIN.", roles.length == 1);
+					assertTrue(clazzName + " with route " + route + " should contain ROLE_SUPERADMIN.", Arrays.asList(roles).contains(RoleConstants.ROLE_SUPERADMIN));
+				}
+				
+				if (route.contains("/api/project") || route.contains("/api/job")) {
```
de-jcup, line 66: I would like  an else if construct here

- if is anonymous: check role defined as anonymous (I think the actuator check is only relevant for future, when provinding own actuators by REST methods, but I appreciate this)
- if integration test : just ignore
- else if admin route, check only admin
- else check at least role user must be set

So this will also check future api parts that at least user access must be defined

Also: an administrator has always user role inside - so I would just check for ROLE_USER in logic in last else part

sechub-test/src/test/java/com/daimler/sechub/test/RoutesTest.java
```diff
@@ -0,0 +1,174 @@
...
+		scanner.addIncludeFilter(new AnnotationTypeFilter(clazz));
+		
+		Set<BeanDefinition> allRestControllerBeans = scanner.findCandidateComponents("com.daimler.sechub");
+		
+		return allRestControllerBeans.stream()
+				.map( bean -> {
+					try {
+						return Class.forName(bean.getBeanClassName());
+					} catch (Exception e) {
+						e.printStackTrace();
```
de-jcup, line 85: Please use slf4j log instead e.printStacktrace

sechub-test/src/test/java/com/daimler/sechub/test/RoutesTest.java
```diff
@@ -0,0 +1,174 @@
...
+					}
+					return null;
```
de-jcup, line 87: I am not sure if it is okay to just return null - this would lead to null entries inside list result.
hmm.. I would appreciate a runtime exception here - maybe illegal state.

sechub-test/src/test/java/com/daimler/sechub/test/RoutesTest.java
```diff
@@ -0,0 +1,174 @@
...
+import org.springframework.web.bind.annotation.RestController;
+
+import com.daimler.sechub.sharedkernel.RoleConstants;
+
+import static org.junit.Assert.*;
+
+@SuppressWarnings("rawtypes")
+public class RoutesTest {
+	
+	private Map<Class, Map<String, ArrayList<String>>> classRoutesRolesMap = new HashMap<Class, Map<String, ArrayList<String>>>();
```
de-jcup, line 30: Please use Java8 style  instead:
```java
private Map<Class, Map<String, ArrayList<String>>> classRoutesRolesMap = new HashMap<>());
```

sechub-test/src/test/java/com/daimler/sechub/test/RoutesTest.java
```diff
@@ -0,0 +1,174 @@
...
+					}
+					return null;
+				})
+				.collect(Collectors.toList());
+	}
+	
+	private void getClassUserRolePaths(Class clazz) {
+		RequestMapping classMappings = AnnotationUtils.findAnnotation(clazz, RequestMapping.class);
+		
+		ArrayList<String> userRoles = new ArrayList<String>();
```
de-jcup, line 95: Please use List at the left side instead ArrayList

sechub-test/src/test/java/com/daimler/sechub/test/RoutesTest.java
```diff
@@ -0,0 +1,174 @@
...
+		List<String> resultPaths = new ArrayList<String>();
+		
+		String basePath = "";
+		
+		if (classMappings != null) {
```
de-jcup, line 100: I am a fan of "guard closes"  and dedicated methods.

So I would prefer to have two additional methods:

```java
private void getClassUserRolePathsFromClassMappings(/*... necessary params*/)
         RequestMapping classMappings = AnnotationUtils.findAnnotation(clazz, RequestMapping.class);
            if (classMappings == null) {
                  return
            }
            // impl...
}


private void getClassUserRolePathsFromMethods(Class<?> clazz /*, ... necessary params*/)
           Method[] methods = clazz.getMethods();
            if (methods == null) {
                  return
            }
            // impl...
}
```

sechub-test/src/test/java/com/daimler/sechub/test/RoutesTest.java
```diff
@@ -0,0 +1,174 @@
...
+		}
+		
+		Method[] methods = clazz.getMethods();
+		
+		for (Method method : methods) {
+			RequestMapping methodMapping = AnnotationUtils.findAnnotation(method, RequestMapping.class);
+			if (methodMapping != null) {
+				String[] paths = methodMapping.path();
+				if (paths != null) {
+					for (String path : paths) {
```
de-jcup, line 128: when using a 
```java
if (path == null){
   continue;
}
```
you would reduce indention level and it increase readaility.

sechub-test/src/test/java/com/daimler/sechub/test/RoutesTest.java
```diff
@@ -0,0 +1,174 @@
...
+						
+						String fullPath = basePath + path;
+						
+						resultPaths.add(basePath + path);
+						
+						String[] methodRoles = getRoles(method);
+						
+						if (classRoutesRolesMap.get(clazz) == null) {
```
de-jcup, line 136: With Java 8 you got a map.computIfAbsent method - maybe more readable when using this instead.

de-jcup (changes requested):

sechub-test/src/test/java/com/daimler/sechub/test/RoutesTest.java
4w31ss3: done

sechub-test/src/test/java/com/daimler/sechub/test/RoutesTest.java
4w31ss3: done

sechub-test/src/test/java/com/daimler/sechub/test/RoutesTest.java
4w31ss3: done

sechub-test/src/test/java/com/daimler/sechub/test/RoutesTest.java
4w31ss3: done

sechub-test/src/test/java/com/daimler/sechub/test/RoutesTest.java
4w31ss3: done

sechub-test/src/test/java/com/daimler/sechub/test/RoutesTest.java
4w31ss3: done

sechub-test/src/test/java/com/daimler/sechub/test/RoutesTest.java
4w31ss3: I thought about splitting this into separate methods too, but I believe, it would introduce a lot of complexity because the second "method"-Iterations need locally shared variables (see userRoles, resultPaths and basePath at top of getClassUserRolePaths).

Also the right "chaining" of iterations over classMappings and MethodsMappings couldn't be granted when splitting them up into separate methods.

de-jcup:

---

de-jcup (approved):

Issue #123, opened 2019-12-16
Admin can create new scan job without a project


Jeeppler:
## Description 

An administrator can start a scan without having to first create a project, assign an owner and an user to the project.

## How to reproduce the issue

Start the SecHub integration test server (`./gradlew startIntegrationTestServer`) and use the following `curl` command:

```
$ curl --insecure --user int-test_superadmin:int-test_superadmin-pwd 'https://localhost:8443/api/project/project1/job' --request POST \
>     -H 'Content-Type: application/json;charset=UTF-8' \
>     -d '{"codeScan":{"fileSystem":{"folders":["testproject1/src/main/java","testproject2/src/main/java"]}},"apiVersion":"1.0"}'
{"jobId":"f0840087-5229-422d-8b41-7bc959050308"}
```

The command will return a `jobId`, even though the project with the Id `project1` does not exist:

```
$ curl --insecure --user int-test_superadmin:int-test_superadmin-pwd 'https://localhost:8443/api/admin/projects' --request GET \
>     -H 'Content-Type: application/json'
[]
```

In case a normal user tries to create a new job without an existing project or being assigned to a project, the request will be denied and SecHub returns an error message:

```
$ curl --insecure --user int-test_onlyuser:int-test_onlyuser-pwd 'https://localhost:8443/api/project/myproject/job' --request POST \
>     -H 'Content-Type: application/json;charset=UTF-8' \
>     -d '{"codeScan":{"fileSystem":{"folders":["testproject1/src/main/java","testproject2/src/main/java"]}},"apiVersion":"1.0"}'
{"status":404,"error":"Not Found","message":"Project myproject does not exist, or you have no access."
...
```

## Expected behavior 

An administrator should not be able to create a new scan without first creating a project, assigning an owner and an user to the project.

PR #405, opened 2020-10-13
#400: checking for existing project before allowing access to super-a…

### LINKED ISSUE #400: Admin users can scan not existing projects
When SecHub admin users start a scan, there is no check if the requested project exists.
Even worse: the whole scan is done including creation of the project in third-party tools.
(In this case, a Checkmarx license was wasted)

-> Please check and implement that in any case the project must exist before scanning is started

To reproduce: (project "phantasy" does not exist on SecHub server)
```
> cat sechub.json
{
  "apiVersion" : "1.0" ,
  "server"   : "https://sechub.example.org",
  "project"  : "phantasy",
  "codeScan": {
    "fileSystem": {
      "folders": ["src"]
    }
  }
}

> export SECHUB_USERID=my-admin-user ; export SECHUB_APITOKEN=.......

> sechub scan
 _____           _   _       _     
/  ___|         | | | |     | |    
\ `--.  ___  ___| |_| |_   _| |__  
 `--. \/ _ \/ __|  _  | | | | '_ \ 
/\__/ /  __/ (__| | | | |_| | |_) |
\____/ \___|\___\_| |_/\__,_|_.__/ Client Version 0.21.0

- Creating new sechub job
Zipping folder: src (/home/myuser/tmp/sechub/demo/src)- Uploading source zip file
- Approve sechub job
- Waiting for job 960103be-c80e-4b56-88ad-e30c7fde22e2 to be done
  ....
- Fetching result (format=json) for job 960103be-c80e-4b56-88ad-e30c7fde22e2
- SecHub report written to sechub_report_phantasy_960103be-c80e-4b56-88ad-e30c7fde22e2.json
  GREEN - no security vulnerabilities identified
```

---

4w31ss3:
Closes #400 #123 

---

### DIFF
--- a/sechub-schedule/src/main/java/com/daimler/sechub/domain/schedule/access/ScheduleAccessRepository.java
+++ b/sechub-schedule/src/main/java/com/daimler/sechub/domain/schedule/access/ScheduleAccessRepository.java
@@ -6,19 +6,20 @@ import static com.daimler.sechub.domain.schedule.access.ScheduleAccess.*;
 import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.data.jpa.repository.Modifying;
 import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.repository.query.Param;
 
 import com.daimler.sechub.domain.schedule.access.ScheduleAccess.ProjectAccessCompositeKey;
 
 public interface ScheduleAccessRepository extends JpaRepository<ScheduleAccess, ProjectAccessCompositeKey> {
-
-	@Query(value="SELECT EXISTS(SELECT * FROM "+TABLE_NAME+" where "+COLUMN_PROJECT_ID+" = ?1)",nativeQuery=true)
-	public Boolean hasProjectUserAccess(String projectId);
+	
+	@Query(value="SELECT EXISTS(SELECT * FROM "+TABLE_NAME+" where "+COLUMN_PROJECT_ID+"=:projectId)",nativeQuery=true)
+	public Boolean hasProjectUserAccess(@Param("projectId") String projectId);
 	
 	@Modifying
-	@Query(value="DELETE FROM "+TABLE_NAME+" where "+COLUMN_USER_ID+" = ?1",nativeQuery=true)
-	public void deleteAccessForUserAtAll(String userId);
+	@Query(value="DELETE FROM "+TABLE_NAME+" where "+COLUMN_USER_ID+"=:userId",nativeQuery=true)
+	public void deleteAccessForUserAtAll(@Param("userId") String userId);
 
 	@Modifying
-	@Query(value="DELETE FROM "+TABLE_NAME+" where "+COLUMN_PROJECT_ID+" = ?1",nativeQuery=true)
-	public void deleteAnyAccessForProject(String projectId);
+	@Query(value="DELETE FROM "+TABLE_NAME+" where "+COLUMN_PROJECT_ID+"=:projectId",nativeQuery=true)
+	public void deleteAnyAccessForProject(@Param("projectId") String projectId);
 }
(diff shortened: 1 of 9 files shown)


sechub-schedule/src/main/java/com/daimler/sechub/domain/schedule/access/ScheduleAccessRepository.java
```diff
@@ -11,9 +11,12 @@
 
 public interface ScheduleAccessRepository extends JpaRepository<ScheduleAccess, ProjectAccessCompositeKey> {
 
+	@Query(value="SELECT EXISTS(SELECT * FROM "+TABLE_NAME+" where "+COLUMN_PROJECT_ID+" = ?1)",nativeQuery=true)
```
de-jcup, line 14: I know this is done also at other query inside this class, but in future please try always to use spring data annotation `@Param("myIdentifier") and use it inside the query by " =:myIdentifier"

Its okay for one parameter, but having multiple ones named parameters are much better because no ordering problems when query changes - its also more readable

sechub-schedule/src/main/java/com/daimler/sechub/domain/schedule/access/ScheduleAccessRepository.java
```diff
@@ -11,9 +11,12 @@
...
+	public Boolean hasProjectUserAccess(String projectId);
```
de-jcup, line 15: Please try to use simple data types where possible. (of course not when using BigInteger etc.) 

Reason: Boolean can be `null` - so having methods which returns a "Boolean" instead of a simple "boolean" do implicit bring up the question at caller side : can this be null? When callers do NOT think about the call and just do an
```if (isXYZExampleForBoolean()) ``` this does autoboxing and when Boolean result is null, then you got an unexpected NPE. 

Yes, in this case this should never happen.
I do not know if hibernate/JPA does internally use a boolean and you got boolean->Boolean->boolean conversion, but you always do a Boolean->boolean autoboxing, so the result can be defined as simple data type.

File: sechub-schedule/src/main/java/com/daimler/sechub/domain/schedule/access/ScheduleUserAccessToProjectValidationService.java [hunk 3]
```diff
@@ -33,8 +33,14 @@
 	 * @param projectId
 	 */
 	public void assertUserHasAccessToProject(String projectId) {
+				
 		if (userContextService.isSuperAdmin()) {
-			/* a super admin has always access */
+			/* a super admin has always access to existing projects */
+			
+			if(!accessRepository.hasProjectUserAccess(projectId)) {
+				throw new NotFoundException("Project " + projectId + " does not exist, or you have no access.");
```
de-jcup, line 41: This happens only when no project access with this project id for any user exists - at schedule domain there is no information available if the project does exist or not.

This happens only for super administerators.
So a message like "Project xyz does not exist, or you have no access"

I prefer:
"Project xyz does not exist, or no user has access at all"

de-jcup (changes requested):
Please do following changes
- look at my comments and make the mentioned smaller changes (or discuss)
- create missing `ScheduleUserAccessToProjectValidationServiceTestScheduleUserAccessToProjectValidationService (just a simple junit test with mockito) and write some tests at least check the new expected behaviour by mocking repository and other involved parts.
- write a new test inside `ScheduleAccessRepositoryDBTest` and check query method works as expected

sechub-schedule/src/main/java/com/daimler/sechub/domain/schedule/access/ScheduleAccessRepository.java
4w31ss3: done

sechub-schedule/src/main/java/com/daimler/sechub/domain/schedule/access/ScheduleUserAccessToProjectValidationService.java
4w31ss3: done

sechub-schedule/src/main/java/com/daimler/sechub/domain/schedule/access/ScheduleAccessRepository.java
4w31ss3: done

de-jcup (approved):

PR #502, opened 2021-01-15
Added audit logging for false positive actions, closes #501

### LINKED ISSUE #501: Integrate audit logging for false positive actions
## Situation
When a user has defined false positive definitions, those parts will be filtered, means not listed inside the report results any longer. But when a user asks an administrator why something is not showing up in a report - and in the mean time another user has removed the false positive entry, it's very difficult to trace/understand the report.

## Wanted 
As an administrator I want to be able to explain to a user why a finding inside a report is filtered by SecHub when the false positive definition has been deleted in the meanwhile.

## Solution
So we add audit logging for false positive actions to have audit information about false positive removals. The audit log shall not contain sensitive data but  just give some hints what happened.

We add audit log entries for situations
- when user adds false positive entries (we just count the amount)
- when user removes false positive entries (we log job uuid and finding id - so filtering behaviour can be traced)

---

de-jcup:
- added audit logging actions
- renamed some methods inside fp job data service so its more
  clear what happens in validation methods
- renamed junit test of fp job data service from
  FalsePositiveJobServiceTest to FalsePositiveJobDataServiceTest
- added mocked audit log service to junit test dependencies

---

### DIFF
--- a/sechub-scan/src/test/java/com/daimler/sechub/domain/scan/project/FalsePositiveJobDataServiceTest.java
+++ b/sechub-scan/src/test/java/com/daimler/sechub/domain/scan/project/FalsePositiveJobDataServiceTest.java
@@ -8,10 +8,11 @@ import org.junit.Before;
 import org.junit.Test;
 
 import com.daimler.sechub.domain.scan.ScanAssertService;
+import com.daimler.sechub.sharedkernel.logging.AuditLogService;
 import com.daimler.sechub.sharedkernel.validation.UserInputAssertion;
 import com.daimler.sechub.sharedkernel.validation.ValidationResult;
 
-public class FalsePositiveJobServiceTest {
+public class FalsePositiveJobDataServiceTest {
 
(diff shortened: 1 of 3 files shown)


sechub-scan/src/test/java/com/daimler/sechub/domain/scan/project/FalsePositiveJobDataServiceTest.java
```diff
@@ -29,11 +31,13 @@ public void before() {
         configService = mock(ScanProjectConfigService.class);
         userInputAssertion = mock(UserInputAssertion.class);
         scanAssertService = mock(ScanAssertService.class);
+        auditLogService =mock(AuditLogService.class);
```
Jeeppler, line 34: Please use the formatter.

Jeeppler (changes requested):
See comment.

Jeeppler (approved):

PR #565, opened 2021-02-24
Feature 386 sereco support for sarif

### LINKED ISSUE #386: Sereco support for SARIF
We will provide SARIF inside Sereco ( see https://sarifweb.azurewebsites.net/) because it is supported by OASIS and becomes a standard of SAST report formats.

It will also help us to simplyfy PDS integration very much! Every product supporting SARIF will be very easy to integration when PDS integration in SecHub is full done.

4w31ss3:
closes #386


sechub-sereco/src/main/java/com/daimler/sechub/sereco/importer/SarifV1JSONImporter.java
```diff
@@ -0,0 +1,181 @@
...
+import com.contrastsecurity.sarif.Result.Level;
+import com.daimler.sechub.commons.model.ScanType;
+import com.daimler.sechub.sereco.metadata.SerecoCodeCallStackElement;
+import com.daimler.sechub.sereco.metadata.SerecoMetaData;
+import com.daimler.sechub.sereco.metadata.SerecoSeverity;
+import com.daimler.sechub.sereco.metadata.SerecoVulnerability;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+@Component
+public class SarifV1JSONImporter extends AbstractProductResultImporter {
```
Jeeppler, line 26: Please rename the file and class from "SarifV1JSONImporter.java" to "SarifJSONImporterV1.java". The file name is ambiguous, because starting with "SarifV1" could mean the "JSON Importer" supports only "SARIF version 1" or the "JSON Importer" is in version one. Moving the version to the end clarifies, that the "JSON Importer" is in version 1.

sechub-sereco/src/main/java/com/daimler/sechub/sereco/importer/SarifV1JSONImporter.java
```diff
@@ -0,0 +1,181 @@
...
+        if (!taxa.isPresent()) {
+            return "";
+        }
+
+        return taxa.get().getId();
+    }
+    
+    private List<Result> resultsWithGroupedLocations(List<Result> results) {
+        
+        HashMap<Integer, Result> mappedResults = new HashMap<>();
```
Jeeppler, line 147: Please use the generic `Map` instead of `HashMap` for the `mappedResults`.

sechub-sereco/src/test/java/com/daimler/sechub/sereco/importer/SarifV1JSONImporterTest.java
```diff
@@ -0,0 +1,155 @@
...
+        sarifEmptyResult = SerecoTestFileSupport.INSTANCE.loadTestFile("sarif/sarif_2.1.0_empty_results.json");
+        sarifThreadflowsExample = SerecoTestFileSupport.INSTANCE.loadTestFile("sarif/sarif_2.1.0_threadflows_example.json");
+    }
+
+    @Test
+    public void sarifreport_can_be_imported() {
+        /* prepare */
+
+        ImportParameter paramBrakeman = ImportParameter.builder().importData(sarifBrakeman).importId("id1").productId("SARIF").build();
+        ImportParameter paramThreadFlows = ImportParameter.builder().importData(sarifThreadflowsExample).importId("id1").productId("SARIF").build();
```
Jeeppler, line 39: Please separate the test for `sarifBrakeman` and `sarifThreadflowsExample`. Those are different test cases.

sechub-sereco/src/test/java/com/daimler/sechub/sereco/importer/SarifV1JSONImporterTest.java
```diff
@@ -0,0 +1,155 @@
...
+    @Before
+    public void before() {
+        importerToTest = new SarifV1JSONImporter();
+        sarifBrakeman = SerecoTestFileSupport.INSTANCE.loadTestFile("sarif/sarif_2.1.0_brakeman.json");
+        sarifEmptyResult = SerecoTestFileSupport.INSTANCE.loadTestFile("sarif/sarif_2.1.0_empty_results.json");
+        sarifThreadflowsExample = SerecoTestFileSupport.INSTANCE.loadTestFile("sarif/sarif_2.1.0_threadflows_example.json");
+    }
+
+    @Test
+    public void sarifreport_can_be_imported() {
```
Jeeppler, line 35: `sarifreport` in one word is hard to read. I would prefer `sarif_report`.

sechub-sereco/src/test/java/com/daimler/sechub/sereco/importer/SarifV1JSONImporterTest.java
```diff
@@ -0,0 +1,155 @@
...
+
+        /* test */
+        assertEquals("Was able to import sarif!", ProductImportAbility.ABLE_TO_IMPORT, ableToImportBrakemanSarif);
+        assertEquals("Was able to import sarif!", ProductImportAbility.ABLE_TO_IMPORT, ableToImportThreadFlowSarif);
+    }
+
+    @Test
+    public void empty_sarifreport_throws_exception() {
+        
+        /* prepare */
```
Jeeppler, line 53: you can remove the `/* prepare */` comment if you do not use it.

sechub-sereco/src/test/java/com/daimler/sechub/sereco/importer/SarifV1JSONImporterTest.java
```diff
@@ -0,0 +1,155 @@
...
+        /* test */
```
Jeeppler, line 54: This should be `/* execute + test */` the `assertThrows` executes `importResult` and asserts that the `IOException` is thrown.

sechub-sereco/src/test/java/com/daimler/sechub/sereco/importer/SarifV1JSONImporterTest.java
```diff
@@ -0,0 +1,155 @@
...
+        assertThrows(IOException.class, () -> {
+            importerToTest.importResult(null);
+        });
+        
+        assertThrows(IOException.class, () -> {
+            importerToTest.importResult("");
```
Jeeppler, line 60: please separate the test for `null` and `empty`.

sechub-sereco/src/test/java/com/daimler/sechub/sereco/importer/SarifV1JSONImporterTest.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+    
+    @Test
+    public void sarifreport_has_errorlevel() throws Exception {
+        /* prepare */
+        SerecoMetaData result = importerToTest.importResult(sarifBrakeman);
+
+        /* execute */
+        List<SerecoVulnerability> vulnerabilities = result.getVulnerabilities();
+        SerecoVulnerability v1 = vulnerabilities.get(0);
```
Jeeppler, line 71: Please use `vulnerability1` instead of just `v1`.

sechub-sereco/src/test/java/com/daimler/sechub/sereco/importer/SarifV1JSONImporterTest.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    @Test
+    public void sarifreport_has_no_description() throws Exception {
+        /* prepare */
+        SerecoMetaData result = importerToTest.importResult(sarifBrakeman);
+
+        /* execute */
+        List<SerecoVulnerability> vulnerabilities = result.getVulnerabilities();
+        SerecoVulnerability v1 = vulnerabilities.get(0);
```
Jeeppler, line 84: Please use vulnerability1 instead of just v1.

sechub-sereco/src/test/java/com/daimler/sechub/sereco/importer/SarifV1JSONImporterTest.java
```diff
@@ -0,0 +1,155 @@
...
+    private String sarifBrakeman;
+    private String sarifEmptyResult;
+    private String sarifThreadflowsExample;
+    
+    @Before
+    public void before() {
+        importerToTest = new SarifV1JSONImporter();
+        sarifBrakeman = SerecoTestFileSupport.INSTANCE.loadTestFile("sarif/sarif_2.1.0_brakeman.json");
+        sarifEmptyResult = SerecoTestFileSupport.INSTANCE.loadTestFile("sarif/sarif_2.1.0_empty_results.json");
+        sarifThreadflowsExample = SerecoTestFileSupport.INSTANCE.loadTestFile("sarif/sarif_2.1.0_threadflows_example.json");
```
Jeeppler, line 31: You load this file, before every test. However, you only use this in one test `sarifreport_threadflow_locations`. Either use `@BeforeClass` and read the file only once or load the file in your test.

See: [@Before vs @BeforeClass vs @BeforeEach vs @BeforeAll](https://www.baeldung.com/junit-before-beforeclass-beforeeach-beforeall)

sechub-sereco/build.gradle
```diff
@@ -5,11 +5,11 @@
  * Root build file: "${rootProject.projectDir}/build.gradle"
  * ============================================================================
  */
+ 
 dependencies {
 
+	compile 'com.contrastsecurity:java-sarif:1.0'
```
Jeeppler, line 11: Please move the dependency name into `libraries.gradle` and use the key/name to reference it in compile.

Jeeppler (changes requested):
See comments.

In addition, the SPDX headers are missing. Some of the SARIF example files are not our property. Make sure you check the licenses of the original repos you have the files from.

sechub-sereco/src/main/java/com/daimler/sechub/sereco/importer/SarifV1JSONImporter.java
4w31ss3: image


This is matching the naming pattern of importers. I think, the V1 is because of the SecHub API V1.

I would like to avoid refactoring of all this classes and their appearances.

sechub-sereco/src/main/java/com/daimler/sechub/sereco/importer/SarifV1JSONImporter.java
4w31ss3: done

sechub-sereco/src/main/java/com/daimler/sechub/sereco/importer/SarifV1JSONImporter.java
Jeeppler: In this case, please add a comment explaining which SARIF versions are supported with a link to the SARIF standard. For example: https://docs.oasis-open.org/sarif/sarif/v2.1.0/sarif-v2.1.0.html

sechub-sereco/src/test/java/com/daimler/sechub/sereco/importer/SarifV1JSONImporterTest.java
4w31ss3: done

sechub-sereco/src/test/java/com/daimler/sechub/sereco/importer/SarifV1JSONImporterTest.java
4w31ss3: done

sechub-sereco/src/test/java/com/daimler/sechub/sereco/importer/SarifV1JSONImporterTest.java
4w31ss3: done

sechub-sereco/src/test/java/com/daimler/sechub/sereco/importer/SarifV1JSONImporterTest.java
4w31ss3: done

sechub-sereco/src/test/java/com/daimler/sechub/sereco/importer/SarifV1JSONImporterTest.java
4w31ss3: done

sechub-sereco/src/test/java/com/daimler/sechub/sereco/importer/SarifV1JSONImporterTest.java
4w31ss3: done

sechub-sereco/build.gradle
4w31ss3: done

sechub-sereco/src/main/java/com/daimler/sechub/sereco/importer/SarifV1JSONImporter.java
```diff
@@ -0,0 +1,187 @@
...
+        SerecoMetaData metaData = new SerecoMetaData();
+
+        sarif.getRuns().stream().forEach(run -> {
+            
+            List<Result> results = resultsWithGroupedLocations(run.getResults());
+            
+            results.stream().forEach(result -> {
+                SerecoVulnerability vulnerability = new SerecoVulnerability();
+
+                vulnerability.setDescription("");
```
Jeeppler, line 58: Why is the description always empty?

In SARIF the description can be empty. However, you should expect a result to have a [message object](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317459), [short description](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317552) or [full description](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317553).  All of those should have at least a `text` attribute.

The SARIF reports which Brakeman creates use the `message` property. The example results from the SARIF tutorial use the massage object as well: https://github.com/microsoft/sarif-tutorials/blob/main/docs/2-Basics.md#results.

Furthermore, the `text` property can have placeholders (see: [Messages with arguments](https://github.com/microsoft/sarif-tutorials/blob/main/docs/3-Beyond-basics.md#messages-with-arguments) and [Messages with placeholders](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317466)). In addition, the `text` property can contain hyperlinks formatted in markdown (see: [Messages with embedded links](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317467)).

You can ignore the placeholder arguments and links for now. The hyperlinks in markdown can easily be understood by users. I am not aware of any tool making use of the placeholder arguments for now. However, having a description of the vulnerability is important.

Jeeppler (changes requested):
See comments.

In addition, please run the `apply-headers` script to ensure the SPDX headers are set.

libraries.gradle
```diff
@@ -102,7 +102,9 @@ ext {
          google_re2j:                           "com.google.re2j:re2j:1.4",
          
          /* epages - see https://github.com/ePages-de/restdocs-api-spec */
-         epages_restdoc_api_spec_mockmvc:       "com.epages:restdocs-api-spec-mockmvc:${libraryVersion.restDocsApiSpec}"
+         epages_restdoc_api_spec_mockmvc:       "com.epages:restdocs-api-spec-mockmvc:${libraryVersion.restDocsApiSpec}",
+         
+         contrastsecurity_sarif:				"com.contrastsecurity:java-sarif:1.0"
```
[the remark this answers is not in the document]
4w31ss3: done

sechub-sereco/src/main/java/com/daimler/sechub/sereco/importer/SarifV1JSONImporter.java
```diff
@@ -0,0 +1,187 @@
...
+import com.daimler.sechub.sereco.metadata.SerecoMetaData;
+import com.daimler.sechub.sereco.metadata.SerecoSeverity;
+import com.daimler.sechub.sereco.metadata.SerecoVulnerability;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+
+/** 
+ * This Importer supports SARIF Version 2.1.0
+ * https://docs.oasis-open.org/sarif/sarif/v2.1.0/sarif-v2.1.0.html
+ *  */
```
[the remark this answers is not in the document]
4w31ss3: done

sechub-sereco/src/test/java/com/daimler/sechub/sereco/importer/SarifV1JSONImporterTest.java
```diff
@@ -0,0 +1,170 @@
...
+public class SarifV1JSONImporterTest {
+
+    private static SarifV1JSONImporter importerToTest;
+    private static String sarifBrakeman;
+    private static String sarifEmptyResult;
+    private static String sarifThreadflowsExample;
+    
+    @BeforeClass
+    public static void before() {
+        importerToTest = new SarifV1JSONImporter();
```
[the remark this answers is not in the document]
4w31ss3: Please have a look into `ImporterRegistry.java`. All importers are created and stored in a `importers` list. Therefore they are reused during runtime and this would match the behavior in this test in some way.

I believe side effects are no problem, or even would be of interest for us if they occur early within our UTs because this would hint to bigger problems with the actual implementation.

sechub-sereco/src/main/java/com/daimler/sechub/sereco/importer/SarifV1JSONImporter.java
4w31ss3: This is analog to the behavior of `CheckmarxV1XMLImporter.java`. No description is set there too. Because of the comment, I thought this is the desired behavior for some tools.

image


I think we need to assess all the tools we will be using to generate SARIF reports before we fill the description field in the future. Some tools could be exposing proprietary information which could be confusing or misleading for our SecHub users.

sechub-sereco/src/test/java/com/daimler/sechub/sereco/importer/SarifV1JSONImporterTest.java
```diff
@@ -0,0 +1,170 @@
...
+    public void sarif_report_has_no_description() throws Exception {
+        /* prepare */
+        SerecoMetaData result = importerToTest.importResult(sarifBrakeman);
+
+        /* execute */
+        List<SerecoVulnerability> vulnerabilities = result.getVulnerabilities();
+        SerecoVulnerability vulnerability = vulnerabilities.get(0);
+
+        /* test */
+        assertEquals("", vulnerability.getDescription());
```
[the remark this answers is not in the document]
4w31ss3: This test only aims to ensure and stress that the description field is empty, because I adapted the behavior of `CheckmarxV1XMLImporter.java`. If we would define an actual SARIF field to be matched to the description field this test would be obsolete.

sechub-sereco/src/main/java/com/daimler/sechub/sereco/importer/SarifV1JSONImporter.java
Jeeppler: @4w31ss3 I was assuming that is the place where we set the description, but apparently not. 

@de-jcup please have a look at this. Maybe, you can explain that.

Jeeppler (approved):
Approved.

sechub-sereco/src/main/java/com/daimler/sechub/sereco/importer/SarifV1JSONImporter.java
de-jcup: @Jeeppler , @4w31ss3 : `CheckmarxV1XMLImporter` does set description to empty - that is correct (and ugly). The reason was that at time of implementation checkmarx always returned only full HTML. So we were not able to set normal description text (about the problem) without formatting etc.

When SARIF defines description as plain text we should always reuse the description and just set it. If this cannot be ensured - e.g. some tools integrate HTML here, we could (maybe in a future request) do a filtering and only convert into our own description fields at import time when format is accepted (e.g. asciidoc/markdown ... )

**Sum up:** I would think we should set the description. Filtering unwanted content should be done as well (but in a later issue)

de-jcup: @4w31ss3 : Please create a new github issue for setting the description field and link it inside the #386  issue. 
@Jeeppler : we must inspect content from different SARIF messages etc. and the implementation to check which messages and how to filter content is not so trivial. So we let this issue as is and do not set the description here. In a leter issue this will be handled

4w31ss3: @de-jcup: I created #575 for setting of the description field.

de-jcup:
After testing around it appeared that the importers did not work inside sechub.

PR #693, opened 2021-07-06
Feature 690 fix dirty flag on release builds

### LINKED ISSUE #690: Fix dirty flag on releases
## Situation
We have in release documentation and also in client output now a "dirty-$timestamp" .
"dirty" means that after checkout of a release tag the git repository was changed, so it's not 100% the release tag part but contains some changes...

Examples: 
- https://github.com/Daimler/sechub/releases/download/v0.25.0-server/sechub-restapi-0.25.0.pdf
- https://github.com/Daimler/sechub/releases/tag/v0.24.0-client

## Wanted
A release may not contain dirty tags

## Solution
- release build breaks when there  are any git changes
- inspect which was the problematic changed part
- fix the problem

de-jcup:
This PR 
- closes #690

### DIFF
--- a/.github/workflows/create-releases.yml
+++ b/.github/workflows/create-releases.yml
@@ -28,16 +28,18 @@ jobs:
         ref: master
     
     
-    # Create local tags, so we build documentation for this tag...
-    - name: "Tag server version: v${{ github.event.inputs.server-version }}-server - if defined" 
+    # Create temporary local tags, so we build documentation for this tag...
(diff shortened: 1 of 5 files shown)


de-jcup:
As a remark: The new `assertReleaseable` gradle task  is not at master branch available. So the review must be done, the PR merged to develop and after this to master. Then we can trigger a release and finally test the changes are as expected. The created draft release can be dropped after the execution.

sven-dmlr (approved):

Issue #689, opened 2021-07-06
Document Diffie-Hellman key increase and set default to 2048


de-jcup:
## Situation
The default Diffie-Hellman key length in Java is still 1024. Reading https://weakdh.org/ you will find:

> If you’re a sysadmin or developer …
> 
> Make sure any TLS libraries you use are up-to-date, that servers you maintain use 2048-bit or larger primes, and that clients you maintain reject Diffie-Hellman primes smaller than 1024-bit.

## Wanted
We mention this in our SecHub and PDS documentation AND we set a default inside `application.yaml` for SecHub Server AND for PDS a default value of `2048`.

## Technical information/solution

Starting with JDK8 it became possible to set DH key length greater than 1024, by setting system property `jdk.tls.ephemeralDHKeySize` to wanted key size.

Links:
- [AdoptOpenJDK11: Javadoc describing history about 1024 key length](https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/19fb8f93c59dfd791f62d41f332db9e306bc1422/src/java.base/share/classes/sun/security/ssl/DHKeyExchange.java#L317-L356)
- [AdoptOpenJDK11: Property entry ](https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/19fb8f93c59dfd791f62d41f332db9e306bc1422/src/java.base/share/classes/sun/security/ssl/DHKeyExchange.java#L254-L255)
- [AdoptOpenJDK11: Min and max values for property ](https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/19fb8f93c59dfd791f62d41f332db9e306bc1422/src/java.base/share/classes/sun/security/ssl/DHKeyExchange.java#L276-L285)

de-jcup:
Setting simply `jdk.tls.ephemeralDHKeySize`inside` application.yaml` is unfortunately not the solution because it is not automatically injected by spring boot as system property.

So we will implement special "SystemPropertyInjector" components which provide specialized spring boot value
`sechub.security.diffiehellman.length`. This value will be set inside `applicaiton.yaml` and then injected automatically as system property `jdk.tls.ephemeralDHKeySize` by the injector.

PR #697, opened 2021-07-07
Changed default diffie hellman length to 2048 #689

de-jcup:
- introduce SystemPropertyInjector classes and
  special keys for PDS and SecHub server, so configurable
  by `application.yaml` files
- `application.yaml` files do now contain 2048 as default value
- documentation will be generated and points to issue at github
  for details

### DIFF
--- a/sechub-pds/src/main/java/com/daimler/sechub/pds/PDSSystemPropertyInjector.java
+++ b/sechub-pds/src/main/java/com/daimler/sechub/pds/PDSSystemPropertyInjector.java
@@ -1,0 +1,30 @@
+// SPDX-License-Identifier: MIT
+package com.daimler.sechub.pds;
+
+import javax.annotation.PostConstruct;
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+
+/**
+ * This component injects some special PDS Spring Boot values into corresponding
+ * system properties. So we can configure those parts in our `application.yaml`
+ * file.
+ * 
+ * @author Albert Tregnaghi
+ *
+ */
+@Component
+public class PDSSystemPropertyInjector {
+
+    // see https://github.com/Daimler/sechub/issues/68 for details
+    @PDSMustBeDocumented(value="Define diffie hellman key length, see https://github.com/Daimler/sechub/issues/689 for details",scope="security")
+    @Value("${pds.security.diffiehellman.length}")
+    private String diffieHellmanLength;
+
+    @PostConstruct
+    public void setDiffieHellmanLength() {
+        System.setProperty("jdk.tls.ephemeralDHKeySize", diffieHellmanLength);
+    }
+}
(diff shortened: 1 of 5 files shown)


sechub-server/src/main/java/com/daimler/sechub/server/SecHubSystemPropertyInjector.java
```diff
@@ -0,0 +1,31 @@
...
+import javax.annotation.PostConstruct;
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+import com.daimler.sechub.sharedkernel.MustBeDocumented;
+
+/**
+ * This component injects some special SecHub Spring Boot values into
+ * corresponding system properties. So we can configure those parts in our
```
Jeeppler, line 13:
```suggestion
 * corresponding JVM system properties. So we can configure those parts in our
```

sechub-pds/src/main/java/com/daimler/sechub/pds/PDSSystemPropertyInjector.java
```diff
@@ -0,0 +1,30 @@
...
+
+import javax.annotation.PostConstruct;
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+
+/**
+ * This component injects some special PDS Spring Boot values into corresponding
+ * system properties. So we can configure those parts in our `application.yaml`
```
Jeeppler, line 12: Those properties are specific to the JVM not to the operating system.

```suggestion
 * JVM system properties. So we can configure those parts in our `application.yaml`
```

sechub-server/src/main/java/com/daimler/sechub/server/SecHubSystemPropertyInjector.java
```diff
@@ -0,0 +1,31 @@
...
+ * `application.yaml` file.
+ * 
+ * @author Albert Tregnaghi
+ *
+ */
+@Component
+public class SecHubSystemPropertyInjector {
+
+    // see https://github.com/Daimler/sechub/issues/68 for details
```
Jeeppler, line 22: This comments does not lead to anything related to Diffie-Hellman or key length. Please remove it.

Jeeppler (changes requested):
See comments.

sechub-server/src/main/java/com/daimler/sechub/server/SecHubSystemPropertyInjector.java
de-jcup: will remove both (pds + sechub server)

de-jcup: done

sechub-server/src/main/java/com/daimler/sechub/server/SecHubSystemPropertyInjector.java
de-jcup: It should be clear that this are JVM system properties .. but okay...

de-jcup: done

Jeeppler (approved):

PR #720, opened 2021-07-14
Integration test profiles now only available on loopback address #719

### LINKED ISSUE #719: Bind integrationtest profiles to loopback address only
## Situation
As default spring boot web applications are available in complete local network.
Starting integration test servers with default passwords (available at github...) seems to be not a good choice, even when the network is trusted.

At the moment developers need to define firewall rules to their Jenkins server instances to whitelist  only
accepted ports. 

see https://stackoverflow.com/questions/46192242/spring-boot-webapp-localhost-only/46875199

## Wanted
For any profile starting with default credentials, we automatically disallow usage outside local machine. So it is secure per default.

## Solution
inside `integrationtest` and `pds_integrationtest` profiles we define server address to loopback only:

```
server.address=127.0.0.1
```

de-jcup:
PR closes #719

### DIFF
--- a/sechub-pds/src/main/resources/application-pds_integrationtest.yml
+++ b/sechub-pds/src/main/resources/application-pds_integrationtest.yml
@@ -2,6 +2,9 @@
 spring.profiles.include:
 - pds_localserver
 
+server:
+  address: 127.0.0.1
+
 # next lines is a default when running inside IDE. Links to sechub-integration aprts
 sechub.pds.config.file: "./../sechub-integrationtest/src/main/resources/pds-config-integrationtest.json"
 sechub.pds.workspace.rootfolder: "./../sechub-integrationtest/build/test-results/pds-runtime"
--- a/sechub-server/src/main/resources/application-integrationtest.yml
+++ b/sechub-server/src/main/resources/application-integrationtest.yml
@@ -5,6 +5,8 @@ spring.profiles.include:
 - mocked_notifications
 - admin_access
 
+server:
+  address: 127.0.0.1
 
 sechub:
   config:


sechub-pds/src/main/resources/application-pds_integrationtest.yml
```diff
@@ -2,6 +2,9 @@
 spring.profiles.include:
 - pds_localserver
 
+server:
```
Jeeppler, line 5: This should be set for localserver as well.

Jeeppler (changes requested):
By default localserver should bind to 127.0.0.1 as well and maybe dev. Users binding to 0.0.0.0 should explicitly have to configure it.

sechub-pds/src/main/resources/application-pds_integrationtest.yml
de-jcup: ### General
Are you talking about SecHub server or PDS ?

I did explicit set loopback to integration tests only because adding to local server profile  This would make the server not working per default!

### Integration test profile situation
The integration test profile has default passwords and that's the reason why we do NOT want to have the server exposed.

### Differ between PDS SecHub
#### PDS
When somebody uses a PDS it's 99% not on the same server except in our integration tests  PDS "implementations" are normally running inside a docker container, docker compose or in k8s. Defining here a "127.0.0.1" would bring up problems / more configuration. The docker run call

```
docker run --rm -p 127.0.0.1:$exposed_port:$container_port/tcp $env_data --name $container_name $image_name
```
should protect the PDS instance.

RESULT: --> for PDS I would not change the behaviour.

### SecHub
For testing purposes etc. it can happen that a developer starts the server with custom / dedicated profiles. But... when not using integration test profile the developer MUST define a password and there are no hardcoded passwords . 

Yes, the server will be visible in network. But we assume that our server software is secure

RESULT: --> for SecHub I would not change the behaviour.

PR #736, opened 2021-07-29
Changed report rendering for finding ids and call steps #576

### LINKED ISSUE #576: HTML Report: Id ambigious
At the moment, the call number is named `Id` instead of `Call` in the HTML report. This is confusing, because the JSON report has an `Id` field. One can mistake the `Id` in the HTML report as the Id in the JSON report. In addition, the call stack should start with 1 and not 0.

Screenshot at 2021-03-04 13-00-44

The solution is to:

1. Introduce a new Id column, which shows the `Id` from the JSON file
2. Rename the `Id` in the call stack to `Call` or `Call Number`
3. Start counting the calls from the callstack at 1

call_stack_should_be

---
<sup>Jeremias Eppler <jeremias.eppler@daimler.com>, Daimler TSS GmbH, [imprint](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)</sup>

### LINKED ISSUE #597: Report and interactive falsePositive Code does not match
Report and interactive falsePositive Code does not match
Report should have first occurence and not last of call stack on top
int that case id 0 on top instead of id 13

image

de-jcup:
This PR 
- closes #576 
- closes #597


sechub-scan/src/main/java/com/daimler/sechub/domain/scan/report/HTMLReportCSSFragementGenerator.java
```diff
@@ -0,0 +1,58 @@
...
+ * 
+ * 5. After css file is as wanted... --> Start this generator. --> will adopt
+ * css data into fragement file.
+ * 
+ * 6. commit push done...
+ * 
+ * @author Albert Tregnaghi
+ *
+ */
+public class HTMLReportCSSFragementGenerator {
```
winzj, line 31: It is nice to have a generator to keep the files in sync. What about moving this into the developer tools, because it is not production code?

winzj (changes requested):

sechub-scan/src/main/java/com/daimler/sechub/domain/scan/report/HTMLReportCSSFragementGenerator.java
de-jcup: Ok. Done.

winzj (approved):

Issue #749, opened 2021-08-06
Documentation CSS File


Jeeppler:
The CSS file `sechub_style.css` is hard to maintain going forward and should be replaced with our own CSS file.

PR #752, opened 2021-08-10
Rewrite CSS File and Remove Images

Jeeppler:
- Rewrote the CSS file
- Removed all files without clear license
- Refactored HTML file

Closes: #749


sven-dmlr (approved):
Looks good. (checked with multiple browsers)

PR #786, opened 2021-09-09
Bugfix + improved PDS period time check by providing milliseconds instead of minutes in configurations

### LINKED ISSUE #785: Default timeout fallback for PDS scan operation uses wrong PDS install setup method
## Situation
The defaults inside `PDSInstallSetupImpl` are 
- one minute for period scan result checks
- 4 hours max until timeout

But: `PDSExecutorConfigSuppport` does not use the 4 hours as fallback for the scan time out but instead also the default scan period time!

### Impact
When somebody does not explicit set a timeout inside the product executor configuration for PDS it does normally happen that PDS scans will be stopped by adapter because of timeout after first try! Current workaround is to set a value inside the executor configuration parameters.

## Wanted
Bug must be fixed! PDS adapters shall work out of the box.

## Solution
- change the code to use correct value
- additional change the methods for the scan period timeout to a more readable variant (currently the methods names are too similar and this was the reason for the bug!)
- add debug log entry while PDS adapter is waiting with information about the elapsed time and maximum time to wait etc.

### LINKED ISSUE #787: Change PDS check period time unit from minutes to milliseconds + introduce new configuration identifiers
### Configuration ids for defaults 
While implementing #785 I did not change the configuration identifier. We should change those config identifiers as well - so it's more understandable. Also the next part (change time unit to milliseconds for periodic ceck) should be handled.

### Minutes for period checks are not a good option for PDS
Also it became clear that having a check period only in minutes has a few disadvantages - e.g. Some scanners are extreme fast and do not need minutes to do a check. Also every integration test using a real PDS instance will wait at least 1 minute. And this increases our build times much! 

_Remark: In our special mocked adapters this is not a problem because the time to wait etc. is handled there in a special way. An in the normal product environments to wait minutes is also correct!_

So we needs this for our PDS integration tests only, because here we do not mock the product communication but do really start a product communication etc. 
Additionally  we had often the situation that scans with PDS and some CLI tools  are extreme fast. So we want to be able to setup a dedicated short time  inside those pds executor configurations as well.

### General
- change config identifiers for default values in `PDSSetupInstallImpl` appropriate and change documentation 
- change PDS runtime configuration entry identifier so it is clear the check period is in milliseconds now
- add to abstractAdapterConfigBuilder setTimeToWaitForNextCheckOperationInMilliseconds and use it only by PDS executors - we keep the other executors "as is" (keep stable).
- change the minimum value for checks in milliseconds to 500 ms (if not already done)
- change default value for PDS check time period to 30 seconds
- change integration test application settings to use 500 ms (so our test checks are very fast)

### Identifier changes
- replace  `sechub.adapter.pds.default.scanresultcheck.period.minutes` with `sechub.adapter.pds.default.check.timetowait.milliseconds`
- replace `sechub.adapter.pds.default.scanresultcheck.timeout.minutes` with
  `sechub.adapter.pds.default.timeout.minutes`

de-jcup:
This PR will
- close #785 
- close #787

### DIFF
--- a/sechub-adapter-pds/src/main/java/com/daimler/sechub/adapter/pds/PDSAdapterV1.java
+++ b/sechub-adapter-pds/src/main/java/com/daimler/sechub/adapter/pds/PDSAdapterV1.java
@@ -61,8 +61,9 @@ public class PDSAdapterV1 extends AbstractAdapter<PDSAdapterContext, PDSAdapterC
 
     private void waitForJobDone(PDSContext context) throws AdapterException {
         PDSAdapterConfig config = context.getConfig();
-        UUID jobUUID = config.getSecHubJobUUID();
-        UUID uuid = context.getPdsJobUUID();
+
+        UUID secHubJobUUID = config.getSecHubJobUUID();
+        UUID pdsJobUUID = context.getPdsJobUUID();
 
         int count = 0;
         boolean jobEnded = false;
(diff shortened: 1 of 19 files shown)


Jeeppler (approved):
I like, that the SecHub JobUUID and the PDS JobUUID are shown together in the PDS Adapter output.

PR #815, opened 2021-10-01
PDS Test Script reusable

Jeeppler:
Improves the PDS test script by:

- allowing to define any product
- defining the number of retries via environment variable
- providing JSON files to create a job


winzj (approved):
I think it is a very good idea to have these changes.

PR #880, opened 2021-11-08
Feature 877 bugfix old reports not fetchable

### LINKED ISSUE #877: Old existing reports cannot be fetched with 0.27.1 server version
## Situation
When fetching a report created in DB before `0.27.1-server ` the download fails with a HTTP 500

## Wanted
Must work with old report data as well

## Solution
- Integration test 
- Fix it

de-jcup:
This PR 
- closes #877


sven-dmlr (approved):
looks good. Thank you!

PR #891, opened 2021-11-15
Fix problem with versions on library modules #890

### LINKED ISSUE #890: Cannot build developer admin UI with gradle in v0.27.2-server
## Situation
Do following:
```
git checkout v0.27.2-server

./gradlew buildDeveloperAdminUI
```
This will fail with:

```
> Task :sechub-developertools:buildDeveloperAdminUI FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':sechub-developertools:buildDeveloperAdminUI'.
> Cannot expand ZIP '/home/albert/develop/projects/sechub/sechub-open-source/sechub-commons-model/build/libs/sechub-commons-model-0.27.2.jar' as it does not exist.

```
### First analyse
The problem is that we provide own version numbers for commons-model and commons-commons-core (so we can update and fix them independently from server releases)

## Wanted 
Build without problems...

@sven-dmlr :  A workaround is to use the current development branch (results in 0.0.0 which works always...) We got the problems only when we got the server version tag.

de-jcup:
This PR closes #890

### DIFF
--- a/build-maven.gradle
+++ b/build-maven.gradle
@@ -19,13 +19,20 @@ subprojects {
         return;
     }
 
+    // we always set the libraries version in this case
+    // If we publish and the version is set, it means the dedicated, published lib version
+    // If not defined, the library version is 0.0.0 which represents a snapshot -
+    // the snapshot itself is related to the spring boot application (server or PDS) which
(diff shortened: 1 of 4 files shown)


sven-dmlr (approved):
Looks good.
A local test on the branch was successful.

Issue #771, opened 2021-09-02
Provide gradle task to format sources


de-jcup:
## Situation
In many code reviews source formatting is mentioned and the eclipse auto formatter  must be applied manually to these files. This is necessary but very cumbersome

## Wanted
Per command line there should be one call to format all source codes automatically - with our wanted setup.

## Solution
- We use a dedicated gradle plugin to format the code - e.g. `spotless` which can use eclipse formatter setup. see https://plugins.gradle.org/plugin/com.diffplug.spotless - if not suitable, there are others: see 
https://plugins.gradle.org/search?term=format

We can use
`./gradlew spotlessCheck` to verify

but also
`./gradlew spotlessApply` to automatically format code (currently only Java)

- We introduce also `prepareGitPush` task which dependsOn `spotlessApply`

de-jcup:
Gradle config parts are described at https://github.com/diffplug/spotless/blob/main/plugin-gradle/README.md

de-jcup:
Hmm... groovy formatting makes problems. Having always some eclipse : 
```
Step 'eclipse groovy formatter' found problem in 'build.gradle':

WARNING: Formatting ignored a conflicting text edit
java.lang.IllegalArgumentException: 
WARNING: Formatting ignored a conflicting text edit
        at com.diffplug.spotless.extra.eclipse.groovy.GrEclipseFormatterStepImpl.format(GrEclipseFormatterStepImpl.java:88)
        at jdk.internal.reflect.GeneratedMethodAccessor595.invoke(Unknown Source)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at com.diffplug.spotless.extra.groovy.GrEclipseFormatterStep.lambda$apply$0(GrEclipseFormatterStep.java:57)
        at com.diffplug.spotless.FormatterFunc.apply(FormatterFunc.java:32)
...

Was not able to get rid of it, so decided to just turn off groovy inspection by spotless
```

de-jcup:
Remark: It was necessary to execute `spotlessApply` in the final commit, otherwise the build would fail because gradle spotless plugin does now always check inside every build that there are no formatting issues...

PR #903, opened 2021-11-29
Feature 728 buld java api

### LINKED ISSUE #728: Build sechub-api-java
This task is a follow up issue to #377. In task #377, the OpenAPI file generation and OpenAPI Client Source Code generation were improved. However, building and testing were left to this issue.

## Steps

1. Build a "FatJar" from the current generated source code
2. Publish "FatJar" to local maven
3. Import and use the "FatJar" in the SecHub Java Client (sechub-java-client)
4. Run tests to make sure the SecHub Java Client works

## Other tasks

- [ ] Adjust github actions workflow
- [ ] Adjust github actions release workflow
- [ ] Adjust Jenkins file

### LINKED ISSUE #746: Rename sechub-client-java to sechub-api-java
## Situation
We have a sechub client which is written in go.

But we have also the gradle subproject "sechub-client-java" which contains a wrapper and generated java code (open api generator) - it is still ongoing and not complete, but will be available in the near future. With this subproject (and the planned go and typescript variants) we provide an API layer for other tools like IDE plugins or caller plattforms, so they do not need to implement API calls by themselves. 

So when we talk about a "client" it is not clear what is really meant.

## Wanted
It shall be clear what is meant by the word "client" and what is the "api layer" for other tools.

## Solution
1. We use the term `client` only with gradle subproject `sechub-cli` .
2. We rename `sechub-client-java` to `sechub-api-java`
    (+ adopt  includes gradle build, documentation etc.)

Further api layer will also use same name pattern : `sechub-api-$language`

de-jcup:
This PR 

- closes #728 
- closes #746


buildSrc/src/main/groovy/VersionData.groovy
```diff
@@ -93,7 +93,7 @@ class VersionData{
     }
     
     /**
-     * Inpsect versoin - if not starting with 0.0.0 this means it's a release ando
+     * Inpsect version - if not starting with 0.0.0 this means it's a release ando
```
Jeeppler, line 96: What is `ando`?

Jeeppler (changes requested):
Found one issue. Otherwise it looks good.

winzj (approved):
Looks good to me. Please look at the other review, looking at the whole file there, I think it is just typing error.

buildSrc/src/main/groovy/VersionData.groovy
de-jcup: I will fix following typos:

- Inpsect : Inpspect
- versoin : version
- ando: i will check

Remark: This part is not changed inside this commit / PR. So normally not being reviewed here - or ?

Jeeppler: My assumption was, that you forgot to change the `ando`. 

In general, it might make sense to use automated spellchecking while running the code formatting (see: #771)

Jeeppler (approved):
Looks good to me.

Issue #918, opened 2021-12-21
PDS: Add HTTP GET on /api/anonymous/check/alive endpoint


sven-dmlr:
# Problem:
When trying to set up Kubernetes `livenessProbe` or `startupProbe` with PDS, HTTP HEAD requests are not supported.
HTTP GET is needed. (see https://github.com/kubernetes/kubernetes/issues/49937)

# Solution:
Please add HTTP GET on /api/anonymous/check/alive endpoint of PDS.
(Related issue for SecHub server was: #378)

PR #920, opened 2021-12-22
introduce check alive via GET

Jeeppler:
Extend the check alive call to include GET as a possible method. This is important for readiness and liveliness probes in Kubernetes.

closes: #918

### DIFF
--- a/sechub-pds/src/main/java/com/daimler/sechub/pds/monitoring/PDSAnonymousCheckAliveRestController.java
+++ b/sechub-pds/src/main/java/com/daimler/sechub/pds/monitoring/PDSAnonymousCheckAliveRestController.java
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: MIT
 package com.daimler.sechub.pds.monitoring;
 
+import org.springframework.http.MediaType;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.RestController;
@@ -14,9 +15,10 @@ public class PDSAnonymousCheckAliveRestController {
 
 	/* @formatter:off */
 	@UseCaseAnonymousCheckAlive(@PDSStep(name="rest call",description = "anybody - even anonymous - checks server alive.",number=1))
-	@RequestMapping(path = PDSAPIConstants.API_ANONYMOUS + "check/alive", method = RequestMethod.HEAD)
-	public void checkAlive() {
-	    /* do nothing here - its just a HEAD request */
+	@RequestMapping(path = PDSAPIConstants.API_ANONYMOUS + "check/alive", method = {RequestMethod.GET, RequestMethod.HEAD}, produces = {MediaType.APPLICATION_JSON_VALUE})
+	public String checkAlive() {
+	    /* empty result, only HTTP STATUS 200 OK is of interest */
+	    return "";
 	}
 	/* @formatter:on */
 }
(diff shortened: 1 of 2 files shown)


sechub-pds/src/main/java/com/daimler/sechub/pds/monitoring/PDSAnonymousCheckAliveRestController.java
```diff
@@ -14,9 +15,10 @@
 
 	/* @formatter:off */
 	@UseCaseAnonymousCheckAlive(@PDSStep(name="rest call",description = "anybody - even anonymous - checks server alive.",number=1))
-	@RequestMapping(path = PDSAPIConstants.API_ANONYMOUS + "check/alive", method = RequestMethod.HEAD)
-	public void checkAlive() {
-	    /* do nothing here - its just a HEAD request */
+	@RequestMapping(path = PDSAPIConstants.API_ANONYMOUS + "check/alive", method = {RequestMethod.GET, RequestMethod.HEAD}, produces = {MediaType.APPLICATION_JSON_VALUE})
```
de-jcup, line 18: My first thought was: Why does the HEAD also result in an empty string?
After reading https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD
> Warning: A response to a HEAD method should not have a body. If it has one anyway, that body must be ignored: any representation headers that might describe the erroneous body are instead assumed to describe the response which a similar GET request would have received.

I think it's really not a good idea to annotate with head and return an empty string here. Also not to annotate with MediaType.APPLICATION_JSON_VALUE for head

Additionally we want both parts generated in documentation (HEAD + GET) - I am not sure if both parts are generated or not (unfortuneately the build results are currently not available).

So please so me a favor and use/create two Methods instead:

`checkAliveByGET` - your implementation (but without HEAD annotation)
and 
`checkAliveByHead` - keep old code as it was without any return string

### Another one:
We have our sprint actuators, so readyness probe should be available - see https://www.baeldung.com/spring-liveness-readiness-probes for example. So why do we need to implement this again?

The reason for the head was originally to be given as little as possible to the outside world - but since the actuators can now be switched on / off individually and can also run on their own ports, the check live implementation with HEAD is actually only necessary for downward compatibility.

We should go the standard way here and use the actuators.
So my proposal would be to mention this inside the issue and close the PR.

de-jcup (changes requested):

sechub-pds/src/main/java/com/daimler/sechub/pds/monitoring/PDSAnonymousCheckAliveRestController.java
de-jcup: Hmm... okay. We have implemented this at server side exactly the same way as you did (one method only). The api documentation on server side did work well (`GET` and `HEAD` were generated in doc via variants in rest doc) - so when we provide REST API doc this will be no problem at all. 

The liveness probes should be used by people which are using K8s - but not everybody does. So we should provide HEAD and GET in our own API here as well.

Summary: Let us keep the "one method" approach

sechub-pds/src/test/java/com/daimler/sechub/pds/monitoring/PDSAnonymousCheckAliveRestControllerMockTest.java
```diff
@@ -53,6 +53,22 @@ public void a_get_execution_status_calls_executionService_and_returns_result() t
 
         /* @formatter:on */
 
+    }
+	
+    @Test
+    public void calling_check_alive_get_returns_HTTP_200() throws Exception {
+        /* prepare */
```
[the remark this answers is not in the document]
de-jcup: Other test method were copied from has also an empty prepare - so keep as is.

de-jcup (approved):

Issue #926, opened 2022-01-03
PDS + GoSec: Remove the Alpine Base Image


Jeeppler:
@sven-dmlr and I decided to focus on Ubuntu only for all future PDS images: https://github.com/Daimler/sechub/pull/922#issuecomment-1004027836. As a result, we will remove the files for the Alpine base image.

Issue #930, opened 2022-01-03
Document decision to use Ubuntu for PDS


Jeeppler:
Document the decision of using Ubuntu for all PDS solutions base images. For more information see: https://github.com/Daimler/sechub/pull/922#pullrequestreview-841298574

sven-dmlr:
Meanwhile, Debian is the base image for PDS solutions.

@Jeeppler Can we close this issue?

Jeeppler:
@sven-dmlr no, this issue cannot be closed. We should document our decision somewhere.

PR #922, opened 2021-12-23
Upgrade GoSec + PDS

Jeeppler:
Upgrade GoSec and refactor docker files

- Upgrade GoSec to 2.9.5
- Add a new mock.sarif.json file generated by GoSec 2.9.5
- Refactor dockerfiles
- Add `wget` command to download checksum for PDS
- Add `--remove-orphans` to build command
- No verbose output of wget

closes: #921

PR #931, opened 2022-01-04
Remove Alpine as Base Image

Jeeppler:
As discussed in #922 and documented in #930, this pull request removes Alpine as base image.

Close: #926


sven-dmlr (approved):
Looks good. Thank you!

Issue #691, opened 2021-07-06
Upgrade to SpringBoot 2.6


Jeeppler:
Upgrade to SpringBoot 2.6.

Jeeppler:
SpringBoot 2.6 is not yet released, but a snapshot is already available: https://docs.spring.io/spring-boot/docs/2.6.0-SNAPSHOT/api/.

A M1 release of 2.6 is getting prepared for release in July: https://github.com/spring-projects/spring-boot/milestone/214

Jeeppler:
SpringBoot 2.6 M2 was released: https://spring.io/blog/2021/08/19/spring-boot-2-6-0-m2-is-now-available

Jeeppler:
Spring Boot 2.6 RC1 was released: https://spring.io/blog/2021/10/21/spring-boot-2-6-0-rc1-is-available-now

Spring Boot 2.6 is scheduled for release on the 18th Nov.

Jeeppler:
Spring Boot 2.6 is available: https://spring.io/blog/2021/11/19/spring-boot-2-6-is-now-available

Jeeppler:
SpringBoot 2.5 will only be supported until May 2022: https://spring.io/projects/spring-boot#support

Jeeppler:
SpringBoot gradle dependencies: https://github.com/spring-projects/spring-boot/blob/v2.6.2/spring-boot-project/spring-boot-dependencies/build.gradle

PR #942, opened 2022-01-13
Upgrade SpringBoot to 2.6

Jeeppler:
- upgrade SpringBoot to 2.6
- upgrade Restdoc
- add Dom4J dependency, because it is not a Hibernate dependency anymore
- rewrite build.gradle of sechub-sereco and
sechub-storage-sharedvolume-spring to new Gradle format
- added mockito inline to sechub-storage-sharedvolume-spring
- downgrade flyway from 8.2.3 to 8.0.5 to be at the same level as SpringBoot

closes: #691

### DIFF
--- a/sechub-sereco/build.gradle
+++ b/sechub-sereco/build.gradle
@@ -8,7 +8,11 @@
  
 dependencies {
 	implementation library.jsoup
-	implementation library.dom4j
+	implementation('org.dom4j:dom4j:2.1.3') {
+		exclude group: 'xpp3', module: 'xpp3'
+		exclude group: 'pull-parser', module: 'pull-parser'
+	}
+
 	    
     implementation project(':sechub-shared-kernel')
     implementation project(':sechub-sarif')
--- a/sechub-sereco/build.gradle
+++ b/sechub-sereco/build.gradle
@@ -7,12 +7,13 @@
  */
  
 dependencies {
-	compile library.jsoup
+	implementation library.jsoup
+	implementation library.dom4j
 	    
-    compile project(':sechub-shared-kernel')
-    compile project(':sechub-sarif')
+    implementation project(':sechub-shared-kernel')
+    implementation project(':sechub-sarif')
 
-    compile project(':sechub-shared-kernel')
+    implementation project(':sechub-shared-kernel')
     
-    testCompile project(':sechub-testframework')
+    testImplementation project(':sechub-testframework')
 }
--- a/sechub-sereco/build.gradle
+++ b/sechub-sereco/build.gradle
@@ -8,7 +8,7 @@
  
 dependencies {
 	implementation library.jsoup
-	implementation('org.dom4j:dom4j:2.1.3') {
+	implementation(library.dom4j) {
 		exclude group: 'xpp3', module: 'xpp3'
 		exclude group: 'pull-parser', module: 'pull-parser'
 	}
--- a/sechub-server/src/main/resources/application.yml
+++ b/sechub-server/src/main/resources/application.yml
@@ -44,6 +44,8 @@ spring:
   main:
     # see https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.1-Release-Notes#bean-overriding
     allow-bean-definition-overriding: true
+    # see https://stackoverflow.com/questions/70036903/spring-boot-application-fails-to-start-after-upgrading-to-2-6-0-due-to-circular
+    allow-circular-references: true
   servlet: 
     # MultiPart file-size limits,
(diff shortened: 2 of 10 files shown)


sechub-sereco/build.gradle
```diff
@@ -7,12 +7,17 @@
  */
  
 dependencies {
-	compile library.jsoup
+	implementation library.jsoup
+	implementation(library.dom4j) {
```
de-jcup, line 11: Do we need the dom4j library because of jsoup?
We normally do all JSON parts with Jackson library...

I am not so happy having jsoup and also dom4j here again. Both are good frameworks, but less is more ... at least after log4shell etc...

If jsoup is the reason why we need dom4j we should think about usage and maybe if it could be possible to drop this dependency /replace it ?

What do you think?

sechub-server/src/main/resources/application.yml
```diff
@@ -44,6 +44,8 @@ spring:
   main:
     # see https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.1-Release-Notes#bean-overriding
     allow-bean-definition-overriding: true
+    # see https://stackoverflow.com/questions/70036903/spring-boot-application-fails-to-start-after-upgrading-to-2-6-0-due-to-circular
+    allow-circular-references: true
```
de-jcup, line 48: Ahem... 
The referenced stackoverflow entry is closed now because of missing debug information
https://stackoverflow.com/questions/70073748/i-cant-update-my-webapp-to-spring-boot-2-6-0-2-5-7-works-but-2-6-0-doesnt
is still open and has a same "solution".

But this can only be a temporary "solution"! We need to solve the problem. Having a circular reference can normally be fixed . So do we have already an github issue for this? Have you already tried to fix the problem with out the "solution" - workaround?

de-jcup (changes requested):

sechub-sereco/build.gradle
Jeeppler: Dom4j has noting to do with jsoup. Jsoup parses HTML and is only required for the [NetsparkerHtmlToAsciiDocConverter](https://github.com/Daimler/sechub/blob/develop/sechub-sereco/src/main/java/com/daimler/sechub/sereco/importer/NetsparkerHtmlToAsciiDocConverter.java). The moment we remove the Netsparker code in our repository, Jsoup will not be required anymore.

Dom4j is a very important part of  parsing the XML reports in the SERECO importer. Until now, Dom4J was a dependency of Hibernate and therefore available. That is probably the reason why the report parsing in SERECO uses Dom4J. 

In Hibernate 5.6 the Dom4J dependency got removed and therefore the report parsing did not work in several places and it resulted in the project not being able to compile. SERECO importers using Dom4j: 

- [NessusV1XMLImporter](https://github.com/Daimler/sechub/blob/develop/sechub-sereco/src/main/java/com/daimler/sechub/sereco/importer/NessusV1XMLImporter.java#L8)
- [NetsparkerV1XMLImporter](https://github.com/Daimler/sechub/blob/develop/sechub-sereco/src/main/java/com/daimler/sechub/sereco/importer/NetsparkerV1XMLImporter.java#L7)
- [CheckmarxV1XMLImporter](https://github.com/Daimler/sechub/blob/develop/sechub-sereco/src/main/java/com/daimler/sechub/sereco/importer/CheckmarxV1XMLImporter.java#L8)

Because of those three importers it was necessary to add Dom4j as a dependency. However, Dom4J comes with the pull-parser and xpp3, which caused problems with Logback. Logback needs a XML parser to read it's own configuration files: 

- [SecHub-Server: logback-spring.xml](https://github.com/Daimler/sechub/blob/develop/sechub-server/src/main/resources/logback-spring.xml)
- [SecHub-PDS: logback-spring.xml](https://github.com/Daimler/sechub/blob/develop/sechub-pds/src/main/resources/logback-spring.xml)

Both, the pull-parser and xpp3 caused problems and therefore they need to be excluded to get the old behavior back. The old behavior relied on JAXP and the default XML parser in Java SE.

sechub-server/src/main/resources/application.yml
Jeeppler: No, I have not tried to fix the problem without the `allow-circular-references` set to `true` as I do not understand the issue well enough. I wanted to have your feedback on this. I opened a issue for this: #944.

sechub-sereco/build.gradle
de-jcup: Funny that I did not remember that I used dom4j for the Netsparker Importer  when I wrote the initial implementation... But was a long time ago ... Thanx for clarification.

PR #952, opened 2022-01-21
apply-headers script refactored + bash support

### LINKED ISSUE #436: Enhance apply-headers.sh for Bash scripts
Currently Bash scripts are not covered by apply-headers.sh which adds SPDX license headers to each file.
So one has to add it manually or license infortmation is missing.

We should enhance the script so when a shebang (`#!`) is found, then the SPDX license header is added on the second line.

---

sven-dmlr:
- use sed to insert spdx messages
- Bash script support (2nd line inserts)
- Batch, SQL, Ruby, C, Python file support
- safe coloring
- removed are-you-sure because changes can be undone easily via git
- removed template files (redundant contents) in favor of doing it inside the script with one variable

closes #436

### DIFF
--- a/apply-headers.sh
+++ b/apply-headers.sh
@@ -1,102 +1,92 @@
 #!/bin/bash
-
 # SPDX-License-Identifier: MIT
 
-RED='\033[0;31m'
-LIGHT_RED='\033[1;31m'
-LIGHT_GREEN='\033[1;32m'
-BROWN='\033[0;33m'
-NC='\033[0m' # No Color
+function isColoredTerminal(){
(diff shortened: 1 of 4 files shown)


Jeeppler (approved):
Looks good to me.

sven-dmlr:
> Looks good to me.

Thanks for the quick review, @Jeeppler  :+1: :tada: 

Before merging, I wait if @de-jcup suggests to add more file extensions to check/apply.
See https://github.com/Daimler/sechub/issues/436#issuecomment-1018676562

Issue #899, opened 2021-11-22
Ensure PDS_SCAN_TARGET_URL never ends with a slash


de-jcup:
### Situation

URI defined within a webScan configuration are already given to `PDS `as `PDS_SCAN_TARGET_URL ` :

https://github.com/Daimler/sechub/blob/cbd90fede4a1894bfc05cbec1bd278bc5c0c6356/sechub-adapter-pds/src/main/java/com/daimler/sechub/adapter/pds/PDSWebScanConfigImpl.java#L85

So when somebody defineds `https://example.com/myApp1` it does not end with a slash but, defining `https://example.com/myApp1/` it does end.
Inside #568 we want to provide includes without the base URI, so easier to read. But PDS implementations should not need have logic inside to determine if the baseURI ends with a slash or not.
So with #568  we will need to ensure the given PDS URI does not end with a slash - so PDS scripts can rely on the data given from Secub ENV entry.

### Wanted
 `PDS_SCAN_TARGET_URL `  contains always a URL string which does ALWAYS not end with slash

### Solution
Change `PDSWebScanConfigImpl` accordingly and implement some tests

Vivek-Prajapatii:
Interested in it

Jeeppler:
@Vivek-Prajapatii go for it. In case, you have any questions feel free to ask.

Vivek-Prajapatii:
ok i understood! 
So i need to append a code which will remove the slash(/) from the end of the URI, if present.
did i understood in the right way, if not kindly ping me.

Jeeppler:
@Vivek-Prajapatii yes, that is exactly right. 

A URL which looks like this: `https://example.org/` -> will become -> `https://example.org`. 
Please write the code and a test. We use Junit4 and Junit5 in our code base. Please use [Junit5](https://junit.org/junit5/) for new tests.

There are currently 2500+ tests. Most of them are unit tests (testing the functionality of a class or method). However, a few are integration tests (testing parts or the entire system). In your case writing a unit test is enough.

Please consider valid and non valid test cases. For example, the following should work:

1. `https://example.org/` -> `https://example.org`
2. `https://example.org/#` -> `https://example.org`
3. `http://example.org/` -> `http://example.org`
4. `http://example.org/#` -> `http://example.org`
5. `https://example.org/?search=abc` ->  `https://example.org`
…

[...]

PR #964, opened 2022-01-29
Target url never ends with a slash

Vivek-Prajapatii:
i've tested the code manually, as i'm new to junit
if something goes wrong kindly ping me.

closes: #899

### DIFF
--- a/sechub-adapter-pds/src/main/java/com/daimler/sechub/adapter/pds/PDSWebScanConfigImpl.java
+++ b/sechub-adapter-pds/src/main/java/com/daimler/sechub/adapter/pds/PDSWebScanConfigImpl.java
@@ -95,7 +95,7 @@ public class PDSWebScanConfigImpl extends AbstractWebScanAdapterConfig implement
            	}
            	catch(Exception e)
  			{
- 				
+ 				throw new IllegalArgumentException(PARAM_KEY_PDS_SCAN_TARGET_URL + " is not a valid URL ",e);
  			}
  			
             jobParameters.put(PDSDefaultParameterKeyConstants.PARAM_KEY_PDS_SCAN_TARGET_URL, config.getTargetAsString());
--- a/sechub-adapter-pds/src/main/java/com/daimler/sechub/adapter/pds/PDSWebScanConfigImpl.java
+++ b/sechub-adapter-pds/src/main/java/com/daimler/sechub/adapter/pds/PDSWebScanConfigImpl.java
@@ -82,6 +82,22 @@ public class PDSWebScanConfigImpl extends AbstractWebScanAdapterConfig implement
         @Override
         protected void customBuild(PDSWebScanConfigImpl config) {
             jobParameters.put(PDSDefaultParameterKeyConstants.PARAM_KEY_PDS_SCAN_TARGET_TYPE, config.getTargetType());
+            
+            // to get the base URl from the URL
+            // https://example.org/abc -> https://example.org
+            
+            try
+            {            
+            	URL url = new URL(PARAM_KEY_PDS_SCAN_TARGET_URL);
+				String path = url.getFile().substring(0, url.getFile().lastIndexOf('/'));
+				String baseUrl = url.getProtocol() + "://" + url.getHost() + path;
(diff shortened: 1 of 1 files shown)


de-jcup:
As mentioned inside #899 we no longer handle this at SecHub site but instead product specific.
But I will add my thoughts for future PRs.

sechub-adapter-pds/src/main/java/com/daimler/sechub/adapter/pds/PDSWebScanConfigImpl.java
```diff
@@ -82,6 +82,22 @@ public final PDSWebScanConfigBuilder setJobParameters(Map<String, String> jobPar
         @Override
         protected void customBuild(PDSWebScanConfigImpl config) {
             jobParameters.put(PDSDefaultParameterKeyConstants.PARAM_KEY_PDS_SCAN_TARGET_TYPE, config.getTargetType());
+            
+            // to get the base URl from the URL
+            // https://example.org/abc -> https://example.org
+            
+            try
```
de-jcup, line 89: We have a dedicated source code format  - see https://github.com/mercedes-benz/sechub/wiki/Collaboration#source-formatter . It can be imported into eclipse . After this you can simply press `CTR` + `SHIFT` + `f` inside editor and your source is formatted as expected.

sechub-adapter-pds/src/main/java/com/daimler/sechub/adapter/pds/PDSWebScanConfigImpl.java
```diff
@@ -82,6 +82,22 @@ public final PDSWebScanConfigBuilder setJobParameters(Map<String, String> jobPar
...
+            {            
+            	URL url = new URL(PARAM_KEY_PDS_SCAN_TARGET_URL);
```
de-jcup, line 91: The logic does not work here.

`PARAM_KEY_PDS_SCAN_TARGET_URL` is just a constant and represents a key which will be later available as an ENV entry inside PDS runtime.

`config.getTargetAsString());` does contain the URL as a simple string.
So if we would implement the behaviour - we no longer need the feature - we would simply do some string operations here.

sechub-adapter-pds/src/main/java/com/daimler/sechub/adapter/pds/PDSWebScanConfigImpl.java
```diff
@@ -82,6 +82,22 @@ public final PDSWebScanConfigBuilder setJobParameters(Map<String, String> jobPar
...
+				String path = url.getFile().substring(0, url.getFile().lastIndexOf('/'));
+				String baseUrl = url.getProtocol() + "://" + url.getHost() + path;
+				PARAM_KEY_PDS_SCAN_TARGET_URL = baseUrl;
+           	}
+           	catch(Exception e)
+ 			{
+ 				throw new IllegalArgumentException(PARAM_KEY_PDS_SCAN_TARGET_URL + " is not a valid URL ",e);
```
de-jcup, line 98: Throwing an exception here would be a hard fail - so if a user just defines it wrong the complete job would crash and an unexpected error would be returned.

But maybe this was not clear enough specified inside the issue.

sechub-adapter-pds/src/main/java/com/daimler/sechub/adapter/pds/PDSWebScanConfigImpl.java
```diff
@@ -82,6 +82,22 @@ public final PDSWebScanConfigBuilder setJobParameters(Map<String, String> jobPar
...
+ 			}
+ 			
             jobParameters.put(PDSDefaultParameterKeyConstants.PARAM_KEY_PDS_SCAN_TARGET_URL, config.getTargetAsString());
```
de-jcup, line 101: Just for information: We put here into jobParameters for the key `PARAM_KEY_PDS_SCAN_TARGET_URL` the config target string. The framework will automatically inject this one at PDS side as an ENV entry

de-jcup (commented):
See my comments. But you have no additional work to do, because we do not longer need the feature. See my comments at the issue.

PR #966, opened 2022-01-31
Feature 106 create job after zipping

### LINKED ISSUE #106: Client must create only a new SecHub job when client operations are successful
## Problem description
If zip fails (e.g. wrong user rights in directory) there is no zip file.
In this case we got something like
```bash
- Creating new sechub job
2019/12/05 09:54:14 open sourcecode-a36079ec-bc3c-462b-a876-0a2f00d2edbc.zip: no such file or directory
```
But the job is already created - so  we got a waiting Zombie job here...

## Solution
Before job is created, the zip file must be created and checked if available or not. After this has been successfully done, the SecHub job shall be created etc.

sven-dmlr:
Changes are basically that the SecHub job gets created after the generation of the local .zip file succeeded.

closes #106


Jeeppler (approved):
Very good. More readable. 🥇

PR #973, opened 2022-02-02
Feature 290 client reduce json payload

### LINKED ISSUE #290: Reduce JSON payload in client: Only send config parts at job creation time
We need most configuration stuff only at job creation time.

To
- check status
- mark/unmark false positives
- cancel
- etc. etc.

The configuration parts (codeScan,webScan,...) are not necessary.

But at the moment the client does always send the complete stuff in every call.

This should be improved

---

sven-dmlr:
- send json only when creating a SecHub job
- refactoring of complex `waitForSecHubJobDoneAndFailOnTrafficLight`/`getSecHubJobState` into small simple functions
(Please look at complete file `job.go` because the diffs are rather confusing)

closes #290


Jeeppler (approved):
Looks good.

PR #979, opened 2022-02-04
Feature 977 client status check before download

### LINKED ISSUE #977: Client: Check status before downloading report
# Current situation
The client tries to download the report regardless the status.

# Wanted
- The client checks the `result` field of the job status and only downloads the report on "OK" value
- Issue an error message if `result` field is not "OK" and do not try to download the report

Code improvement: make `jobStatusResult` part of the `Context` struct so this information is generally available.

sven-dmlr:
- check status before download
- exit on non-OK with error message
- simplified code by including jobStatusResult into Context struct

closes #977


Jeeppler (approved):

winzj (approved):
Please link the issue inside the pull request description.

PR #983, opened 2022-02-09
Avoid circular reference problem on container start, fixes #944

### LINKED ISSUE #944: SpringBoot 2.6: Fix the circular reference issues allow-circular-references: true
# Problem

SpringBoot 2.6 prohibits circular references between beans by default: https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.6-Release-Notes#circular-references-prohibited-by-default.  The workaround is to set `allow-circular-references: true` in the `sechub-server/src/main/resources/application.yml`. The workaround works, but is far from ideal.

This problem became visible while working on #691.

## Details
On Spring Boot Application Start of `sechub-server` (it does not happen at `PDS`start) we have following output inside console:
```
2022-02-09 10:36:33.445 ERROR 27969 --- [           main] o.s.b.d.LoggingFailureAnalysisReporter   : 

***************************
APPLICATION FAILED TO START
***************************

Description:

The dependencies of some of the beans in the application context form a cycle:

   domainMessageService defined in file [/home/albert/develop/projects/sechub/sechub-review-open-source/sechub-shared-kernel/bin/main/com/daimler/sechub/sharedkernel/messaging/DomainMessageService.class]
      ↓
   scanMessageHandler (field com.daimler.sechub.domain.scan.project.ScanProjectConfigAccessLevelService com.daimler.sechub.domain.scan.ScanMessageHandler.projectAccessLevelService)
┌─────┐
|  scanProjectConfigAccessLevelService (field com.daimler.sechub.domain.scan.project.ScanProjectConfigService com.daimler.sechub.domain.scan.project.ScanProjectConfigAccessLevelService.scanProjectConfigService)
↑     ↓
|  scanProjectConfigService (field com.daimler.sechub.domain.scan.ScanAssertService com.daimler.sechub.domain.scan.project.ScanProjectConfigService.scanAssertService)
↑     ↓
|  scanAssertService (field com.daimler.sechub.domain.scan.project.ScanProjectConfigAccessLevelService com.daimler.sechub.domain.scan.ScanAssertService.accessLevelService)
└─────┘


Action:

Relying upon circular references is discouraged and they are prohibited by default. Update your application to remove the dependency cycle between beans. As a last resort, it may be possible to break the cycle automatically by setting spring.main.allow-circular-references to true.


```

# Solution

Fix the circular references issue and remove the `allow-circular-references: true` in `sechub-server/src/main/resources/application.yml`.

de-jcup:
- some assert services where used by caller side but reuse them as well
- so it was not clear for the framework which one would be
  initialized first.
- To avoid this, the assert services get the callers injected lazy, so
  only necessary at runtime when instances are already created

This PR closes #944

### DIFF
--- a/sechub-scan/src/main/java/com/daimler/sechub/domain/scan/ScanAssertService.java
+++ b/sechub-scan/src/main/java/com/daimler/sechub/domain/scan/ScanAssertService.java
@@ -2,6 +2,7 @@
 package com.daimler.sechub.domain.scan;
 
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Lazy;
 import org.springframework.stereotype.Service;
 
 import com.daimler.sechub.domain.scan.access.ScanUserAccessToProjectValidationService;
@@ -21,6 +22,7 @@ public class ScanAssertService {
     @Autowired
     ScanUserAccessToProjectValidationService userAccessValidation;
 
+    @Lazy
     @Autowired
     ScanProjectConfigAccessLevelService accessLevelService;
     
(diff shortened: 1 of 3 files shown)


Jeeppler (approved):
Looks good to me.

PR #997, opened 2022-02-15
Deleting a non existing profile returns now 404, closes #767

### LINKED ISSUE #767: Delete non-existing execution profile must return HTTP 404
Delete execution profile api call always returns HTTP 200 even if the profile name does not exist.

**Example** (against a freshly started local integrationtest server):
```
> sechub-api.sh profile_list
[]
> sechub-api.sh profile_delete abcdef
Execution profile "abcdef" will be deleted. This cannot be undone.
 Are you sure? (y/n) y
HTTP/1.1 200
```

**Expected bahavior:**
HTTP 404 (not found)

de-jcup:


sechub-scan/src/test/java/com/daimler/sechub/domain/scan/product/config/DeleteProductExecutionProfileServiceTest.java
```diff
@@ -0,0 +1,64 @@
...
+        serviceToTest.deleteProductExecutionProfile(profile.id);
+
+        /* test */
+        verify(repository).deleteById(profile.id);
+    }
+    
+    @Test(expected = NotFoundException.class)
+    public void non_existing_profile_cannot_be_deleted_and_throws_not_found_exception() {
+        /* prepare */
+        when(repository.findById("profileid1")).thenReturn(Optional.empty());
```
winzj, line 55: I think `findById("profileid1")` should be replaced with `findById("not-existing")` here, because in the `/* execute */` section "not-existing" is used as profile id.

winzj (changes requested):

sechub-scan/src/test/java/com/daimler/sechub/domain/scan/product/config/DeleteProductExecutionProfileServiceTest.java
de-jcup: Yes - you are right. Even when the mockito framework does return an optional empty always here. At least it does confuse people while reading...

I will change to any() instead so it's absoletely clear that whatever is tried to by find will be a "Optional.empty()" as an result.

de-jcup: Will also change to Junit5 + check that findById is really call.

winzj (approved):

PR #1036, opened 2022-02-21
fix html artifact generation #1035

### LINKED ISSUE #1035: gradle build workflow still uses `docs/latest`
ToDo:
In `gradle.yml`: Take HTML artifacts from `sechub-doc/build/docs/final-html/` instead of `docs/latest`.

sven-dmlr:
- ./gradlew documentation is sufficient
- source HTML directory location fixed

closes #1035


de-jcup (approved):
Looks good.

PR #1048, opened 2022-02-22
Introduce auto cleanup (administration domain) #19

### LINKED ISSUE #19: Automatically remove old scans
### About
To prevent full hard drives there should be a option to automatically remove old scan data.

## Wanted
As an administrator I can configure a "auto cleanup configuration" and after which age a Job and it's data will be automatically dropped.

#### Domain:administration
- [x] REST API for admin domain
- [x] Stores new configuration inside database 
- [x] An event with message id `AUTO_CLEANUP_CONFIGURATION_CHANGED`is triggered containing the calculated days for auto cleaning
- [x] Introduce a `AdministrationConfig` database table like already exist for scheduler : `SchedulerConfig` (adopt behavior, just for administration domain)
- [x] Enhance `AdministrationConfig` and introduce time field  `auto_cleanup_in_days`  - per default 0, means auto cleanup disabled

#### Domain: administration
- [x] REST API for admin domain
- [ ] Open API generation check
- [x] When no cleanup configuration existed before, a default will be returned with "0 month" - so still being disabled

#### Domain: scan
- [x] Rename old `ScanConfig` to `ScanMappingConfig` and all related parts (like services etc.) - will be done with #1059 
- [x] Introduce new `ScanConfig` and adopt like done inside  introduce time field  `auto_cleanup_in_days`  - per default 0, means auto cleanup disabled
- [x] on `AUTO_CLEANUP_CONFIGURATION_CHANGED` the days for cleanup are fetched from event and stored in db
- [x] Introduce `ScanAutoCleanupTriggerService` which is annotated with @schedule and is configurable by SpringValue annotion
- [x]  delete old `scan_product_result` entries 
- [x]  delete old `scan_project_log` entries 
- [x]  we currently do **NOT** delete `scan_report` entries at the moment, this will be done with #1010 because currently we can only use these database rows for statistics

#### Domain: schedule
- [x] on `AUTO_CLEANUP_CONFIGURATION_CHANGED` the days for cleanup are fetched from event and stored in db
- [x] Introduce `ScheduleAutoCleanupTriggerService` which is annotated with @schedule and is configurable by SpringValue annotion
- [x]  drops old schedule_sechub_job entries 

#### Domain: administration
- [x] on `AUTO_CLEANUP_CONFIGURATION_CHANGED` the days for cleanup are fetched from event and stored in db
- [x] Introduce a `AdministrationAutoCleanupConfig` which represents the auto cleanup configuration
      This configuration contains a "remove-jobs-older-than" field which contains a "time-unit" and "amount" entry.
- [x] Introduce `AdministrationAutoCleanupTriggerService` which is annotated with @schedule and is configurable by SpringValue annotion. Does use the `AdministrationConfigService` to fetch data
- [x] `AdministrationAutoCleanupService` : deletes old adm_job_information entries 

## Documentation
- [x] Rest doc
- [x] Use case documentation are descriptive (+ don't forget the technical use cases)
- [x] Short explanation of concept "auto cleanup" inside "concepts"
- [ ] The format of the time unit can be defined in "days", "weeks", "month", "years". This must be documented

## Integration test
- [x]  setup an integration test (difficult but should be possible)

### LINKED ISSUE #1059: Rename (old) `ScanConfig` to `ScanMappingConfiguration`
## Situation
While implementing #19 it came clear that  the existing `ScanConfig`does not fit into `SchedulerConfig` or the introduced `AdministrationConfig` implementations (these are JPA entities and are represented in database by ONE row).

(old) `ScanConfig` is used only for representing the calculated results of a list of `ScanMapping` entries which is currently only a result of name pattern mappings.  Also it is very irritating that this is not the global configuraiton for scan domain at all.

## Wanted
Rename `ScanConfig` to `ScanMappingConfiguration` . So after the renaming we can introduce the "real" `ScanConfig" for #19 

## Solution

- [x] change classes
- [x] update existing documentation where neccessary
- [x] Rest problem check when introducing ' ScanConfig'  inside #19 ?
      NO, not at all. Mappings are as before : `api/admin/config/mapping/{mappingId`

### LINKED ISSUE #1072: Ensure domain message handling is always correct annotated
## Situation
Sometimes developers do not correctly use the messaging annotations:
- IsReceivingAsyncMessage
- IsRecevingSyncMessage
- IsSendingAsyncMessage
- IsSendingSyncMessage

### Problem
- either it does not work (which is normally directly fixed at development time)
- the messaging works ... but **generated messaging diagrams have failures**

## Wanted
A developer shall automatically get noticed about wrong or missing domain message annotations

## Solution 
write a junit test which checks validity

### LINKED ISSUE #1073: Fix missing messaging annotations
With #1072 implementation following documentation problems also occurred:
```
jjava.lang.AssertionError: 
Generated diagram problem:Message SCAN_ABANDONDED:is defined, but never received
Generated diagram problem:Message UNSUPPORTED_OPERATION:is defined, but never received
	at org.junit.Assert.fail(Assert.java:89)
	at com.mercedesbenz.sechub.docgen.messaging.MessagingAnnotationsCorrectDefinedTest.all_message_ids_are_received_at_least_one_time(MessagingAnnotationsCorrectDefinedTest.java:185)
```

### LINKED ISSUE #1076: Test applications (*TestMain) do no longer run in IDEs
## Situation
There exist some simple test applications which are settled inside `/src/test/java`. Eclipse does no longer allow to run these test application because context is now separated like in gradle and so only available for tests

Means, these handy applications can currently not be used any longer

## Wanted
The execution shall be possible again

## Solution
All of these test applications will 
- be transformed to simple junit5 tests 
- will be renamed, so ending always with `ManualTest` instead of `*TestMain`
- run only when `sechub.manual.test.by.developer` is set to `true` (so launch configurations must set system property by `-Dsechub.manual.test.by.developer=true`)
- contain only one test method called `manualTestByDeveloper` so it is clear that a develper must run the "test" manally
- but drop those parts which are no longer relevant

### LINKED ISSUE #1077: Enhance link text for message link list generation
## Situation
Currently the links from domain messages to related use cases are not very descriptive:
image

## Wanted
Link text should be like in headlines:
image

So instead `UC_041`we want a `UC_041-ADMIN_RESTARTS_JOB`

## Solution
Change `UseCaseEventOverviewPlantUmlGenerator` accordingly.

### LINKED ISSUE #1124: Documentation generation failure - scheduling definitions empty
## Situation
Inside technical documentation the scheduling system properties are not visible:
image

## Wanted
The scheduler settings from code must be visible inside generated documentation

## Solution
The currently generated content is not valid asciidoc table content - must be fixed in generator:

Example snippet from `gen_scheduling.adoc`:
```
[options="header",cols="1,1,1,1"]
.Scope 'scheduler'
|===
|Type   |Definition   |Description
//----------------------
|Fixed|initial delay:${sechub.config.trigger.nextjob.initialdelay:5000} fixed delay:${sechub.config.trigger.nextjob.delay:10000}|Job scheduling is triggered by a cron job operation - default is 10 seconds to delay after last execution. The initial delay is 5000 milliseconds is defined. It can be configured different,so when you need to startup a cluster time shifted, simply change the initial delay values in your wanted way.

```
One problem is that the header cols are 4 columns, but output is only 3 columns, so wrong rendered.

de-jcup:
This PR
- closes #19
- closes #1059
- closes #1072 
- closes #1073
- closes #1076
- closes #1077
- closes #1124


sechub-scan/src/main/java/com/mercedesbenz/sechub/domain/scan/log/ProjectScanLog.java
```diff
@@ -58,6 +58,8 @@
     public static final String STATUS_FAILED = "FAILED";
     public static final String STATUS_ABANDONED = "ABANDONDED";
 
+    public static final String QUERY_DELETE_LOGS_OLDER_THAN = "DELETE FROM ProjectScanLog l WHERE l." + PROPERTY_STARTED + " <:cleanTimeStamp";;
```
Jeeppler, line 61:
```suggestion
    public static final String QUERY_DELETE_LOGS_OLDER_THAN = "DELETE FROM ProjectScanLog l WHERE l." + PROPERTY_STARTED + " < :cleanTimeStamp";
```

Furthermore, the difference between the letter `l` and `1` is difficult to identify. To go around this issue, I would prefer `log`. It would then be `ProjectScanLog log WHERE log`…

sechub-scan/src/main/java/com/mercedesbenz/sechub/domain/scan/report/ScanReport.java
```diff
@@ -43,6 +43,9 @@
     /* +............................ JPQL .....................................+ */
     /* +-----------------------------------------------------------------------+ */
     public static final String CLASS_NAME = ScanReport.class.getSimpleName();
+    public static final String PROPERTY_REPORT_STARTED = "started";
+
+    public static final String QUERY_DELETE_REPORTS_OLDER_THAN = "DELETE FROM ScanReport r WHERE r." + PROPERTY_REPORT_STARTED + " <:cleanTimeStamp";
```
Jeeppler, line 48:
```suggestion
    public static final String QUERY_DELETE_REPORTS_OLDER_THAN = "DELETE FROM ScanReport r WHERE r." + PROPERTY_REPORT_STARTED + " < :cleanTimeStamp";
```

sechub-scan/src/test/java/com/mercedesbenz/sechub/domain/scan/ScanMessageHandlerTest.java
```diff
@@ -32,32 +34,50 @@
...
+        injectedAsynchronousHandlers.add(messageHandlerToTest);
         List<SynchronMessageHandler> injectedSynchronousHandlers = new ArrayList<>();
 
         fakeDomainMessageService = new FakeDomainMessageService(injectedSynchronousHandlers, injectedAsynchronousHandlers);
 
     }
 
     @Test
-    public void when_sending_message_id_PROJECT_ACCESS_LEVEL_CHANGED_changeProjectAccessLevel_is_called() {
+    void handler_receiving_auto_cleanup_calls_config_serice_with_message_data() {
```
Jeeppler, line 63:
```suggestion
    void handler_receiving_auto_cleanup_calls_config_service_with_message_data() {
```

Jeeppler (changes requested):
First part of the review.

I was unable to find tests which test negative values for the deletion time. I would like to see a test, testing for `-1` days or months to make sure negative values are not accepted.

sechub-shared-kernel/src/main/java/com/mercedesbenz/sechub/sharedkernel/messaging/AdministrationConfigMessage.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * This message data object contains all necessary information about
+ * administration configuration change which can be interesting for messaging.
+ *
+ * @author Albert Tregnaghi
+ *
+ */
+@JsonIgnoreProperties(ignoreUnknown = true) // we do ignore to avoid problems from wrong configured values!
+@MustBeKeptStable("This configuration is used by communication between (api) scan,report, scheduler and administration")
```
Jeeppler, line 16:
```suggestion
@MustBeKeptStable("This configuration is used by communication between (api) scan, report, scheduler and administration")
```

sechub-shared-kernel/src/main/java/com/mercedesbenz/sechub/sharedkernel/usecases/admin/config/UseCaseAdminFetchesAutoCleanupConfiguration.java
```diff
@@ -0,0 +1,27 @@
...
+import com.mercedesbenz.sechub.sharedkernel.usecases.UseCaseIdentifier;
+
+/* @formatter:off */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+@UseCaseDefinition(
+		id=UseCaseIdentifier.UC_ADMIN_FETCHES_AUTO_CLEANUP_CONFIGURATION,
+		group=UseCaseGroup.CONFIGURATION,
+		apiName="adminFetchesAutoCleanupConfiguration",
+		title="Admin fetchesauto cleanup configuration",
```
Jeeppler, line 21:
```suggestion
		title="Admin fetches auto cleanup configuration",
```

sechub-shared-kernel/src/main/java/com/mercedesbenz/sechub/sharedkernel/usecases/autocleanup/UseCaseAdministrationAutoCleanExecution.java
```diff
@@ -0,0 +1,28 @@
...
+
+/* @formatter:off */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+@UseCaseDefinition(
+		id=UseCaseIdentifier.UC_ADMINISTRATION_AUTO_CLEANUP_EXECUTION,
+		group=UseCaseGroup.TECHNICAL,
+		apiName=APIConstants.NO_API_AVAILABLE,
+		title="Sechub administration domain auto cleanup",
+		description="The administration domain does execute an auto cleanup. Periodically it will be checked if there is any outdated job information. The timeperiod is detrmine")
```
Jeeppler, line 23:
```suggestion
		description="The administration domain does execute an auto cleanup. Periodically it will be checked if there is any outdated job information. The time period is determine")
```

sechub-shared-kernel/src/main/java/com/mercedesbenz/sechub/sharedkernel/usecases/autocleanup/UseCaseScanAutoCleanExecution.java
```diff
@@ -0,0 +1,28 @@
...
+
+/* @formatter:off */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+@UseCaseDefinition(
+		id=UseCaseIdentifier.UC_SCAN_AUTO_CLEANUP_EXECUTION,
+		group=UseCaseGroup.TECHNICAL,
+		apiName=APIConstants.NO_API_AVAILABLE,
+		title="Sechub scan domain auto cleanup",
+        description="The scan domain does auto cleanup old data.This is done periodically.")
```
Jeeppler, line 23:
```suggestion
        description="The scan domain does auto cleanup old data. This is done periodically.")
```

sechub-shared-kernel/src/test/java/com/mercedesbenz/sechub/sharedkernel/TimeCalculationServiceTest.java
```diff
@@ -0,0 +1,45 @@
...
+        systemTime = mock(SystemTimeProvider.class);
+        date_2021_02_01_time_08_15 = LocalDateTime.of(2021, 2, 1, 8, 15);
+
+        serviceToTest.systemTime = systemTime;
+    }
+
+    @NullSource
+    @CsvSource({ "0", "1", "90", "365" })
+    @ParameterizedTest
+    void now_minus_days_calulation_works(Long days) {
```
Jeeppler, line 32: What happens if one deducts so many days, that the result will be negative? Do I get an exception?

sechub-testframework/src/main/java/com/mercedesbenz/sechub/test/TestConstants.java
```diff
@@ -0,0 +1,13 @@
...
+public class TestConstants {
+
+    /**
+     * System property for defining manual test /test applications.
+     */
+    public static final String MANUAL_TEST_BY_DEVELOPER = "sechub.manual.test.by.developer";
+
+    public static final String DESCRIPTION_DISABLED_BECAUSE_A_MANUAL_TEST_FOR_GENERATION = "This test can be used by developers to generate content. "
+            + "Not necesary for normal CI tests. A developer can run the 'test' to produce very fast generated outtput, "
+            + "without starting the spring boot application or any gradle task(so much faster)";
```
Jeeppler, line 12:
```suggestion
            + "without starting the SpringBoot application or any Gradle task (so much faster)";
```

Jeeppler (changes requested):
See comments.

de-jcup:
> I was unable to find tests which test negative values for the deletion time. I would like to see a test, testing for `-1` days or months to make sure negative values are not accepted.

Thank you, this was not only missing tests, but an implementation failure.

I added 
- `TimeCalculationServiceTest#now_minus_days_calulation_works` 
  (the calculation service WILL ADD in case of negative values, it's responsible for general calculation and shall be reusable)
- `AdministrationAutoCleanupServiceTest#cleanup_executes_NOT_delete_job_information_for_minus_1_day`
  (also changed implementation, because there was no check)
- and some tests in `AdministrationConfigServcieTest` also about minus days - those configurations may not be accepted
- I also changed implementation parts in update service but also in execution time, so problem can not happen any longer

sechub-doc/src/main/java/com/mercedesbenz/sechub/docgen/reflections/Reflections.java
```diff
@@ -211,7 +212,7 @@ void fetchJavaClassNamesToInspect(File sourceDirectory, File file, JavaContentFi
                 clazz = Class.forName(className);
                 classesToInspect.add(clazz);
             } catch (ClassNotFoundException e) {
-                LOG.warn("Ignored class:{}", className);
+                LOG.trace("Class not found in sechub-doc classpath, only on source code:{}", className);
```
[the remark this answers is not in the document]
de-jcup: I added a comment inside code + changed the log output, so it is more clear what is meant. Please look inside - I think it is now more clear.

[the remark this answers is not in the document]
de-jcup: I also changed back from trace to warn - maybe we should (inside another issue) - change the classpath of `sechub-doc`.

```
10:26:55.715 [main] WARN com.mercedesbenz.sechub.docgen.reflections.Reflections - The class 'com.mercedesbenz.sechub.analyzer.cli.SecHubAnalyzerApplication' was not found in ...
```
Some of the warnings can be ignored, but when we add some new sub projects and the documentation does not contain it, we should have at least a log output for debugging...

sechub-doc/src/test/java/com/mercedesbenz/sechub/docgen/messaging/MessagingAnnotationsCorrectDefinedTest.java
```diff
@@ -0,0 +1,338 @@
...
+    public void all_message_ids_are_received_at_least_one_time() throws Exception {
+        /* prepare */
+        Set<MessageID> allSentMessages = new HashSet<>();
+        allSentMessages.addAll(context.messagesReceivedAsynchronous);
+        allSentMessages.addAll(context.messagesReceivedSynchronous);
+
+        /* test */
+        for (MessageID message : MessageID.values()) {
+            if (!allSentMessages.contains(message)) {
+                if (context.messagesAnswerResultSynchronous.contains(message)) {
```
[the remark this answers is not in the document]
de-jcup: It was not clear for myself if the changes you proposed would lead to exact same test - but nevertheless I think the test is not very readable (even for myself now).

So I decided to change the test and make it more readable - and remove also the `continue`. Please look at the changes.

sechub-integrationtest/src/main/java/com/mercedesbenz/sechub/integrationtest/api/TestAPI.java
```diff
@@ -997,4 +1010,93 @@ public static void waitProjectDoesNotExist(TestProject project) {
...
+    }
+
+    private static void resetIntegrationTestAutoCleanupInspector() {
+        String url = getURLBuilder().buildIntegrationTestResetAutoCleanupInspectionUrl();
+        getSuperAdminRestHelper().post(url);
+    }
+
+    /**
+     * Will ensure auto cleanup configuration is set to given days and that every
+     * domain is synched.
```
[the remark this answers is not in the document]
de-jcup: I prefer synched. Will keep it as is.

sechub-integrationtest/src/main/java/com/mercedesbenz/sechub/integrationtest/api/TestAPI.java
```diff
@@ -997,4 +1010,93 @@ public static void waitProjectDoesNotExist(TestProject project) {
...
+
+        if (!wanted.equals(autoCleanupConfig)) {
+            /* turn off by setting to 0 days */
+            TestAutoCleanupData data = new TestAutoCleanupData(days, TestCleanupTimeUnit.DAY);
+
+            as(SUPER_ADMIN).updateAutoCleanupConfiguration(data);
+
+            waitUntilAutoCleanupConfigurationChangedTo(data);
+        }
+        waitUntilEveryDomainHasAutoCleanupSynchedToDays(days);
```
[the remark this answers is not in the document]
de-jcup: Will keep ch

sechub-integrationtest/src/test/java/com/mercedesbenz/sechub/integrationtest/scenario1/AutoCleanupScenario1IntTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+        /* prepare */
+        resetAutoCleanupDays(0);
+
+        /* execute */
+        as(SUPER_ADMIN).updateAutoCleanupConfiguration(new TestAutoCleanupData(1, TestCleanupTimeUnit.DAY));
+
+        /* test */
+        /* @formatter:off */
+        waitUntilEveryDomainHasAutoCleanupSynchedToDays(1);
```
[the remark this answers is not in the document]
de-jcup: will keep

sechub-integrationtest/src/test/java/com/mercedesbenz/sechub/integrationtest/scenario1/AutoCleanupScenario1IntTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+        /* @formatter:on */
+
+    }
+
+    @Test
+    public void auto_cleanup_NEVER_executed_in_any_domain_when_admin_configures_cleanupdays_0() {
+
+        /* prepare */
+        resetAutoCleanupDays(4711);// so different to 0 which we will set next
```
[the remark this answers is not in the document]
de-jcup: We have spotless and I rely on spotless. 
Such things like a space between anything for code not being inside "formatter off" region will not be fixed by me any longer.
So I will NOT do this.

If is possible to change the eclipse formatter settings this way and to provide this in spotless via the formatter xml - fine - lets do it, but I will not do any such manual formatting changes.

sechub-shared-kernel/src/test/java/com/mercedesbenz/sechub/sharedkernel/TimeCalculationServiceTest.java
de-jcup: The time calculation service does not know that negative values are not acceptable. it will add them.

I added tests to ensure the behaviour.

I changed the implementation inside configuration update mechanism and also inside the execution + added tests for this.
So
- it cannot be configured with negative values by admins
- even when somebody changes value in database, the execution will just log an error but not accept it any longer.

Thanx for finding that.

sechub-testframework/src/main/java/com/mercedesbenz/sechub/test/TestConstants.java
de-jcup: I changed to "Spring Boot" and "Gradle" 

but our suggestion "SpringBoot" is not correct. They write it not this way: see https://spring.io/projects/spring-boot

sechub-schedule/src/main/java/com/mercedesbenz/sechub/domain/schedule/SchedulerJobBatchTriggerService.java
```diff
@@ -90,9 +90,10 @@ protected void postConstruct() {
     }
 
     // default 10 seconds delay and 5 seconds initial
-    @MustBeDocumented("Job scheduling is triggered by a cron job operation - default is 10 seconds to delay after last execution. " + "The initial delay is "
-            + DEFAULT_INITIAL_DELAY_MILLIS + " milliseconds is defined. It can be configured different,so when you need to startup a cluster "
-            + "time shifted, simply change the initial delay values in your wanted way.")
+    @MustBeDocumented(value = "Job scheduling is triggered by a cron job operation - default is 10 seconds to delay after last execution. "
+            + "The initial delay is " + DEFAULT_INITIAL_DELAY_MILLIS
+            + " milliseconds is defined. It can be configured different,so when you need to startup a cluster "
```
Jeeppler, line 95: The second part is difficult to understand. How about:

It can be configured differently. This is useful when you need to startup a cluster. Simply change the initial delay values in to allow the cluster to startup.

Jeeppler (changes requested):
See comments.

sechub-schedule/src/main/java/com/mercedesbenz/sechub/domain/schedule/SchedulerJobBatchTriggerService.java
de-jcup: You are right about the last sentence.
- I also fixed here the grammar issue like done before for `AutoCleanupConstants#TRIGGER_STEP_MUST_BE_DOCUMENTED` .
- I also fixed the last sentence in `AutoCleanupConstants#TRIGGER_STEP_MUST_BE_DOCUMENTED` .

sechub-doc/src/main/java/com/mercedesbenz/sechub/docgen/spring/ScheduleDescriptionGenerator.java
```diff
@@ -57,7 +57,7 @@ public String generate(ClasspathDataCollector collector) {
         for (Map.Entry<String, SortedSet<TableRow>> entries : rowMap.entrySet()) {
             SortedSet<TableRow> table = entries.getValue();
 
-            sb.append("[options=\"header\",cols=\"1,1,1,1\"]\n");
+            sb.append("[options=\"header\",cols=\"1,1,1\"]\n");
```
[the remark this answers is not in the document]
de-jcup: When you do not align columns they are always same size: 

image

Currently we do more or less the same when using "cols=1,1,1":
image

If we would keep this, we could really remove the cols definition - your are right.

But I think wasting here so much space is not good here: "Type" is always only some small text but column has same size and the others need more .
So I have changed the cols attribute now to "cols=1,6,6" and it looks much better (and it makes sense to use cols here):
image

Jeeppler (approved):
Looks good to me.

PR #1080, opened 2022-02-28
Use gradle 7.4, make gradle 8.0 compatible + update spring boot 2.64 + dependency updates

### LINKED ISSUE #495: Switch to Gradle 7.4
## Situation
Currently we cannot update to gradle 7.x

### Reasons
Prepare for Gradle 7.0. Gradle 7.0 will have several breaking changes:

~~~
Deprecated Gradle features were used in this build, making it incompatible with Gradle 7.0.
Use '--warning-mode all' to show the individual deprecation warnings.
See https://docs.gradle.org/6.7.1/userguide/command_line_interface.html#sec:command_line_warnings
~~~

To show where changes are necessary use:

~~~
./gradlew --warning-mode all
~~~

## Wanted
We want to use gradle version 7.4 and have an updated gradle wrapper as well

## Solution
- Apply necessary changes in #1085 and after this, setup the gradle version to 7.4 
- Also update the gradle wrapper parts to 7.4

### LINKED ISSUE #1057: Upgrade SpringBoot to 2.6.4
Upgrade to SpringBoot 2.6.3

- https://spring.io/blog/2022/01/20/spring-boot-2-6-3-is-now-available
- https://github.com/spring-projects/spring-boot/releases/tag/v2.6.3

### LINKED ISSUE #1081: Integrate owasp dependency check gradle plugin in SecHub build
## Situation
At the moment SecHub does not provide SCA tooling. So it cannot scan itself here.

## Wanted
To evaluate a later SCA-configuration possibility via PDS + OWASP dependency check we integrate it inside our current gradle build.
So developers can check via commandline for CVEs etc.

## Solution
Integrate https://plugins.gradle.org/plugin/org.owasp.dependencycheck into sechub build so we can use it as long as there is no sechub SCA integration itself.

### LINKED ISSUE #1082: Improve library dependency handling
## Situation
As mentioned inside https://github.com/mercedes-benz/sechub/pull/1080

> It is not enough to just upgrade the SpringBoot version. SpringBoot 2.6.4 upgraded a number of packages: https://github.com/spring-projects/spring-boot/releases/tag/v2.6.4. Some of the packages like `slf4j` are explicitly mentioned in `libraries.gradle`. The problem is, we have to upgrade the dependencies in `libraries.gradle` manually, otherwise we include the same dependencies multiple times.

So we should always use the spring boot library dependencies where possible.

## Wanted
It must be very much easier to apply a new spring boot version - without manually updating any library(if possible)

### Spring Boot dependencies
The spring boot dependency management gradle plugin does already manage all spring boot dependencies automatically. To have same libraries in gradle sub projects, where we have no spring boot  dependency plugin available, we use now `dependagen`:  https://www.github.com/de-jcup/dependagen

At the moment (v0.2.0) `dependagen`
- must be checked out from GIT locally
- set the spring boot version to use inside template (if necessary)
- start the main class as described in `README.md`
- copy the generated `spring_boot_dependagen.gradle` to `$rootFolder/gradle/spring_boot_dependagen.gradle`

### Own additional ibraries
Here we use https://github.com/ben-manes/gradle-versions-plugin to check for newer versions available.
We still decide manually if we want to update or not!

Just call:
```
./gradlew dependencyUpdates
```

### LINKED ISSUE #1085: Make build gradle 7.0 compatible
## Situation
`compile` and  `testCompile` are no longer supported by gradle version >=7.0 and lead to your current build with gradle 6.x to warnings.

## Wanted
no warnings and ready to switch to gradle >=7.0

## Solution
Change build scripts

### LINKED ISSUE #1093: Make build  gradle 8.0 compatible
## Situation
After changes done in #1085 there were warnings about deprecation warnings for 8.0. 

## Wanted
No warnings

## Solution
Fix them as well.

### LINKED ISSUE #1119: Replace `Pair` with `Map`+ remove apache.commons.lang3 dependency
## Situation
The only usage of apache commons lang3 library is inside netsparker adapter. 

## Wanted
Remove this dependency

## Solution
`NetsparkerAdapterWebLoginSupportV1` uses "Pair" to generate json output later. This can be done by a `TreeMap` as well (so keys are sorted)

de-jcup:
This PR 
- closes #1057 1057
- closes #1081 
- closes #1085
- closes #495
- closes #1093
- closes #1082
- closes #1119


Jeeppler (changes requested):
It is not enough to just upgrade the SpringBoot version. SpringBoot 2.6.4 upgraded a number of packages: https://github.com/spring-projects/spring-boot/releases/tag/v2.6.4. Some of the packages like `slf4j` are explicitly mentioned in `libraries.gradle`. The problem is, we have to upgrade the dependencies in `libraries.gradle` manually, otherwise we include the same dependencies multiple times.

libraries.gradle
```diff
@@ -9,116 +9,135 @@
...
-        junit5_impl: 'org.junit.jupiter:junit-jupiter-api:${libraryVersion.junit5Version}',
-        junit5_runtime: 'org.junit.jupiter:junit-jupiter-engine:${libraryVersion.junit5Version}',
-        junit5_params: 'org.junit.jupiter:junit-jupiter-params:${libraryVersion.junit5Version}',
+        /* spring */
+        springboot_starter_thymeleaf:          "org.springframework.boot:spring-boot-starter-thymeleaf",
+
+        springboot_devtoolssf:                 "org.springframework.boot:spring-boot-devtools",
+
+        springboot_starter_data_jpa:           "org.springframework.boot:spring-boot-starter-data-jpa",
+        springboot_starter_data_mongodb:       "org.springframework.boot:spring-boot-starter-data-mongodb",
```
Jeeppler, line 77: Why do we need the MongoDB dependency?

Jeeppler (approved):
Looks good. However, I found one odd dependency: spring boot starter mongodb. Why do we need that?

libraries.gradle
de-jcup: We do not need this - will be removed.

PR #1138, opened 2022-03-14
Changed restdoc tests #1125

### LINKED ISSUE #1125: Documentation generation - REST API description does only contain curl requests
## Situation
The generated REST API description does only contain curl requests - for example:

image

## Wanted
As a developer I want more information about the REST API.

## Analyze
Reading https://docs.spring.io/spring-restdocs/docs/current/reference/html5/ respectively 
https://docs.spring.io/spring-restdocs/docs/current/reference/html5/#documenting-your-api-default-snippets

> By default, six snippets are written:
> 
>     <output-directory>/index/curl-request.adoc
> 
>     <output-directory>/index/http-request.adoc
> 
>     <output-directory>/index/http-response.adoc
> 
>     <output-directory>/index/httpie-request.adoc
> 
>     <output-directory>/index/request-body.adoc
> 
>     <output-directory>/index/response-body.adoc

These are the default, but we have not 
- "request-fields.adoc"
- "resopnse-fields.adoc" 

files generated per default

Reading https://docs.spring.io/spring-restdocs/docs/current/reference/html5/#configuration
the "request-fields.adoc" files should still be generated when defined.

Reading 
https://docs.spring.io/spring-restdocs/docs/current/reference/html5/#configuration-default-snippets
it seems the configuration must be changed.

### LINKED ISSUE #1140: Improve REST documentation output
## Situation
While  fixing #1125 it became clear that the output of the REST documentation still needs some improvements

## Wanted
- Example and definition parts of REST documentation must be divided into separate blocks
- The definition shall contain a "general request information" without any example data inside, but shows request path, method and returned status code at one glance (inside a table)
- The definition shall contain also parameter/field descriptions
- The example shall contain
   - curl example (as before)
   - response body (except where too big) or if empty an entry with _(empty)_  instead, to show that the expected result is empty
- The different variants shall have better names for the headlines.
  _(instead of "GET" wie use "GET variant", for "JSON" we use "JSON variant" etc.)_

### LINKED ISSUE #1141: AsciidocGenerator not working correctly when called inside IDE
## Situation
When a developer start `AsciidocGenerator.java` directly from IDE (eclipse) the output is not like done in gradle build.

## Wanted
As a developer I want to be able to start the `AsciidocGenerator` from my IDE.

## Analyze
Eclipse does (now) no longer provide test classes and resources to runtime classpath when "normal" application code has been started (which is IMO an improvement and a correct behavior). But this leads to the situation, that the annotated RestDocTests and all other meta data from tests no longer is accessible this way.

## Solution
We write a `AsciidocGeneratorManualTest` which calls the generator. So it has access to test data again and will be not executed by normal builds.
_See `TestConstants.java` for details_

de-jcup:
This PR will
- close #1125 
- close #1140
- close #1141

### DIFF
--- a/projects.gradle
+++ b/projects.gradle
@@ -98,6 +98,7 @@ projectType = [
             ]
     ]
 }
+
 /* dynamically define java projects */
 projectType.springBootProjects.addAll(projectType.springBootSecHubServerProjects)
 projectType.springBootProjects.addAll(projectType.springBootPDSProjects)
--- a/sechub-doc/build.gradle
+++ b/sechub-doc/build.gradle
@@ -300,7 +300,7 @@ task generateOpenapi() {
     	//
     	// The steps are:
     	//
-    	// 1. read in the OpenAPI file
+    	// 1. read in the generated OpenAPI file
     	// 2. get the values which contain a `oneOf` element
     	// 3. overwrite the `oneOf` with `string`
     	// 4. write the modified JSON structure back into the generated OpenAPI 3 file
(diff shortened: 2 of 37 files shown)


sven-dmlr (approved):
Looks good.
The generated api documention looks pretty now (pdf and html) :+1:

PR #1143, opened 2022-03-15
Make temporary directory configurable #467

### LINKED ISSUE #467: Client: Make temporary directory configurable
## client: Scan action will store zip on disk which may cause permission denied
At the moment the "code scan" action zips the code-folder before the upload. If the current user doesn't have write access to this directory (e.g. in a CI context) the check will fail because of a permission denied error. 

We suggest three possible solutions: 
- Zipping to "temp" directory
- Zipping file without writing to disk
- Create parameter to configure Zip output directory

---
<sup>Paul Romahn <paul.romahn@daimler.com>, Daimler TSS GmbH,  [imprint](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)</sup>

sven-dmlr:
- passing via cmdline argument added
- passing via env var added
- unit tests
- verification if the dir exists and is a dir
- documentation update
- documentation wording improvements
- zipfile renamed because at creation time, no jobUUID exists yet

closes #467

### DIFF
--- a/sechub-cli/src/mercedes-benz.com/sechub/cli/config_test.go
+++ b/sechub-cli/src/mercedes-benz.com/sechub/cli/config_test.go
@@ -5,7 +5,10 @@ package cli
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 	"testing"
+
+	sechubTestUtil "mercedes-benz.com/sechub/testutil"
 )
 
 func TestConfigByFlagsThrowsNoError(t *testing.T) {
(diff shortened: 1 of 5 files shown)


sechub-cli/src/mercedes-benz.com/sechub/cli/config_test.go
```diff
@@ -160,3 +163,54 @@ func Example_normalizeCMDLineArgs() {
 	// [./sechub -configfile my-sechub.json -wait 30 scan]
 	// [./sechub -version]
 }
+
+func Example_tempFile() {
```
Jeeppler, line 167: Those are actually two test cases. One config is configured to use "." the other "/tmp/my_dir". I would prefer two functions for that. For each test case its own function. Something like:

~~~
Example_tempFile_current_dir() {}
Example_tempFile_tmp_dir() {}
~~~

This would be more in tune with the way we write our test cases in the Java part. In addition, it is much easier to read and understand if it would be two cases.

sven-dmlr: Okay. Split into two separate tests.

Jeeppler (approved):

PR #1160, opened 2022-03-24
file uploads are now done via io.Pipe #933

### LINKED ISSUE #933: Client: High RAM-usage while upload
In my project we did an cov-build that yielded an emit-db with ~12.3GB.
When i do an `sechub scan`, the RAM usage jumps to 27+GB while the upload step is running.

<sup>Florian Dobisch<florian.dobisch@daimler.com>, Daimler TSS GmbH, [Provider Information](https://github.com/Daimler/daimler-foss/blob/master/PROVIDER_INFORMATION.md)</sup>

sven-dmlr:
- file content is no more buffered in-memory
- now very low memory footprint

closes #933


Jeeppler (approved):
Good job.

PR #1212, opened 2022-04-05
Handle problems with missing body in server errors

### LINKED ISSUE #1206: The error message of a `ForbiddenException` is not returned in body but body is empty
## Situation
When a project is deactivated (via project access level) , a `ForbiddenException` is thrown and the caller gets a `403 Forbidden` as expected, but the body is empty instead of the given exception error message.

## Wanted
The message of the exception must be available inside the body / the body may not be empty

## Analyze
There is something really wrong here. The exception handling is no longer working as expected (and designed for SecHub).

### What happens in log files
After reading some logs we have the situation that the forbidden exception was logged correctly with 

HTTP status code: 403,
JSON:
```
"status" : 403,

"error" : "Forbidden",

"message:" : "Project xyz does currently not allow write access.",

"details" : []
```

After a dedicated timestamp we got 
HTTP status code: 403,
JSON:
```
```
so body is empty.

### Which version does it correctly, which wrong?
After further log evaluation it became clear that 
- v0.28.0 did log correctly (so JSON filled)
- newer releases do no longer log correctly (body empty)

### Which changes could be relevant? 
- with `v0.29.0-server` we switched from spring boot `2.5.8` to `2.6.2`
- there were no relevant issues for logging inside `v0.29.0-server` or `v0.29.1-server`

So it seems to be the spring boot update.

## Solution
- [x] analyze and find the underlying problem/reason
- [x] extend test api to be check not only for http status numbers but also for expected JSON  and/or string content
- [x] adopt 2 existing integration tests with new test api parts and check it's failing currently
- [x] adopt all other http status tests with new api and check JSON content
- [x] fix the problem to have wanted behaviour again

de-jcup:
This PR will 
- close #1206


winzj (approved):
Looks good to me.

PR #1233, opened 2022-04-11
fix-1204-rename-admin_config-table-to-adm_config 

### LINKED ISSUE #1204: Rename `admin_config` table to `adm_config`
## Situation
For our administration tables we use the prefix `adm`but we have introduced `admin_config` for #19 which is not correct

## Wanted
Correct name pattern

## Solution
- Write a flyway script that renames the table to `adm_config`
- Change the constant AdministrationConfig.TABLE_NAME

### LINKED ISSUE #1205: Fix flyway undo script 15
## Situation
Currently the `U15__provide_auto_cleanup.sql` has a bug inside and does not work:

image

## Solution
Change to `DROP TABLE if exists` ...

winzj:
This PR
- closes #1205 
- closes #1204


de-jcup:
Good Job :-)

de-jcup (approved):

PR #1260, opened 2022-04-22
Feature 1174 client intelligent polling

### LINKED ISSUE #1174: Client: Intelligent polling
## Current situation:
When doing a `sechub scan`, after uploading, the client will wait 1 minute (or what is defined in `$SECHUB_WAITTIME_DEFAULT`) before next "get status" call.

E.g. GoSec just takes a coupple of seconds until the result is there. So almost a minute would be wasted until the cliend would begin the report download.

## Wanted:
Automatically poll more fequently the first times. Then increas the wait time until `$SECHUB_WAITTIME_DEFAULT` and then keep the intervall.

sven-dmlr:
- client polls in increasing intervals, so results from quick backends (e.g. PDS-GoSec) are available earlier
- maximal poll interval is still the defined wait time (default: 60s)
- unit tests added
- docs + client help output updated
- code readability improved (moved progress dot printing into an extra func)

closes #1174

### DIFF
--- a/sechub-cli/src/mercedes-benz.com/sechub/cli/config.go
+++ b/sechub-cli/src/mercedes-benz.com/sechub/cli/config.go
@@ -163,11 +163,6 @@ func NewConfigByFlags() *Config {
 	// Parse command line options
 	flag.Parse()
 
-	// We use the configFromInit struct already prefilled with defaults and from environment variables
-	oneSecond := 1 * time.Second
-	configFromInit.waitNanoseconds = int64(configFromInit.waitSeconds) * oneSecond.Nanoseconds()
-	configFromInit.timeOutNanoseconds = int64(configFromInit.timeOutSeconds) * oneSecond.Nanoseconds()
-
 	if flagHelp {
 		configFromInit.action = showHelpAction
 	} else if flagVersion {
@@ -260,8 +255,15 @@ func assertValidConfig(config *Config) {
 	// For convenience: lowercase user id and project id if needed
 	config.user = lowercaseOrNotice(config.user, "user id")
 	config.projectID = lowercaseOrNotice(config.projectID, "project id")
+
 	// Remove trailing slash from url if present
 	config.server = strings.TrimSuffix(config.server, "/")
+
+	config.waitSeconds = validateWaitTimeOrWarning(config.waitSeconds)
+	config.waitNanoseconds = int64(config.waitSeconds) * int64(time.Second)
+
+	config.timeOutSeconds = validateTimeoutOrWarning(config.timeOutSeconds)
+	config.timeOutNanoseconds = int64(config.timeOutSeconds) * int64(time.Second)
 }
 
 // isConfigFieldFilled checks if field is not empty or is 'true' in case of boolean type
@@ -369,4 +371,24 @@ func validateOutputLocation(config *Config) bool {
 		}
 	}
 	return true
+}
+
+func validateWaitTimeOrWarning(waittime int) int {
+	// Verify wait time and ensure MinimalWaitTimeSeconds
+	if waittime < MinimalWaitTimeSeconds {
+		sechubUtil.LogWarning(
+			fmt.Sprintf("Desired wait intervall (%d s) is too small. Setting to %d seconds.", waittime, MinimalWaitTimeSeconds))
+		waittime = MinimalWaitTimeSeconds
+	}
+	return waittime
+}
+
+func validateTimeoutOrWarning(timeout int) int {
+	// Verify timeout and ensure MinimalTimeoutInSeconds
+	if timeout < MinimalTimeoutInSeconds {
+		sechubUtil.LogWarning(
+			fmt.Sprintf("Desired timeout (%d s) is too small. Setting to %d seconds.", timeout, MinimalTimeoutInSeconds))
+		timeout = MinimalTimeoutInSeconds
+	}
+	return timeout
 }
--- a/sechub-cli/src/mercedes-benz.com/sechub/cli/config.go
+++ b/sechub-cli/src/mercedes-benz.com/sechub/cli/config.go
@@ -104,7 +104,7 @@ func prepareOptionsFromCommandline(config *Config) {
 	flag.BoolVar(&flagVersion,
 		versionOption, false, "Shows version info and terminates")
 	flag.IntVar(&config.waitSeconds,
-		waitOption, config.waitSeconds, "Wait time in seconds - For status checks of action='"+scanAction+"' and for retries of HTTP calls. Can also be defined in environment variable "+SechubWaittimeDefaultEnvVar)
+		waitOption, config.waitSeconds, "Maximal wait time in seconds - For status checks of action='"+scanAction+"' and for retries of HTTP calls. Can also be defined in environment variable "+SechubWaittimeDefaultEnvVar)
 }
 
 func parseConfigFromEnvironment(config *Config) {
--- a/sechub-cli/src/mercedes-benz.com/sechub/cli/config.go
+++ b/sechub-cli/src/mercedes-benz.com/sechub/cli/config.go
@@ -104,7 +104,7 @@ func prepareOptionsFromCommandline(config *Config) {
 	flag.BoolVar(&flagVersion,
 		versionOption, false, "Shows version info and terminates")
 	flag.IntVar(&config.waitSeconds,
-		waitOption, config.waitSeconds, "Maximal wait time in seconds - For status checks of action='"+scanAction+"' and for retries of HTTP calls. Can also be defined in environment variable "+SechubWaittimeDefaultEnvVar)
+		waitOption, config.waitSeconds, "Maximum wait time in seconds - For status checks of action='"+scanAction+"' and for retries of HTTP calls. Can also be defined in environment variable "+SechubWaittimeDefaultEnvVar)
 }
 
 func parseConfigFromEnvironment(config *Config) {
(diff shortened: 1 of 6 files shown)


sechub-cli/src/mercedes-benz.com/sechub/cli/config.go
```diff
@@ -370,3 +372,23 @@ func validateOutputLocation(config *Config) bool {
 	}
 	return true
 }
+
+func validateWaitTimeOrWarning(waittime int) int {
```
Jeeppler, line 376:
```suggestion
func validateWaitTimeOrWarning(waitTime int) int {
```

sechub-cli/src/mercedes-benz.com/sechub/cli/config_test.go
```diff
@@ -353,3 +353,29 @@ func Test_validateOutputLocation_empty(t *testing.T) {
 	sechubTestUtil.AssertEquals(".", config.outputFolder, t)
 	sechubTestUtil.AssertEquals("", config.outputFileName, t)
 }
+
+func Test_validateWaitTimeOrWarning(t *testing.T) {
+	// PREPARE
+	bigvalue := 10000000
```
Jeeppler, line 359:
```suggestion
	bigValue := 10000000
```

sechub-cli/src/mercedes-benz.com/sechub/cli/constants.go
```diff
@@ -17,13 +19,25 @@ const DefaultTempDir = "."
 
 // DefaultWaitTime - Wait time in seconds.
 // Will be used
-// - for automatic status checks etc. when action=scan
+// - for maximum pause between automatic status checks when action=scan
 // - for pause between retries for failed HTTP calls
 const DefaultWaitTime = 60
 
+// MinimalWaitTimeSeconds - We don't allow intervals shorter than this to protect the SecHub server
+const MinimalWaitTimeSeconds = 4
```
Jeeppler, line 27:
```suggestion
const MinimalWaitTimeSeconds = 3
```

sechub-cli/src/mercedes-benz.com/sechub/cli/job.go
```diff
@@ -53,24 +53,44 @@ func waitForSecHubJobDone(context *Context) (status jobStatusResult) {
...
+// Print progress dots ... 50 dots, then next line...
+func printProgressDot(cursor int) int {
+	if cursor == 0 {
+		// initial identation
+		fmt.Print(strings.Repeat(" ", 29))
+	}
+
+	fmt.Print(".")
+
+	if cursor++; cursor == 50 {
```
Jeeppler, line 78: This statement is strange. What happens here? Is the `cursor` incremented by one and afterwards the equality check `cursor == 50` happens?

```suggestion
        cursor++
	if cursor == 50 {
```

sechub-cli/src/mercedes-benz.com/sechub/cli/constants.go
```diff
@@ -17,13 +19,25 @@ const DefaultTempDir = "."
...
+
+// InitialWaitIntervallSeconds WaitIntervallIncreaseFactor - defines client's polling behaviour:
+// 2s - 3s - 4.5s - 7s - 10s - 15s - 23s - 34s - 51s - 60s - 60s - 60s ...
```
Jeeppler, line 30: Is 4.5s really possible? Casting something to `int64` should return an int value. In addition, why not using a uint64 data type?

sechub-cli/src/mercedes-benz.com/sechub/cli/job.go
```diff
@@ -53,24 +53,44 @@ func waitForSecHubJobDone(context *Context) (status jobStatusResult) {
...
+		time.Sleep(time.Duration(waitNanoseconds))
+		waitNanoseconds = computeNextWaitInterval(waitNanoseconds, context.config.waitNanoseconds)
 	}
 	sechubUtil.PrintIfNotSilent("\n", context.config.quiet)
 
 	return status
 }
 
+// Print progress dots ... 50 dots, then next line...
+func printProgressDot(cursor int) int {
```
Jeeppler, line 70: In case you do not define the size, the underlying CPU architecture/platform will select a size (either 32 bit or 64 bit).

```suggestion
func printProgressDot(cursor uint64) uint64 {
```

sechub-cli/src/mercedes-benz.com/sechub/cli/job.go
```diff
@@ -53,24 +53,44 @@ func waitForSecHubJobDone(context *Context) (status jobStatusResult) {
...
+
+	if cursor++; cursor == 50 {
+		fmt.Print("\n")
+		cursor = 0
+	}
+
+	return cursor
+}
+
+func computeNextWaitInterval(current int64, max int64) int64 {
```
Jeeppler, line 86: Why not using `uint64`? The values should never be negative.

sechub-doc/src/docs/asciidoc/documents/code2doc/client/02_sechub_client.adoc
```diff
@@ -86,9 +86,9 @@ Does scan synchronous which means
  - approve job as being ready to start
  - wait for job done,
  - fetch automatically result report to output folder (html or json)
- - break your build when not `green`
+ - returns non-zero on `red` (so usually your build will break on critical or high findings)
```
Jeeppler, line 89:
```suggestion
 - returns non-zero on `red` (in most cases your build will break on critical or high findings)
```

sechub-doc/src/docs/asciidoc/documents/code2doc/client/02_sechub_client.adoc
```diff
@@ -86,9 +86,9 @@ Does scan synchronous which means
...
 
-TIP: For security reasons always pass your api token via the environment variable `SECHUB_APITOKEN`.
+IMPORTANT: For security reasons always pass your api token via the environment variable `SECHUB_APITOKEN`.
```
Jeeppler, line 91:
```suggestion
IMPORTANT: For security reasons always pass your API-Token via the environment variable `SECHUB_APITOKEN`.
```

Jeeppler (changes requested):
See comments.

sechub-cli/src/mercedes-benz.com/sechub/cli/config.go
```diff
@@ -104,7 +104,7 @@ func prepareOptionsFromCommandline(config *Config) {
 	flag.BoolVar(&flagVersion,
 		versionOption, false, "Shows version info and terminates")
 	flag.IntVar(&config.waitSeconds,
-		waitOption, config.waitSeconds, "Wait time in seconds - For status checks of action='"+scanAction+"' and for retries of HTTP calls. Can also be defined in environment variable "+SechubWaittimeDefaultEnvVar)
+		waitOption, config.waitSeconds, "Maximal wait time in seconds - For status checks of action='"+scanAction+"' and for retries of HTTP calls. Can also be defined in environment variable "+SechubWaittimeDefaultEnvVar)
```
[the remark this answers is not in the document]
sven-dmlr: Thanks for the better wording :+1: 
The blanks around the `+` cannot be added, because the Go linter always removes them when saving in the IDE.

sechub-cli/src/mercedes-benz.com/sechub/cli/config.go
sven-dmlr: changed.

sechub-cli/src/mercedes-benz.com/sechub/cli/config_test.go
sven-dmlr: changed

sechub-cli/src/mercedes-benz.com/sechub/cli/constants.go
sven-dmlr: changed

sechub-cli/src/mercedes-benz.com/sechub/cli/job.go
sven-dmlr: It's valid Go syntax. :)
I've checked the documentation: "A statement can precede conditionals; any variables declared in this statement are available in all branches."
Assignments inside the `if` are most widely used to declare variables that exist only in the scope of the `if..else` structure.
This is here not the case, so moved `cursor++` as proposed.

sechub-cli/src/mercedes-benz.com/sechub/cli/job.go
sven-dmlr: The value of the `cursor` variable is between 0 and 50.
So `uint8` as type is sufficient.
Changed to `uint8`.

sechub-cli/src/mercedes-benz.com/sechub/cli/constants.go
sven-dmlr: Yes: 4.5s is possible. It's converted into nanoseconds.
The `time` functions expect `int64`, so we use it here also.

sechub-cli/src/mercedes-benz.com/sechub/cli/job.go
sven-dmlr: `time.Duration()` expects `int64` and not `uint64`. We avoid conversions when using `int64`.

sechub-doc/src/docs/asciidoc/documents/code2doc/client/02_sechub_client.adoc
sven-dmlr: changed wording

Jeeppler (approved):

PR #1271, opened 2022-04-27
Added missing documentation parts about client only options #1148

### LINKED ISSUE #1148: Document `excludes` and `additionalFilenameExtensions` are client only configuration values
## Situation
At the moment we provide
- excludes
- additionalFilenameExtensions

in sechub configurationfiles. But we only use this at sechub client side. The information about the excludes or additional file name extensions i currently not given to server side.

While implementing #1098 it was unclear how we handle this in future

## Wanted
Clarification/Decision if we keep this "client only" or take it to the server side model as well

de-jcup:
This PR
- closes #1148

### DIFF
--- a/sechub-doc/src/docs/asciidoc/documents/shared/configuration/sechub_config.adoc
+++ b/sechub-doc/src/docs/asciidoc/documents/shared/configuration/sechub_config.adoc
@@ -73,8 +73,7 @@ include::sechub_config_example1_sourcescan_filesystem.json[]
 <1> Define code scan
 <2> Use filesystem - so means upload sources to server
 <3> Upload will contain sources from `gamechanger-android/src/main/java` and `gamechanger-server/src/main/java` and their sub folders
-<4> Client only option - if you use REST directly, you can ignore this. +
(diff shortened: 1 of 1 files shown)


sven-dmlr (changes requested):
IMHO these changes are not needed.

All relevant information is already in the NOTE in line 52ff
```
NOTE: The configuration parts inside `fileSystem` entry are interpreted by
      `{sechub}` client! When you directly use the REST API you do not need
      to define this data but you have to upload sources as a `zip`file by
      your own over REST!
```

To additionally add it here, only brings confusion, because people expect here the explanation of the bullet points.

de-jcup:
> IMHO these changes are not needed.
> 
> All relevant information is already in the NOTE in line 52ff
> 
> ```
> NOTE: The configuration parts inside `fileSystem` entry are interpreted by
>       `{sechub}` client! When you directly use the REST API you do not need
>       to define this data but you have to upload sources as a `zip`file by
>       your own over REST!
> ```
> 
> To additionally add it here, only brings confusion, because people expect here the explanation of the bullet points.

Yes indeed, I missed the "note" parts. I just added the same for <4> as already defined at <5>. 
So I will remove the information in <4> and <5>, so we have the note only .

sven-dmlr (approved):

PR #1288, opened 2022-05-02
Feature 1274 extend sechub sarif

### LINKED ISSUE #1274: Extend sechub-sarif with the "versionControlProvenance" property to support version control information
### Situation
Currently the sechub-sarif model does not support version control information.

### Wanted
 For future PDS solutions that provide information like `branch` or `commitId` it will be useful to support the `versionControlProvenance` property defined inside SARIF 

### Solution
Extend the sechub-sarif model by implementing the `versionControlProvenance` property as described here:
https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317497

### LINKED ISSUE #1282: Support taxonomy property in sechub-sarif
### Situation
Currently we get some taxonomy information for the report from the `List<ReportingDescriptorRelationship> relationships` attribute from `ReportingDescriptor.java`

### Wanted
We want to support the SARIF taxonomies property and the taxa property.

### Solution
A taxonomy technically does not need a new object, because from a SARIF perspective it is a `toolComponent object`. The same goes for a taxon, which is represented by a `reportingDescriptor object`. This means taxonomies and taxa is a list of those objects. We can add the needed attributes inside these classes and use them.

See: [taxonomies and taxa](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317536) for more detailed information.

winzj:
This PR:

- closes #1274 
- closes #1282


de-jcup:
Please look into `SarifReportSupportTest.java` and add some tests for the new parts as well. Maybe you have to add a new example file as well.

de-jcup:
I created https://github.com/mercedes-benz/sechub/issues/1299 because of the situatoin with rules and taxonomies and also the missing "clearness" about usage of immutable objects (e.g. automatic created lists) , versus `Optional,` versus `null`

de-jcup (changes requested):
See comments.

de-jcup (approved):
Looks fine for me.

Issue #1285, opened 2022-05-02
Add Job Creation and SPDX report commands to SecHub API script


Jeeppler:
Add the commands for:

- creating a job
- uploading a file
- marking a job as ready
- getting a SPDX report

to the Sechub Api script

PR #1297, opened 2022-05-03
Job crate, upload and approve as well as get SPDX report commands

Jeeppler:
- added job create, upload and approve commands
- added get spdx report commands #1285

closes: #1285

### DIFF
--- a/sechub-developertools/scripts/sechub-api.sh
+++ b/sechub-developertools/scripts/sechub-api.sh
@@ -287,14 +287,14 @@ function sechub_job_create {
   local PROJECT_ID="$1"
   local JSON_FILE="$2"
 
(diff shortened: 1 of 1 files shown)


sechub-developertools/scripts/sechub-api.sh
```diff
@@ -615,6 +659,8 @@ function sechub_user_signup_decline {
 
 # -----------------------------
 # main
+readonly JSON_TYPE='Content-Type: application/json'
```
sven-dmlr, line 662: Here, a variable is introduced that is only used twice.

It is okay to define such a variable, but then every occurrence of `Content-Type: application/json` in the script should be replaced.
Please either remove the variable or replace the literal everywhere with the variable.

sven-dmlr (changes requested):
Looks good overall.
Please see my comment about the `JSON_TYPE` variable.

sechub-developertools/scripts/sechub-api.sh
sven-dmlr: Thought twice about it: 
I'd like it more to stay with the literal and remove the variable.
Then it's easier to compare with the `curl` examples on the documentation and one has not to look up the variable.

PR #1313, opened 2022-05-05
Document data section #1218

### LINKED ISSUE #1149: Change OpenApi definition to a generic api definition inside web scan JSON configuration
## Situation
Currently we can already define open api files inside our JSON. But ... what if we provide more e.g. like `graphApi` ? 
In this case users could define both variants.

## Wanted
Possiblity to define webscan api configuration more general

## Solution
Will be changed in code + documentation
See next comment for details.

### LINKED ISSUE #1163: Define and document the TAR and zip structure for data sections
## Situation
This is (not only) a sub issue of #1166  

When  https://github.com/mercedes-benz/sechub/issues/1154 and https://github.com/mercedes-benz/sechub/issues/1098 are  implemented a user is able to upload multiple binaries, sources etc. give them names and reference the files/folders inside 
the sechub configuration file

But doing this, we must 
- define how those files are stored inside TAR and ZIP files
- think about default names (e.g. inside existing codescan configuration - there is a file system setup)
- think for specialized analyze phase (currently not implemented) behaviour and also https://github.com/mercedes-benz/sechub/issues/1164 
 were we want to filter files
  
## Wanted
- documentation
- clear definition how the structure inside tars and zips looks like

### General Definition
We map the JSON model into the file system by using always following path pattern:

```
/__data__/$storageDataType/$dataObjectName/$originPath
```

- `$storageDataType` can be `sources` or `binaries` (like in JSON)
- `$dataObjectName` is just the name defined inside JSON for the data section
- `$originPath` represents the origin path from file system

### Examples
- look at the next comments for examples

### Next steps
Implement

### LINKED ISSUE #1218: Explain `data`-section parts inside documentation
## Situation
With #1098 we introduced the data section. Details are inside the issue but not inside our official documentation.

## Wanted
Documentation shall explain how 
- users can define data in data-sectin
- technical parts in tech doc
- architectural parts in architecture doc

## Solution
Update /add missing documentation

### LINKED ISSUE #1314: TAR and ZIP files must be able to use data reference names without special treatment
## Situation
With #1098 we introduced the `data` configuration element which source an binary configuration elements have unique names which can be referenced.

Currently following is validated:
- min length
- max length
- uniqueness inside model

But the names can be everything.

While writing documentation about `data` in
- #1218 
- #1163 
it became clear that when the names are used for tar and zip structures they
must be valid file pathes as well.

But currently they aren't! If we would use "~/" or "../.../" this would be very problematic...

## Wanted
TAR and ZIP files must be able to use always the unique data reference names 
without any encoding or escaping mechanism.

## Solution
We simply add another validation to the sechub configuration model:
We only accept reference names which do contain `a-z`, `0-9`, `-` or `_`

de-jcup:
This PR
- closes #1149 
- closes #1218 
- closes #1163 
- closes #1314


sven-dmlr (changes requested):
Looks good; please check suggested changes.

PR #1330, opened 2022-05-11
fixed owaspzap wrapper jar version #1329

### LINKED ISSUE #1329: wrapperowaspzap .jar has server version instead of PDS version
## ToDo:
- Change Gradle build files so wrapperowaspzap-<version>.jar has PDS release version
- Change github actions workaround for the same

This will fix the broken Actions run when server and PDS are released at the same time.

sven-dmlr:
closes #1329


haerter-tss (approved):
Looks good.

PR #1338, opened 2022-05-12
Implement Owasp Zap wrapper ability to deactivate scan rules #1294

### LINKED ISSUE #1294: The ZAP wrapper should be able to deactivate certain rules
### Situation
Currently we install and use all scan rules with the release tag for our PDS-OWASPZAP. Some of the rules can produce a lot of alerts and therefore a lot of false-positives. 

### Wanted
We want to be able to deactivate certain rules.

### Solution
The `sechub-wrapper-owasp-zap` shall be able to take a file specifying rules to deactivate.

To specify this file we need a new command line option for the wrapper and maybe a default env variable that is checked, if no file was specified, like  `ZAP_RULE_DEACTIVATION_FILE`. The wrapper needs to know the `type`(passive, active, ...) and the `id` to be able to call the OWASP ZAP API correctly. This file could specify rules in the following way, so it is easier to read than just using a list of IDs.
`owasp-zap-wrapper-config.json:`
```json
{
  "deactivated" : [ {
    "rules" : [ {
      "ref" : "path-traversal-6",
      "info" : "Reason for deactivation."
    } ]
  } ]
}
```

The file below contains the complete set of rules available, so the wrapper can find out the `id` and the `type` using the `ref` from the file above. The `timestamp` and the `origin` are meta information, to determine when the information was last updated or if the location has changed. This file only needs to be  updated if new rules are added or rules are changed with a new release.
`owasp-zap-ruleset.json:`
```json
{
  "timestamp" : "2022-05-02:13:43:33",
  "origin" : "https://www.zaproxy.org/docs/alerts/",
  "rules" : [ {
    "id" : "6",
    "ref" : "path-traversal-6",
    "name" : "Path Traversal",
    "type" : "active",
    "link" : "https://www.zaproxy.org/docs/alerts/6/"
  }, {
    "id" : "3",
    "ref" : "session-id-in-url-rewrite-3",
    "name" : "Session ID in URL Rewrite",
    "type" : "passive",
    "link" : "https://www.zaproxy.org/docs/alerts/3/"
  }]
}
```

winzj:
This PR:
  - closes #1294

### DIFF
--- a/sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/config/data/OwaspZapFullRuleset.java
+++ b/sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/config/data/OwaspZapFullRuleset.java
@@ -1,0 +1,77 @@
+// SPDX-License-Identifier: MIT
+package com.mercedesbenz.sechub.owaspzapwrapper.config.data;
+
+import java.util.LinkedList;
+import java.util.List;
+
+import com.mercedesbenz.sechub.commons.model.JSONable;
+import com.mercedesbenz.sechub.owaspzapwrapper.cli.MustExitCode;
+import com.mercedesbenz.sechub.owaspzapwrapper.cli.MustExitRuntimeException;
+
+public class OwaspZapFullRuleset implements JSONable<OwaspZapFullRuleset> {
+
+    private String timestamp;
+    private String origin;
+    private List<Rule> rules;
+
+    public OwaspZapFullRuleset() {
+        this.rules = new LinkedList<>();
+    }
+
+    public String getTimestamp() {
+        return timestamp;
+    }
+
+    public void setTimestamp(String timestamp) {
+        this.timestamp = timestamp;
+    }
+
+    public String getOrigin() {
+        return origin;
+    }
+
+    public void setOrigin(String origin) {
+        this.origin = origin;
+    }
+
+    /**
+     *
+     * @return list of RuleReference objects or <code>null</code> if not set.
+     */
+    public List<Rule> getRules() {
+        return rules;
+    }
+
+    public void setRules(List<Rule> rules) {
+        this.rules = rules;
+    }
+
+    /**
+     *
+     * @return Rule object or <code>null</code> if not found.
+     */
+    public Rule findRuleByRef(String ref) {
+        if (ref == null) {
+            throw new MustExitRuntimeException("Rule ref to be search for must not be 'null'!", MustExitCode.SCAN_RULE_ERROR);
+        }
+        if (rules == null) {
+            throw new MustExitRuntimeException("Full ruleset must not be 'null', otherwise this cannot be search for ref: " + ref,
+                    MustExitCode.SCAN_RULE_ERROR);
+        }
+
+        for (Rule rule : rules) {
+            if (ref.equals(rule.getRef())) {
+                return rule;
+            }
+        }
+        throw new MustExitRuntimeException(
+                "Rule could not be found inside full ruleset. Check installed ruleset for missing rules and specified rules to deactivate for wrong specifications.",
+                MustExitCode.SCAN_RULE_ERROR);
+    }
+
+    @Override
+    public Class<OwaspZapFullRuleset> getJSONTargetClass() {
+        return OwaspZapFullRuleset.class;
+    }
+
+}
--- a/sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/config/data/Rule.java
+++ b/sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/config/data/Rule.java
@@ -1,0 +1,72 @@
+// SPDX-License-Identifier: MIT
+package com.mercedesbenz.sechub.owaspzapwrapper.config.data;
+
+public class Rule {
+
+    private String id;
+    private String ref;
+    private String name;
+    private String type;
+    private String link;
+
+    /**
+     *
+     * @return id or <code>null</code> if not set.
+     */
+    public String getId() {
+        return id;
+    }
+
+    public void setId(String id) {
+        this.id = id;
+    }
+
+    /**
+     *
+     * @return ref or <code>null</code> if not set.
+     */
+    public String getRef() {
+        return ref;
+    }
+
+    public void setRef(String ref) {
+        this.ref = ref;
+    }
+
+    /**
+     *
+     * @return Name or <code>null</code> if not set.
+     */
+    public String getName() {
+        return name;
+    }
+
+    public void setName(String name) {
+        this.name = name;
+    }
+
+    /**
+     *
+     * @return type or <code>null</code> if not set.
+     */
+    public String getType() {
+        return type;
+    }
+
+    public void setType(String type) {
+        this.type = type;
+    }
+
+    /**
+     *
+     * @return link or <code>null</code> if not set.
+     */
+    public String getLink() {
+        return link;
+    }
+
+    public void setLink(String link) {
+        this.link = link;
+    }
+
+}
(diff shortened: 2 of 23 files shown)


sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/config/data/OwaspZapFullRuleset.java
```diff
@@ -0,0 +1,77 @@
...
+        this.rules = rules;
+    }
+
+    /**
+     *
+     * @return Rule object or <code>null</code> if not found.
+     */
+    public Rule findRuleByRef(String ref) {
+        if (ref == null) {
+            throw new MustExitRuntimeException("Rule ref to be search for must not be 'null'!", MustExitCode.SCAN_RULE_ERROR);
```
Jeeppler, line 55:
```suggestion
            throw new MustExitRuntimeException("Rule reference to be search for must not be 'null'!", MustExitCode.SCAN_RULE_ERROR);
```

sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/config/data/OwaspZapFullRuleset.java
```diff
@@ -0,0 +1,77 @@
...
+
+    public void setRules(List<Rule> rules) {
+        this.rules = rules;
+    }
+
+    /**
+     *
+     * @return Rule object or <code>null</code> if not found.
+     */
+    public Rule findRuleByRef(String ref) {
```
Jeeppler, line 53:
```suggestion
    public Rule findRuleByReference(String reference) {
```

sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/config/data/Rule.java
```diff
@@ -0,0 +1,72 @@
+// SPDX-License-Identifier: MIT
+package com.mercedesbenz.sechub.owaspzapwrapper.config.data;
+
+public class Rule {
+
+    private String id;
+    private String ref;
```
Jeeppler, line 7:
```suggestion
    private String reference;
```

sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/config/data/Rule.java
```diff
@@ -0,0 +1,72 @@
+// SPDX-License-Identifier: MIT
+package com.mercedesbenz.sechub.owaspzapwrapper.config.data;
+
+public class Rule {
```
Jeeppler, line 4: It might be beneficial to make the objects immutable (no setters). Is there any reason, why the rule needs to change after it's creation?

sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/scan/AbstractScan.java
```diff
@@ -289,8 +298,40 @@ protected void setupAdditonalProxyConfiguration() throws ClientApiException {
...
+        if (fullRuleset == null && deactivatedRuleReferences == null) {
+            return;
+        }
+        List<RuleReference> rulesReferences = deactivatedRuleReferences.getDeactivatedRuleReferences();
+        if (rulesReferences == null) {
+            return;
+        }
+
+        for (RuleReference ruleRef : rulesReferences) {
+            Rule ruleToDeactivate = fullRuleset.findRuleByRef(ruleRef.getRef());
```
Jeeppler, line 313: Would a Map or Set not be a better option in this case? Both have the advantage of not needing  to do a search each time.

sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/scan/AbstractScan.java
```diff
@@ -289,8 +298,40 @@ protected void setupAdditonalProxyConfiguration() throws ClientApiException {
...
+
+            if (isPassiveRule(ruleToDeactivate.getType())) {
+                clientApi.pscan.disableScanners(ruleToDeactivate.getId());
+            } else if (isActiveRule(ruleToDeactivate.getType())) {
+                // null specifies the default policy
```
Jeeppler, line 318: policy for what? Reporting, scanning?

sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/scan/AbstractScan.java
```diff
@@ -289,8 +298,40 @@ protected void setupAdditonalProxyConfiguration() throws ClientApiException {
...
+                clientApi.ascan.disableScanners(ruleToDeactivate.getId(), null);
+            }
+        }
+    }
+
+    private boolean isPassiveRule(String type) {
+        return "passive".equals(type);
```
Jeeppler, line 325: do a type to lower case. Some user could provide a rule type of "PaSsive" or so.

sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/scan/AbstractScan.java
```diff
@@ -289,8 +298,40 @@ protected void setupAdditonalProxyConfiguration() throws ClientApiException {
...
+    }
+
+    private boolean isActiveRule(String type) {
+        return "active".equals(type);
```
Jeeppler, line 329: Same as for `passive`. Please convert `type` to lower case before the equals.

sechub-wrapper-owasp-zap/src/test/java/com/mercedesbenz/sechub/owaspzapwrapper/config/OwaspZapScanConfigurationFactoryTest.java
```diff
@@ -313,11 +324,94 @@ void report_file_from_setting_is_used_in_result() {
...
+        Rule rule = iterator.next();
+        assertEquals("12345", rule.getId());
+        assertEquals("rule-ref", rule.getRef());
+        assertEquals("rule-name", rule.getName());
+        assertEquals("active", rule.getType());
+        assertEquals("link-to-rule", rule.getLink());
+    }
+
+    @Test
+    void ruletodeactivate_returned_by_helper_is_inside_result() {
```
Jeeppler, line 363:
```suggestion
    void rule_to_deactivate_returned_by_helper_is_inside_result() {
```

Jeeppler (changes requested):
See comments.

sechub-wrapper-owasp-zap/src/main/resources/owaspzap-ruleset-helper/owaspzap_ruleset_helper.py
```diff
@@ -0,0 +1,70 @@
...
+alert_url = 'https://www.zaproxy.org/docs/alerts/'
+
+
+def generate_owaspzap_ruleset(wanted_status, output_file):
+    response = get(alert_url)
+
+    soup = BeautifulSoup(response.content, 'html.parser')
+    table_body = soup.find('tbody')
+    if table_body is None:
+        raise TypeError('Value of table body variable "tbody" was "None".')
```
Jeeppler, line 19:
```suggestion
        raise TypeError('Unable to find table body. The table body should exists. Value of table body variable "tbody" was "None".')
```

sechub-wrapper-owasp-zap/src/test/java/com/mercedesbenz/sechub/owaspzapwrapper/config/OwaspZapScanConfigurationFactoryTest.java
```diff
@@ -313,11 +326,49 @@ void report_file_from_setting_is_used_in_result() {
     }
 
     @Test
-    void sechub_webconfig_helper_retrieveMaxScanDurationInMillis_is_called_with_correct_webconfig() {
+    void commandlinesettings_null_throws_mustexitruntimeexception() {
```
Jeeppler, line 329:
```suggestion
    void commandline_settings_null_throws_must_exit_runtime_exception() {
```

Jeeppler (approved):
Looks good.

PR #1360, opened 2022-06-02
Fixes PDS shutdown problem #998

### LINKED ISSUE #998: PDSExecutionWatcher leads to hanging process on integration-test shutdown
## Situation
In #996 review we encountered that the PDS server is still running after integration tests has been started.
Inside this issue this problem will be solved.

### Reproduce problem
We simply start a PDS integration test server with
`./gradlew startIntegrationTestPDS`

After the gradle build is done, we  will see with
`ps aux| grep "pds"` that the server is up and running

`./gradlew stopIntegrationTestPDS`

After the gradle build is done, we  will see with
`ps aux| grep "pds"` that the server is STILL up and running - so this is a bug.

de-jcup:
This PR
- closes #998

### DIFF
--- a/sechub-pds/src/main/java/com/mercedesbenz/sechub/pds/IntegrationTestPDSRestController.java
+++ b/sechub-pds/src/main/java/com/mercedesbenz/sechub/pds/IntegrationTestPDSRestController.java
@@ -19,6 +19,7 @@ import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.RestController;
 
+import com.mercedesbenz.sechub.pds.execution.PDSExecutionService;
 import com.mercedesbenz.sechub.pds.job.PDSJob;
 import com.mercedesbenz.sechub.pds.job.PDSJobRepository;
 import com.mercedesbenz.sechub.pds.job.PDSWorkspaceService;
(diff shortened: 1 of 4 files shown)


sven-dmlr (changes requested):
I've tested the behavior:

After `./gradlew startIntegrationTestPDS` and following `./gradlew stopIntegrationTestPDS` the process is gone. :heavy_check_mark: 

But: After `./gradlew integrationtest` has finished, still the PDS process is running.
And it cannot be stopped with `./gradlew stopIntegrationTestPDS`.
Only a `kill <process id>` stops the PDS.
Please check how this can be fixed.

de-jcup:
It is very strange, but this happens also at my machine when starting the complete integration tests via scripts + gradle.
Finally, in the last commit, I did a hard System.exit(0) at the end. This is only for integration tests, so it is okay (I normally would NOT do this in normal production code - but her it is okay). I kept the former introduced preDestroy hook because it works outside integration test scripts very well.

sechub-pds/src/main/java/com/mercedesbenz/sechub/pds/IntegrationTestPDSRestController.java
```diff
@@ -60,8 +64,13 @@ public void isAlive() {
     @RequestMapping(path = PDSAPIConstants.API_ANONYMOUS + "integrationtest/shutdown", method = RequestMethod.GET, produces = {
             MediaType.APPLICATION_JSON_VALUE })
     public void shutdownServer() {
-        LOG.info("Integration test server shutdown initiated by closing context...");
+        LOG.info("Integration test server shutdown wished...");
```
sven-dmlr, line 67: Suggest to use a more common wording:
```suggestion
        LOG.info("Integration test server shutdown requested...");
```

sven-dmlr (approved):
Looks good now and works as expected. :+1: 

Please consider the proposed wording change.
(Anyway I have approved the PR, because there's no major finding)

Issue #1359, opened 2022-05-31
[PMD+PDS] Run multiple instances


Jeeppler:
To unlock even more power, we can run multiple instances of PMD+PDS at the same time. PDS can turn any security scanning tool into a distributed and scalable solution.  To achieve this, a database and data storage, to which all PMD+PDS instances have access to, is required. 

For a blueprint on how to do this have a look at: GoSec+PDS: `sechub-pds-solutions/gosec`. To do this locally, additional Bash and Docker-Compose scripts are needed. The `50-start-multiple-docker-compose.sh` script starts multiple PMD+PDS instances by using the help of a database docker container and load balancer. We use ngnix as reverse proxy for the local setup. A default user needs to be added to the database. In addition, there is a little bit of configuration work needed for the database and load balancer.

The second setup script `51-start-multiple-object-storage-docker-compose.sh` starts a S3 compatible object storage container in addition to the database container and load balancer. We use [SeaweedFS](https://github.com/chrislusf/seaweedfs) as S3 compatible object storage.

Part of epic: #764

zigfridus:
@Jeeppler 
Assign me to this issue please.

Jeeppler:
@zigfridus I assigned you the issue.

zigfridus:
@Jeeppler 
review my PR please.

PR #1361, opened 2022-06-02
Create PMD + PDS mock environment.

### LINKED ISSUE #1358: [PMD + PDS] Mock file
With #1353 PMD is now available as a SecHub PDS Solution. Meaning, that SecHub can scan with PMD. Most people, want to only scan their application to make sure to have little or even better no (security) findings in their source code.

However, if we wanted to test PMD we need some vulnerable code with security vulnerabilities (in this case weaknesses, to be precise). There are a number of ways to get code with vulnerabilities (sorted from less effort to more effort):

1. The tool provides a test data set.
2. Samples with weaknesses. For example, the NIST has a database with example files: https://samate.nist.gov/SARD/test-suites
3. Vulnerable applications. For example, OWASP VWAD has a list of applications: https://owasp.org/www-project-vulnerable-web-applications-directory/
4. Write examples.

Once we found a vulnerable example or wrote one, we can use it to scan and test the tool. Depending on the tool this can take some time. In case, we only want to test the integration between SecHub and PMD+PDS, we can simply return a pre-recorded result file.
To do this we:

1. Scan the example with the PDS+PMD solution (scan using different parameters, if optional parameters can influence the result).
2. Take the result file(s) and put it into the `mocks` folder in the PMD+PDS solution folder (`pmd/docker/mocks`).
3. Add a `PDS_PMD_MOCK` product in the `pds-config.json`. Add optional parameters, if the real product has some. 
4. Write a `pmd-mock.sh` script, which returns the per-recorded mock files. If you have different parameters which can have an influence on the result, return different mock files for each option.

Describing it is more difficult as it is. Have a look at the GoSec+PDS or Multi+PDS for examples.

Furthermore, the mock files can be used to test whether SecHub can read the results or not. SARIF and other standard formats can have large variations between the results, as each author implements the standard slightly different. SecHub contains many unit and integration tests, we can reuse the mock files in those tests.

Part of epic: #764

zigfridus:
Closes #1358

### DIFF
--- a/sechub-pds-solutions/pmd/docker/PMD-Ubuntu.dockerfile
+++ b/sechub-pds-solutions/pmd/docker/PMD-Ubuntu.dockerfile
@@ -47,6 +47,7 @@ COPY pds-config.json "$PDS_FOLDER"/pds-config.json
 
 # Copy PMD scripts and rulesets
 COPY pmd.sh "$SCRIPT_FOLDER"/pmd.sh
+COPY pmd_mock.sh "$SCRIPT_FOLDER"/pmd_mock.sh
 COPY ruleset-security.xml "$SCRIPT_FOLDER"/ruleset-security.xml
 COPY ruleset-all.xml "$SCRIPT_FOLDER"/ruleset-all.xml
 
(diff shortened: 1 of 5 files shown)


Jeeppler (approved):
@zigfridus well done 🥇. I assume it was harder to understand the issue, than implement it.

One small issue is the pull request description. I would appreciate one or two sentences regarding how you solved the issue. You can use bullet point for it. Just linking the issue, means I have to look through the code and the issue. This is particularly important in case the pull request is large, where it might be difficult to immediately see how you solved the issue.

In your case, something like:

"I scanned the CWE506_Embedded_Malicious_Code__aes_encrypted_payload_01.java from https://samate.nist.gov/SARD/downloads/test-suites/2017-10-01-juliet-test-suite-for-java-v1-3.zip to create the mocks."

Next you could implement: https://github.com/mercedes-benz/sechub/issues/1359
Which is a precursor for #1376.

zigfridus:
@Jeeppler 
Thank you for your remark about my description. I took it into consideration and I'll write more detailed description in my future commits.

Issue #1339, opened 2022-05-12
PDS SCAN TARGET TYPE in OWSAP ZAP image


Jeeppler:
Implement and test the PDS Scan Target Type in OWASP ZAP image.

Possible after #1265 was implemented.

Issue #994, opened 2022-02-11
Document `PDS_SCAN_TARGET_TYPE` parameter


Jeeppler:
# Issue

The `PDS_SCAN_TARGET_TYPE` parameter is not mentioned in the documentation for the PDS: https://mercedes-benz.github.io/sechub/latest/sechub-product-delegation-server.html

# Solution

Document the `PDS_SCAN_TARGET_TYPE` parameter in the PDS documentation.

PR #1385, opened 2022-06-08
Feature 1339 pds scan target type

Jeeppler:
- Added the documentation for the `PDS_SCAN_TARGET_TYPE` #994
- Removed config of `PDS_SCAN_TARGET_TYPE`, because it is set implicitly #1339 
- Upgrade ZAP Extensions
- Upgrade PDS_VERSION in Dockerfile, to test it locally

closes: #994
closes: #1339


winzj (approved):
Looks good to me.

PR #1401, opened 2022-06-17
trafficlight `OFF` implemented #810

### LINKED ISSUE #810: Client: Implement traffic light `OFF`
SecHub client must be able to handle the `OFF` value introduced with #809 .

- At this point #806 must be already done.
- In case of `OFF` (or a missing traffic light) we must always break because there is no longer a way to decide if the build is broken or not. The user shall be informed to inspect the messages from serve side printed by the client.

sven-dmlr:
Please see my comment in #810

closes #810

### DIFF
--- a/sechub-cli/src/mercedes-benz.com/sechub/cli/job.go
+++ b/sechub-cli/src/mercedes-benz.com/sechub/cli/job.go
@@ -147,10 +147,7 @@ func printSecHubJobSummaryAndFailOnTrafficLight(context *Context) {
 		}
 	case "GREEN":
 		fmt.Println("  GREEN - no severe security vulnerabilities identified")
-	case "OFF":
-		fmt.Println("  SCAN FAILED - detection of security vulnerabilities failed on server or product")
-		os.Exit(ExitCodeFailed)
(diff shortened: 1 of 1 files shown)


sechub-cli/src/mercedes-benz.com/sechub/cli/job.go
```diff
@@ -147,8 +147,11 @@ func printSecHubJobSummaryAndFailOnTrafficLight(context *Context) {
 		}
 	case "GREEN":
 		fmt.Println("  GREEN - no severe security vulnerabilities identified")
+	case "OFF":
+		fmt.Println("  SCAN FAILED - detection of security vulnerabilities failed on server or product")
+		os.Exit(ExitCodeFailed)
 	case "":
```
de-jcup, line 153: As discussed: I would suggest to use the same mechanism as for "" also for "OFF". So we have the same error message and user shall inspect the status messages from server side

de-jcup (changes requested):

de-jcup (approved):
:rocket: Thank you :-)

PR #1426, opened 2022-06-23
Fix 1419 ajaxspider no implementor

### LINKED ISSUE #1419: PDS OWASP ZAP AjaxSpider No Implementor found
### Problem
Starting the AjaxSpider inside the current PDS OWASP ZAP results in an `ClientApiException`, with the reason `No Implementor`. The reason for this kind of error message means, that the ajax spider extension could not be accessed via the API. Some modules were not installed correctly inside the Image.

### Solution
After some research I noticed, that some dependencies for the OWASP ZAP AjaxSpider Extension are prompted to be installed using the desktop version. The installation used inside the PDS OWASP ZAP missed installing these dependencies. So we need to install these dependencies manually.

### LINKED ISSUE #1421: PDS OWASP ZAP change base image from ubuntu to debian
### Situation
Some packages need to be installed using `snap install` using an ubuntu base image. We don't want to use `snap` inside docker containers.

### Solution
- Change the base image from ubuntu to debian. 
- Update the corresponding parts of the documentation

winzj:
This PR:
-adds 2 extensions to the Owasp Zap required to use the ajax spider
-changes the base image from ubuntu to debian
-installs firefox-esr required for the ajax spider
-updates documentation and other parts from ubuntu to debian

- closes #1419 
- closes #1421

### DIFF
--- a/sechub-pds-solutions/owasp-zap/01-start-single-docker-compose.sh
+++ b/sechub-pds-solutions/owasp-zap/01-start-single-docker-compose.sh
@@ -16,5 +16,5 @@ else
 fi
 
 
-echo "Starting single Ubuntu container."
-docker-compose --file docker-compose_pds_owasp_zap_ubuntu.yaml up --build --remove-orphans
+echo "Starting single Debian container."
+docker-compose --file docker-compose_pds_owasp_zap_debian.yaml up --build --remove-orphans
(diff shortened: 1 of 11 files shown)


Jeeppler (approved):
Good work 🥇

Issue #1428, opened 2022-06-24
PDS SCANCODE improve script development time


Jeeppler:
# Problem

Developing the Scancode image or any other image usually requires to restart the container for every change on the "scan script". As a result, the development cycles are long.

# Solution

Mount the script folder into the container as read/write volume.

Issue #1425, opened 2022-06-23
PDS SCANCODE change base image from ubuntu to debian


winzj:
### Situation
Some packages need to be installed using `snap install` using an ubuntu base image. We don't want to use `snap` inside docker images.

### Solution
- Change the base image from ubuntu to debian. 
- Update the corresponding parts of the documentation

Issue #1418, opened 2022-06-22
[PDS+ScanCode] Add binaries max upload size parameter to configuration


Jeeppler:
Add the `pds.upload.binaries.maximum.bytes` to the PDS configuration.

See: https://github.com/mercedes-benz/sechub/issues/1278

PR #1434, opened 2022-06-27
Implement ability to set the upload size parameter for binaries

Jeeppler:
- Make the binary upload size parameter configurable in the Helm chart #1418
- Define a variable to set the binary upload size parameter locally #1418
- Use Debian rather than Ubuntu for the container #1425
- Improve the script development time by mounting the scripts #1428 

closes: #1418
closes: #1425
closes: #1428


sven-dmlr (changes requested):
Looks good! :+1: 
Found some minor things. Please check the comments.

Additionally:
- in `10-create-image.sh` line 51 the Dockerfile should be `docker/ScanCode-Debian.dockerfile`

sven-dmlr (approved):

PR #1451, opened 2022-07-04
Checkmarx via PDS

### LINKED ISSUE #1415: Implement Checkmarx wrapper for PDS usage
## Situation 

This is a sub issue of #1320

With #1320 it became clear that the data section extraction and filtering process would lead to massive problems (load, cpu usage, storage etc.) when do it at SecHub side. It would be much more better to do this decoupled on a PDS side where archive extraction and filtering is already automated.

## Wanted
We want a possibility to use Checkmarx from a PDS

### General
- create gradle sub project `sechub-wrapper-checkmarx`
- it shall be a spring boot application
- it has the dependency to `sechub-adapter-checkmarx`

### Handling Metadata at PDS side
The checkmarx adapter needs a meta data callback which stores and reads meta data information.
With this information the adapter is able to handle restarts and to gain back a former checkmarx session , e.g. 
when the JVM of the adapter has been crashed before.

The meta data callback was formerly created on sechub side.
And the checkmarx adapter does directly store the metadata into sechub job entity via the callback.

When we switch to the PDS side, we need the call mechanism as well! 
But we have to change it a little bit:

1. SecHub must check if the sechub job has a PDS job uuid already inside meta data
2. If PDS job uuid exists, the PDS adapter must be called with restart and the former PDS job uuid inside
3. When a PDS adapter is called with restart, the adapter must 
   - check if pds job does exist
   - fail when restartable
   - set former PDS job as canceled
   - start again 
   - reuse existing meta data

### LINKED ISSUE #1453: Provide PDS job metadata and make it available in PDS launcher scripts
## Situation
In #1415  we want to support Checkmarx by PDS where we need the meta data inside PDS because we want to
be able to restart a (JVM crashed) PDS job again (see #1465)

## Wanted
Launcher scripts shall be able to read and write meta data - without direct database access. But the meta data
must be inside the PDS job entities.

Also an administrator must be able to fetch the meta data (for debugging).

## Solution
- Introduce  `PDS_JOB_METADATA_FILE`
  - The launcher script will be able to store individual meta data into a file at `$workspaceLocation/metadata.txt`.
  - We introduce new environment variable: `PDS_JOB_METADATA_FILE` which automatically contains the location
- Usecase `PDS Admin fetches job meta data`
  - We provide a REST interface
- Storage
  - The mechanism used for fetching stream data (error and output) will be enhanced to automate also the fetching of meta data inside.
  -  We introduce a new column meta_data in PDS jobs
  - 
- The workspace service will automatically inject meta data from PDS job entity into file system while preparing the workspace

### LINKED ISSUE #1454: Introduce PDS environment variable +  job  parameter for scripts to trust all certificates
## Situation
At test phases or at developer machines, we often have self signed certificates, which are not accepted by default trust store settings.

With #1415 we need a possibility to define if we use "trust all" for communication from launcher script (via wrapper) with Checkmarx server or not.

This is related/familiar to #505 but not the same. In #505 the communication from SecHub directly to product shall have also a generic name. But this is only necessary, when we have still direct communication with products (what is not longer sure, because we will use most times PDS instead.)

## Wanted
We want a standard mechanism and not only for the Checkmarx PDS solution!

## Solution
We introduce new default job parameter  "pds.config.script.trustall.certificates.enabled" which is disabled ("false") per default and always sent to PDS.

It will be available inside PDS launcher scripts by environment variable `PDS_CONFIG_SCRIPT_TRUSTALL_CERTIFICATES_ENABLED`.

### LINKED ISSUE #1457: Restarted SecHub job reuses existing PDS job
## Situation
When a sechub job is restarted (e.g. because of a SecHub JVM crash) always a new PDS job is started.

## Wanted
When a SecHub job is restarted and a PDS job has been run/or already running, the PDS job shall be reused when possible

## Solution
Inside `PDSAdapterV1` we implement following logic

### Sechub
- For each product result for the job, get the meta data
- Inside the meta data: fetch PDS_JOB_UUID
- If PDS_JOB_UUID is available check the result:
   - FAILED, CANCELED, CANCEL_REQUESTED: just start a new PDS job again, change pds adapter meta data to new PDS job UUID 
   - CREATED, RUNNING, QUEUED: use the existing pds job UUID for processing, but do only necessary steps
   - DONE: just fetch the PDS result
- If NO PDS_JOB_UUID is available, create new PDS job etc. etc. like before

### LINKED ISSUE #1475: `PDS` must be able to provide mapping like `SecHub`
## Situation
While implementing #1415 an obstacle was found: 

```
CheckmarxExecutorConfigSuppport.getTeamIdForNewProjects(String projectId)
CheckmarxExecutorConfigSuppport.getPresetIdForNewProjects(String projectId)
```
This is necessary to map new `SecHub` project names to Checkmarx team ids and preset ids.

### Example for the need of mappings
An example::

> ... Every new project which starts with `brachius_" shall have automatically assigned `Checkmarx` team id : XYZ, 
> all other shall have team id: ABC

--> so we just configure the dedicated mapping `checkmarx.newproject.teamid.mapping`and it will work.

### Involved classes
The implementation does use the mapping feature of `SecHub` which results in usage of 
```
NampePatternProvider.java
NamePatternIdProvider.java
MappingData.java
MappingEntry.java
```
from `shared-kernel`. The data itself is inside the `SecHub` database.

### Problem
The new Checkmarx wrapper in #1415 has not (and may not have) access to the `SecHub` database. This means the PDS
and also the wrapper application have no access to this data.

## Wanted
The mapping must work in the Checkmarx wrapper application as well

### General

- [ ] We need the mapping classes inside PDS and also Jackson support so we move them to `sechub-commons-model`.

### Variant1: SecHub transfers the mapping data
We introduce a new common `PDS` executor config parameter : `pds.config.use.sechub.mappings` which will 
contain wanted mappping ids as comma separated list.

This means for the Checkmarx PDS that we use here always
`pds.config.use.sechub.mappings=checkmarx.newproject.teamid.mapping, checkmarx.newproject.presetid.mapping`

Instead of sending the config key to PDS, SecHub will fetch the mappings for the given  ids and use them as generated job parameters. Important: The PDS solution must define those as mandatory/optional parameters in its PDS server config file. Otherwise the PDS will not accept and the script/wrapper will not receive them.

#### Example by Checkmarx
- SecHub will send `checkmarx.newproject.teamid.mapping` and `checkmarx.newproject.presetid.mapping` with JSON values inside to PDS side
- The PDS solution server configuration will have optional/mandatory  parameter definitions for them
- The PDS will fetch the results and accept the job parameters and set them as environment variables for script execution
- The PDS solution script has got he environment variables set and starts the PDS checkmarx solution wrapper which will
  have the environment variables as well and can use them via spring boot value injection.

### Variant 2: PDS holds the mappings itself
We could
- Provide Mapping feature inside PDS
- introduce a new REST API for PDS 
- introduce a special mapping setup inside PDS server config
- automate fetching of configured mapping setup for the PDS product identifier while running and convert the mappings to job parameters which are automatically injected for launcher script as ENV variables

PDS Server Configuration file could be enhanced like this:
```json
{
    "apiVersion" : "1.0",
    
    "serverId" : "unique_id_for_server_or_cluster",
    
    "products" : [
        {
            "id" : "PRODUCT_1",
            "path" : "/srv/security/scanner1.sh",
            "scanType" : "codeScan",
            "description" : "codescanner script needs environment variable ENV_CODESCAN_LEVEL set containing 1,2,3",
            
            "parameters" : {
                "optional" : [
                    {
                        "key" : "product1.add.tipoftheday",
                        "description" : "add tip of the day as info"
                    }  
                ]
            },
             "mappings" : {
                    "mandatory" : [
                    {
                        "key" : "checkmarx.newproject.teamid.mapping",
                        "description" : "A mapping containing name patterns from project name to team id"
                    }  
                ],
                "optional" : [
                    {
                        "key" : "checkmarx.newproject.presetid.mapping",
                        "description" : "An optional preset mapping. So we can use different presets for projects"
                    }  
                ]
             }
            
        }
    ]
}
```

### LINKED ISSUE #1500: A PDS release shall contain also a "sechub-pds-wrapper-checkmarx" artifact
## Situation
There is no PDS checkmarx wrapper jar available online

## Wanted
For #1467 we need the possiblity to download the checkmarx wrapper from #1415 by docker container from github releases.

## Solution
Update release workflow

de-jcup:
This PR
- [x] closes #1453
- [x] closes #1415
- [x] closes #1454
- [x] closes #1457
- [x] closes #1475 
- [x] closes #1500


winzj (changes requested):
Good Job. Thank you @de-jcup for the pair review help (we did the review together).
Please have a look at the comments.

winzj (approved):
Great work. Looks good to me.

PR #1455, opened 2022-07-05
Wrapper OWASP ZAP can read rules to deactivate from env variable #1452

### LINKED ISSUE #1452: Wrapper OWASP ZAP should be able to read rule references to deactivate from ENV variable (comma separated)
### Situation
Currently the wrapper OWASP ZAP can read json files to deactivate certain scan rules. With #1294 introduced a format to handle rules by combining the rule name and the rule id into a rule reference. We want the ability to set this rule references in an easy way via the `executor config`, to make testing and and adjusting the scan easier to handle. To test different kind of rule sets fast while the PDS OWASP ZAP is running, this is cumbersome.

### Solution
We introduce the env variable `ZAP_DEACTIVATED_RULE_REFERENCES`, which contains a comma separated list of rule references like:

 `ZAP_DEACTIVATED_RULE_REFERENCES=Cross-Site-Scripting-(DOM-Based)-40026,Timestamp-Disclosure-10096`

The format for the rule references can be found at this test file for example: [owaspzap-full-ruleset.json](https://github.com/mercedes-benz/sechub/blob/develop/sechub-wrapper-owasp-zap/src/test/resources/zap-available-rules/owaspzap-full-ruleset.json#L5). You can find the rule references at the beginning of each rule, for example line 5,11,... of the file above. If this variable is specified, the file on the command line will not be respected, to make testing more flexible.

winzj:
This PR implements:
- closes #1452

### DIFF
--- a/sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/scan/AbstractScan.java
+++ b/sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/scan/AbstractScan.java
@@ -180,7 +180,7 @@ public abstract class AbstractScan implements OwaspZapScan {
             waitForNextCheck();
             clientApi.pscan.recordsToScan();
             numberOfRecords = Integer.parseInt(((ApiResponseElement) clientApi.pscan.recordsToScan()).getValue());
-            LOG.info("For scan {}: Passive scan number of records left for scanning: " + numberOfRecords, scanConfig.getContextName());
+            LOG.info("For scan {}: Passive scan number of records left for scanning: {}", scanConfig.getContextName(), numberOfRecords);
         }
         LOG.info("For scan {}: Passive scan completed.", scanConfig.getContextName());
         remainingScanTime = remainingScanTime - (System.currentTimeMillis() - startTime);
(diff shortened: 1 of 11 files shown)


Jeeppler (changes requested):
See comments.

sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/scan/AbstractScan.java
```diff
@@ -282,6 +282,9 @@ protected void setupBasicConfiguration() throws ClientApiException {
         // enable all passive scanner rules by default
         // null specifies the default scan policy
         clientApi.ascan.enableAllScanners(null);
+
+        // use firefox in headless mode by default
+        clientApi.ajaxSpider.setOptionBrowserId("firefox-headless");
```
Jeeppler, line 287: Is there a command line argument for deciding what browser should be used? It would be nice to be able to switch between browsers. In any case, make a new issue for this feature, if you agree, that it should be possible to change the browser used for scanning.

Jeeppler (approved):
Looks good 💯 . Merge it.

In addition, please have a look at the single comment I left.

PR #1473, opened 2022-07-11
Implement a solution to limit the memory resources of pds-owaspzap #1472

### LINKED ISSUE #1472: Limit the ressources for PDS-OWASPZAP
### Situation
The PDS-OWASPZAP sometimes uses a lot of memory because there is not limitation set. Especially the AjaxSpider uses a lot of resources because of the used web browser in headless mode.

### Solution
We want to find a limitation so the PDS-OWASPZAP cannot use to much of the memory available. The resource limitations shall be configurable, like all other values used in the Helm Chart.

winzj:
### Remark
I mistakenly changed the version of the Chart.yaml to `0.7.0` on the previous Pull Request, but did not change the Helm Chart at all. Because of this I did not update the Helm Chart version by now in this Pull Request. **Should this be done anyway?**

- closes #1472


Jeeppler (approved):
Looks good. In addition, please update the Helm Chart version to 0.8.0.

Issue #1479, opened 2022-07-14
Make new Logo Visible


Jeeppler:
Include the new logo #1400 into:

- Static Website
- README.md
- Documentation

PR #1482, opened 2022-07-14
Feature 1479 make new logo visible

Jeeppler:
- new logo in Readme.md
- changed imprint to MBTI
- changed logos in HTML page
- resized SVG file
- renamed new logos to `sechub-logo.png` to ensure it will be included into the documentation.

closes: #1479


docs/main.css
```diff
@@ -1,4 +1,5 @@
 /* SPDX-License-Identifier: MIT */
+/* SPDX-FileCopyrightText: 2022 Mercedes-Benz Tech Innovation GmbH */
```
de-jcup, line 2: I do not think we need this here. We have the MIT license - that is enough. We have not embedded inside the CSS any images/svgs etc.

de-jcup: Please remove copyright information - we use MIT license here and so it is not necessary.

de-jcup (changes requested):
Good job. But one thing I want to have changed - see comments

de-jcup (approved):

PR #1494, opened 2022-07-20
Fix PDS OWASP ZAP by extending pds-config #1493

### LINKED ISSUE #1493: Forgot `zap.deactivated.rule.references` inside pds-config.json
### Problem
I forgot to add `zap.deactivated.rule.references` inside the `pds-config.json` to the optional parameters.

### Solution
This needs to be added, otherwise the parameter will be ignored because it is not known.

winzj:
- closes #1493


Jeeppler (approved):
Looks good.

Issue #1509, opened 2022-07-26
`OWASP ZAP+PDS` Use shared resources


Jeeppler:
OWASP ZAP+PDS should use the same object-storage, loadbalancer and database files as the rest of the PDS-Solutions.

PR #1512, opened 2022-07-26
OWASP ZAP PDS use shared resources #1509

Jeeppler:
- Docker-compose files now use shared database, loadbalancer and object storage #1509
- Scan scripts can be modified from outside while the container is running. As a result changes are directly applied
- Shorter README by removing manual steps

closes: #1509


winzj (approved):
Looks good to me.

Issue #1466, opened 2022-07-06
Dockerfile for SecHub server


Jeeppler:
Provide a Dockerfile for SecHub server.

At the beginning the server will run in development mode and uses the H2 database.

Part of: #320

PR #1515, opened 2022-07-27
Feature 1466 dockerfile for sechub server

Jeeppler:
A dockerfile for the SecHub server #1466

- the dockerfile has three build types

  1. Download SecHub Release
  2. Build SecHub Release
  3. Copy SecHub Jar

- added scripts to setup products automatically

Note for the reviewer:  Please pay attention to the Dockerfile and the scripts to setup the products automatically. The main idea of the scripts is to create small building blocks which can be used standalone, but put together as well. 

closes: #1466


sven-dmlr (changes requested):
It's good to have scripts for setting up projects, profiles and all the stuff by script.
Please check my review comments.

Did you agree the folder name `sechub-solution` with @de-jcup ?
I would not expect these scripts under that name; although, I don't have a proposal for a "better" one. :grin:

sven-dmlr (approved):

Issue #1521, opened 2022-08-04
quickstart guide needs update regarding golang


sven-dmlr:
## Problem:
Following the Quickstart guide, the compilation of the SecHub client will fail because the outdated go version of Ubuntu/Debian/Alpine distro.
https://mercedes-benz.github.io/sechub/latest/sechub-quickstart-guide.html

## Todo:
Describe how to download the latest go version and do not use the distribution's golang package.

Issue #1520, opened 2022-08-04
Update quickstart guide - move path to front


company-jcup:
## Situation
Inside  https://mercedes-benz.github.io/sechub/latest/sechub-quickstart-guide.html
we have got
``` 
export PATH="`pwd`/sechub-cli/build/go/platform/linux-amd64:`pwd`/sechub-developertools/scripts:$PAT
``` 

## Wanted
the path variable should be moved to the beginning for security aspects

## Solution
move...

PR #1524, opened 2022-08-04
Changed the documentation

expp121:
* Moved the styles for all pages into a separate file
    * Merged Debian and Ubuntu section in quickStartGuide
    * Added Warning for Go lang version (Issue https://github.com/mercedes-benz/sechub/issues/1521)
    * Swapped the place of the PATH variable to the beginning, becuase it poses a security risk (Issue https://github.com/mercedes-benz/sechub/issues/1520)


de-jcup (changes requested):
@expp121 : thank your for effort- 

But ... unfortunately this will not work, because all content inside docs/latest is generated by our build scripts (via Asciidoctor).  

Because this misunderstanding could be more often, I wrote an entry inside the wiki :
https://github.com/mercedes-benz/sechub/wiki/Documentation#how-to-contribute-to-documentation.

Please refer to this wiki entry and make your changes in the necessary `.adoc` files only.
Maybe it would be a good choice to create a complete new PR where you have only the adoc changes inside.

PR #1525, opened 2022-08-04
PDS Product time out writes now execution data into DB, fixes #1523

### LINKED ISSUE #1523: Product timeout at PDS side does not write error and output stream results into database
## Situation
When a PDS script needs longer than the configured (or default) product timeout time (can be configured inside PDS server config file), the error and output streams and also the messages are not fetched!

## Wanted
Even on a product timeout we want to have the execution result must contain output and error streams + messages

## Solution
Change implementation

de-jcup:
This PR
- closes #1523


winzj (approved):
Looks good to me.

PR #1540, opened 2022-08-10
Feature 1505 fix low on ephemeral storage

### LINKED ISSUE #1505: PDS-owaspzap sometimes fails because node runs low on ephemeral-storage
### Situation
After some scans or long running scans sometimes the scan fails because the node runs low on ephemeral storage. This issue might be caused by big log files, log files that are not removed or maybe both. This needs to be investigated.

### Wanted
Enough ephemeral-storage, so the container can keep running.

winzj:
- closes #1505


Jeeppler (approved):
Looks good.

PR #1543, opened 2022-08-10
Enhanced server validation for false positives #1541

### LINKED ISSUE #1541: When a false-positive list without entries is uploaded, SecHub shall log into audit and not into security log
## Situation
When user adds 0 false positives, the `FalsePositiveJobDataListValidationImpl` does mark this as invalid, because minimum is one entry.

As a resulta a "NotAcceptable" exception is thrown which leads to security log entries and we have no audit log entry for the user and the project.

## Wanted
When project id is correct and a user has access to the project (and of course the authentication is correct as well...) we should always log inside audit log and just ignore it when it is 0 entires.

## Solution
- Change `FalsePositiveJobDataListValidationImpl` - accept also without entries
- Fast guard inside `FalsePositiveJobDataService`: when 0 entries directly just return after the audit log...

### LINKED ISSUE #1542: False positive job data validation results not added
## Situation
While implementing #1541 a bug was detected:
The validation `FalsePositiveJobDataListValidationImpl` does check currently only the api version and the size of the list, but does
not add the results from called job validation.

## Wanted
Validation error results shall be added

## Solution
Change implementation

de-jcup:
This PR
- closes #1541
- closes #1542

### DIFF
--- a/sechub-doc/src/docs/asciidoc/documents/code2doc/client/02_sechub_client.adoc
+++ b/sechub-doc/src/docs/asciidoc/documents/code2doc/client/02_sechub_client.adoc
@@ -168,6 +168,8 @@ include::../../shared/false-positives/false-positives-REST-API-content-example1.
 <4> ID of finding, which shall be marked as false positive
 <5> Comment (optional) describing the reason why this is handled as false positive
 
+NOTE: Providing an empty list here does *NOT delete* the declared false positives! +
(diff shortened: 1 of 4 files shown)


sven-dmlr (approved):
Tested successfully: The 406 message is gone. :+1: 

@de-jcup : I added a note about empty fp list to the client documentation. Please check and adapt as you like.

Issue #1534, opened 2022-08-09
Add Supporting Docker-Compose Files to Containerized SecHub


Jeeppler:
SecHub can work with PostgreSQL and a file share or object storage. Add supporting compose/docker files to SecHub.

Part of: #320

PR #1568, opened 2022-08-17
Feature 1534 supporting docker compose files

Jeeppler:
Add supporting infrastructure to docker-compose file

- adds loadbalancer
- adds database
- adds object storage
- adds the ability to set memory limits for local testing

these changes allow to run SecHub locally in a cluster

closes: #1534


sechub-solution/docker/SecHub-Debian.dockerfile
```diff
@@ -63,6 +68,7 @@ RUN mkdir --parent "$BUILD_FOLDER" && \
     cd "sechub" && \
     "./buildExecutables" && \
     if [[ -n "$TAG" ]]; then git checkout tags/"$TAG" -b "$TAG"; fi && \
```
sven-dmlr, line 70: git checkout after building makes no sense.
So maybe the tag/branch handling should happen before the build.

sechub-solution/docker/SecHub-Debian.dockerfile
```diff
@@ -63,6 +68,7 @@ RUN mkdir --parent "$BUILD_FOLDER" && \
...
+    if [[ -n "$BRANCH" ]]; then git checkout "$BRANCH"; fi && \
     cp "sechub-server/build/libs/sechub-server-0.0.0.jar" --target-directory "$SECHUB_ARTIFACT_FOLDER"
```
sven-dmlr, line 72: When building a tagged version, the jar will not be named `0.0.0`.
Suggestion:
```suggestion
    cp "sechub-server/build/libs/sechub-server-*.jar" --target-directory "$SECHUB_ARTIFACT_FOLDER"
```

sven-dmlr (changes requested):
Good job!
Please check my comments.

sven-dmlr (approved):

PR #1571, opened 2022-08-18
x-file-size header set #1552

### LINKED ISSUE #1552: Client: Set the "x-file-size" for binary uploads
In #1471 we understood, that the S3 client requires the file length upfront. With #1551 we introduce the `x-file-size` header field for binary uploads. 

This issue is about setting the `x-file-size` header in the client. The client needs to get the file size for binary files and set the `x-file-size`.

Depends on: #1551

### LINKED ISSUE #1572: Client: Add output for tar file creation
## Situation
Currently there is no output when adding single files to a binary tar file for upload - so the client seems to be stuck e.g. when uploading a multiple GB file.

For folders, the message is like this:
```
2022-08-19 11:58:41 (+02:00) Zipping folder: Code (/home/user/src/code)
```

## Wanted
Add output for binaries upload like this:
```
2022-08-19 12:41:13 (+02:00) Creating tar archive for upload (/home/user/test/binaries-myproject.tar)
2022-08-19 12:41:13 (+02:00) Adding file '/data/mybigisofile.iso' to tar
```

sven-dmlr:
- also HTTP debugging consolidated

closes #1552 
closes #1572

### DIFF
--- a/sechub-cli/src/mercedes-benz.com/sechub/cli/controller-uploadsourcezip.go
+++ b/sechub-cli/src/mercedes-benz.com/sechub/cli/controller-uploadsourcezip.go
@@ -3,7 +3,6 @@
 package cli
 
 import (
-	"fmt"
 	"os"
 
 	sechubUtil "mercedes-benz.com/sechub/util"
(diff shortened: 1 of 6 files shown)


winzj (approved):
Good job, looks good to me.

Issue #1608, opened 2022-08-31
Containers: helper script to visualize resource usage


Jeeppler:
# Problem

Some PDS-Solutions like FindSecurityBugs+PDS and others require a good amount of memory for the analysis. However, monitoring the memory and CPU usage requires visually look at the resource usage of the `docker stats` command.

# Solution

Write script which records the `docker stats` information periodically. Write a second script which visualizes the resource usage, especially: memory and CPU.

PR #1609, opened 2022-08-31
Helper scripts for resource recording and visualization #1608

Jeeppler:
Helper scripts for resource recording and visualization #1608

- two scripts to record and visualize the resource usage of Docker container

closes: #1608


winzj (approved):
Good idea👍 
Looks good, can be merged.

PR #1611, opened 2022-08-31
Create small CLI tool for testing the OpenAPI java client #1506.

### LINKED ISSUE #1506: Create Demo/Test App for OpenAPI Java Client
Create a demo/test application which uses the OpenAPI Java client.

zigfridus:
Create small CLI tool for testing the OpenAPI java client #1506.
Closes #1506.

### DIFF
--- a/sechub-api-java/src/main/java/com/mercedesbenz/sechub/api/java/demo/cli/SecHubClientConfigurationFactory.java
+++ b/sechub-api-java/src/main/java/com/mercedesbenz/sechub/api/java/demo/cli/SecHubClientConfigurationFactory.java
@@ -16,19 +16,16 @@ public class SecHubClientConfigurationFactory {
 
     public SecHubClient create(CommandLineSettings settings) {
         URI serverUri = null;
-        if (settings.getServerUri() != null)
-            serverUri = URI.create(settings.getServerUri());
+        if (settings.getServer() != null)
+            serverUri = URI.create(settings.getServer());
 
-        int serverPort = settings.getServerPort();
         String userId = settings.getUserId();
         String apiToken = settings.getApiToken();
+        String trustAll = settings.getTrustAll();
 
-        String envServerUri = environmentVariableReader.readAsString(EnvironmentVariableConstants.SECHUB_SERVER_URI);
+        String envServerUri = environmentVariableReader.readAsString(EnvironmentVariableConstants.SECHUB_SERVER);
         if (serverUri == null && envServerUri != null) {
             serverUri = URI.create(envServerUri);
-        }
-        if (serverPort <= 0 || serverPort > 65535) {
-            serverPort = environmentVariableReader.readAsInt(EnvironmentVariableConstants.SECHUB_SERVER_PORT);
         }
         if (userId == null) {
             userId = environmentVariableReader.readAsString(EnvironmentVariableConstants.SECHUB_USERID);
@@ -36,15 +33,13 @@ public class SecHubClientConfigurationFactory {
         if (apiToken == null) {
             apiToken = environmentVariableReader.readAsString(EnvironmentVariableConstants.SECHUB_APITOKEN);
         }
+        if (trustAll == null) {
+            trustAll = environmentVariableReader.readAsString(EnvironmentVariableConstants.SECHUB_TRUSTALL);
+        }
 
         if (!isServerUriValid(serverUri)) {
             throw new SecHubClientConfigurationRuntimeException(
-                    "Sechub server's URI is invalid. Please set the Sechub server's URI as an environment variable (SECHUB_SERVER_URI) or as a program argument (--serverUri).");
-        }
-
-        if (serverPort <= 0 || serverPort > 65535) {
-            throw new SecHubClientConfigurationRuntimeException("Sechub server's port is set to " + serverPort
-                    + ". Please set the Sechub server's port as an environment variable (SECHUB_SERVER_PORT) or as a program argument (--serverPort).");
+                    "Sechub server's URI is invalid. Please set the Sechub server's URI as an environment variable (SECHUB_SERVER) or as a program argument (--server).");
         }
 
         if (userId == null) {
@@ -57,18 +52,29 @@ public class SecHubClientConfigurationFactory {
                     "Privileged user's API token is null. Please set the privileged user's API token as an environment variable (SECHUB_APITOKEN) or as a program argument (--apiToken).");
         }
 
-        return SecHubClient.create(userId, apiToken, serverUri.toString(), serverPort);
+        if (!isTrustAllValid(trustAll)) {
+            throw new SecHubClientConfigurationRuntimeException(
+                    "trustAll value is invalid. Please set the trustAll as an environment variable (SECHUB_TRUSTALL) or as a program argument (--trustAll). Correct values are true and false.");
+        }
+        return trustAll == null ? SecHubClient.create(userId, apiToken, serverUri)
+                : SecHubClient.create(userId, apiToken, serverUri, Boolean.parseBoolean(trustAll));
     }
 
     private boolean isServerUriValid(URI serverUri) {
         if (serverUri == null)
             return false;
-        if (!serverUri.getScheme().equals("http") && !serverUri.getScheme().equals("https"))
+        if (!serverUri.getScheme().equals("https"))
             return false;
         if (!serverUri.getPath().isEmpty())
             return false;
-        if (serverUri.getPort() != -1)
+        if (serverUri.getPort() <= 0 || serverUri.getPort() > 65535)
             return false;
         return true;
+    }
+
+    private boolean isTrustAllValid(String trustAll) {
+        if (trustAll == null || trustAll.equals("false") || trustAll.equals("true"))
+            return true;
+        return false;
     }
 }
--- a/sechub-api-java/src/main/java/com/mercedesbenz/sechub/api/java/demo/cli/SecHubClientConfigurationFactory.java
+++ b/sechub-api-java/src/main/java/com/mercedesbenz/sechub/api/java/demo/cli/SecHubClientConfigurationFactory.java
@@ -1,0 +1,56 @@
+// SPDX-License-Identifier: MIT
+package com.mercedesbenz.sechub.api.java.demo.cli;
+
+import com.mercedesbenz.sechub.api.java.SecHubClient;
+import com.mercedesbenz.sechub.api.java.demo.cli.exceptions.SecHubClientConfigurationRuntimeException;
+
+public class SecHubClientConfigurationFactory {
+
+    EnvironmentVariableReader environmentVariableReader;
+
+    public SecHubClientConfigurationFactory() {
+        environmentVariableReader = new EnvironmentVariableReader();
+    }
+
+    public SecHubClient create(CommandLineSettings settings) {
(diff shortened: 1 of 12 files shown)


sechub-api-java/src/main/java/com/mercedesbenz/sechub/api/java/demo/cli/SecHubClientConfigurationFactory.java
```diff
@@ -0,0 +1,56 @@
...
+    public SecHubClient create(CommandLineSettings settings) {
+        String serverUri = settings.getServerUri();
+        int serverPort = settings.getServerPort();
+        String userId = settings.getUserId();
+        String apiToken = settings.getApiToken();
+
+        if (serverUri == null) {
+            serverUri = environmentVariableReader.readAsString(EnvironmentVariableConstants.SECHUB_SERVER_URI);
+        }
+        if (serverPort <= 0) {
```
Jeeppler, line 24: The size of `int` is dependent on the hardware. Today it varies between 32 and 64 bit. A port needs to be int between 0 and 65535. Java does not have unsigned short integer types, therefore using int is fine, but an additional check for ports over 65535 is needed.

sechub-api-java/src/main/java/com/mercedesbenz/sechub/api/java/demo/cli/SecHubClientConfigurationFactory.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    public SecHubClient create(CommandLineSettings settings) {
+        String serverUri = settings.getServerUri();
+        int serverPort = settings.getServerPort();
+        String userId = settings.getUserId();
+        String apiToken = settings.getApiToken();
+
+        if (serverUri == null) {
+            serverUri = environmentVariableReader.readAsString(EnvironmentVariableConstants.SECHUB_SERVER_URI);
```
Jeeppler, line 22: Java does have an URI type which you could parse in addition to ensure it is well formatted. Furthermore, it might make sense to ignore some part of the URI (for example, query parameters) and check the protocol (needs to be HTTP(S)).

For example by using `URI.create()`: https://docs.oracle.com/javase/8/docs/api/java/net/URI.html#create-java.lang.String-

sechub-api-java/src/main/java/com/mercedesbenz/sechub/api/java/demo/cli/SechubServerTest.java
```diff
@@ -0,0 +1,82 @@
...
+import com.mercedesbenz.sechub.api.java.auth.Authentication;
+import com.mercedesbenz.sechub.api.java.auth.HttpBasicAuth;
+import com.mercedesbenz.sechub.api.java.demo.cli.exceptions.OpenAPITestToolRuntimeException;
+
+public class SechubServerTest {
+
+    private static final Logger LOG = LoggerFactory.getLogger(SechubServerTest.class);
+    SecHubClient secHubClient = null;
+    ApiClient apiClient = null;
+    StringBuilder fullConnectionURI = new StringBuilder();
```
Jeeppler, line 22: A StringBuilder is a good way to create a string in Java. Strings in Java are immutable. Meaning, while concatenating two strings one creates a third string. This can be expensive in cases in which the String contains a lot of text.

However, StringBuilder should not be use as a class attribute. You use a StringBuilder to "create/make" a new string.

```suggestion
    String fullConnectionURI = "";
```

sechub-api-java/src/main/java/com/mercedesbenz/sechub/api/java/demo/cli/SechubServerTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+    public SechubServerTest(SecHubClient secHubClient) {
+        this.secHubClient = secHubClient;
+        fullConnectionURI.append(secHubClient.getHostUri()).append(":").append(secHubClient.getHostPort());
```
Jeeppler, line 26:
```suggestion
        StringBuilder sb = new StringBuilder();
        sb.append(secHubClient.getHostUri()).append(":").append(secHubClient.getHostPort());
        fullConnectionURI = sb.toString();
```

sechub-api-java/src/main/java/com/mercedesbenz/sechub/api/java/demo/cli/SechubServerTest.java
```diff
@@ -0,0 +1,82 @@
...
+            throw new OpenAPITestToolRuntimeException("Sechub server is not alive.");
+
+        if (!isListOfUsersFetched())
+            throw new OpenAPITestToolRuntimeException("Can not fetch list of users from the Sechub server.");
+    }
+
+    private boolean isServerAlive() {
+        LOG.info("*** Check on whether Sechub server is alive.");
+        apiClient = Configuration.getDefaultApiClient();
+        apiClient.setBasePath(fullConnectionURI.toString());
```
Jeeppler, line 40:
```suggestion
        apiClient.setBasePath(fullConnectionURI);
```

sechub-api-java/src/main/java/com/mercedesbenz/sechub/api/java/demo/cli/SechubServerTest.java
```diff
@@ -0,0 +1,82 @@
...
+        LOG.info("*** Trying to fetch list of users.");
+
+        HttpBasicAuth basicAuthentication = new HttpBasicAuth();
+        basicAuthentication.setUsername(secHubClient.getUsername());
+        basicAuthentication.setPassword(secHubClient.getSealedApiToken());
+        Map<String, Authentication> authMap = new HashMap<>();
+        authMap.put("basic", basicAuthentication);
+
+        apiClient = new ApiClient(authMap);
+        apiClient.setBasePath(fullConnectionURI.toString());
```
Jeeppler, line 64:
```suggestion
        apiClient.setBasePath(fullConnectionURI);
```

Jeeppler (changes requested):
Overall well done. Code looks good. However, I found a few issues.

Jeeppler:
@winzj please have a look at this pull request as well.

zigfridus:
@Jeeppler 
Thank you for your review. I will update my code according to your comments.

Jeeppler:
@zigfridus perfect. To ask for a re-review press the little circle.

image

We always request changes or approve a pull-request. It can take some time to approve a pull-request, depending on the workload of potential reviewers.

sechub-api-java/src/main/java/com/mercedesbenz/sechub/api/java/demo/cli/SecHubClientConfigurationFactory.java
```diff
@@ -0,0 +1,74 @@
...
+    public SecHubClient create(CommandLineSettings settings) {
+        URI serverUri = null;
+        if (settings.getServerUri() != null)
+            serverUri = URI.create(settings.getServerUri());
+
+        int serverPort = settings.getServerPort();
+        String userId = settings.getUserId();
+        String apiToken = settings.getApiToken();
+
+        String envServerUri = environmentVariableReader.readAsString(EnvironmentVariableConstants.SECHUB_SERVER_URI);
```
Jeeppler, line 26: Instead of having a separate `SECHUB_SERVER_URI` and `SECHUB_SERVER_PORT` variables use a combined `SECHUB_SERVER` variable. It is easier for the user. The same applies to the command line.

The `sechub-api.sh` script expects a single `SECHUB_SERVER` argument. It would be nice if your demo app would work the same way.

sechub-api-java/src/main/java/com/mercedesbenz/sechub/api/java/demo/cli/SecHubClientConfigurationFactory.java
```diff
@@ -0,0 +1,74 @@
...
+            throw new SecHubClientConfigurationRuntimeException(
+                    "Sechub server's privileged user id is null. Please set the Sechub server's privileged user id as an environment variable (SECHUB_USERID) or as a program argument (--userId).");
+        }
+
+        if (apiToken == null) {
+            throw new SecHubClientConfigurationRuntimeException(
+                    "Privileged user's API token is null. Please set the privileged user's API token as an environment variable (SECHUB_APITOKEN) or as a program argument (--apiToken).");
+        }
+
+        return SecHubClient.create(userId, apiToken, serverUri.toString(), serverPort);
```
Jeeppler, line 60: There should be another variable called `SECHUB_TRUSTALL`  if this is set to `true` the second `SecHubClient create(String username, String apiToken, String hostUri, int hostPort, boolean trustAll)` constructor should be called.

The `SECHUB_TRUSTALL` is necessary to accept self-signed certificates. Otherwise, one will get a error message like:

~~~
javax.ws.rs.ProcessingException: javax.net.ssl.SSLHandshakeException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
…
Caused by: javax.net.ssl.SSLHandshakeException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
…
Caused by: sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
…
Caused by: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
…
[main] ERROR com.mercedesbenz.sechub.api.java.demo.cli.OpenAPITestTool - Sechub server testing failed
~~~

By default Java only trusts certificates which it knows.

You do not have to implement the trustAllCertificate functionality. I will create a separate issue for that.

Jeeppler, line 60: Please change the `SecHubClient.java` create methods. They should work using an URI rather than two separate values `hostUri` and `hostPort`. This is a design flaw, it does not make a lot of sense to work with a string and integer. Java is object oriented and has the URI object as part of the standard API. 

You should be able to create the `SecHubClient` with this call:

~~~
return SecHubClient.create(userId, apiToken, serverUri);
~~~

or with

~~~
return SecHubClient.create(userId, apiToken, serverUri, trustAllCertificates);
~~~

That should feel more "natural", passing a URL.

sechub-api-java/src/main/java/com/mercedesbenz/sechub/api/java/demo/cli/SecHubClientConfigurationFactory.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    private boolean isServerUriValid(URI serverUri) {
+        if (serverUri == null)
+            return false;
+        if (!serverUri.getScheme().equals("http") && !serverUri.getScheme().equals("https"))
```
Jeeppler, line 66: As far as I know, SecHub always requires a `https` connection. Either using a self-signed certificate or using a real certificate from an authority.

Jeeppler (changes requested):
See comments. 

NOTE: To make sure the Java SecHub client trusts all certificates is not easy.

Jeeppler (approved):
Good job 👍. 

One step closer to a usable Java API client 💯

PR #1619, opened 2022-09-08
Fix openapi scanning #1615

### LINKED ISSUE #1615: OWASP ZAP Wrapper OpenApi Scanning exits because missing condition
### Situation
During the implementation of the open api scanning there was a misunterstanding regarding the data section part on my side. I thought the ENV variable:
`PDS_JOB_EXTRACTED_SOURCES_FOLDER` on PDS side will only be set if there are sources extracted to the PDS. I relied on this condition, but this is not enough.

### Solution
Check if the data section was found, otherwise assume no API definition was uploaded.

winzj:
- closes #1615


Jeeppler (approved):
Looks good.

Issue #1618, opened 2022-09-07
Fix build on development branch


Jeeppler:
# Problem

With the completion of #1506 the generated SecHub API Client for Java is actually used. However, it also broke the build, as the small CLI application depends on the generated SecHub API Client source code. The problem is, the dependencies are not available during the Gradle configuration phase, because they are not yet generated.

It the other problem is, the GitHub Workflow checks do not run in case of external contributions. Have a look at the other issue for more information: #1617.

# Solution

Fix the failing build. The step of building the OpenAPI Test/Demo App needs to be postponed until the required parts are in-place/build.

The steps are described in the `README.adoc` file: https://github.com/mercedes-benz/sechub/blob/develop/sechub-api-java/src/main/java/com/mercedesbenz/sechub/api/java/demo/cli/README.adoc#build

PR #1623, opened 2022-09-08
Fix build by moving the demo app to examples #1618

Jeeppler:
- move demo app to examples
- upgrade tests to Junit 5
- prepare for future work

closes: #1618


sechub-api-java/build.gradle
```diff
@@ -6,46 +6,34 @@
...
+    
+    testImplementation project(':sechub-testframework')
     testImplementation spring_boot_dependency.junit_jupiter
+}
 
-    testImplementation spring_boot_dependency.mockito_core
-    testImplementation library.mockito_inline
-    testImplementation spring_boot_dependency.hamcrest
+sourceSets {
+    main {
```
de-jcup, line 24: This seems to be non necessary code.

If this is a preparation for the next steps we should have a comment here.
Or we just remove all parts here.

sechub-api-java/build.gradle
```diff
@@ -90,8 +78,8 @@ task callOpenAPIJavaGenerator(type: org.openapitools.generator.gradle.plugin.tas
             licenseUrl					: "https://github.com/mercedes-benz/sechub/blob/develop/LICENSE",
             developerEmail				: "",
             developerName				: "",
-            developerOrganization 		: "Daimler TSS",
-            developerOrganizationUrl	: "https://www.daimler-tss.com/",
+            developerOrganization 		: "Mercedes-Benz Tech Innovation",
```
de-jcup, line 81: Could you please fix the indention and replace tabs by spaces ? 
You could also setup eclipse to use always spaces instead of tabs, so this will  not happen again.

de-jcup (approved):
There were two minor findings by myself. Feel free to fix them or to leaf it as is. See my comments for the details - I approved it anyway.

Inside the first PR comment I changed the `closes #1619`  issue to `closes #1618` - so it is correctly connected with the issue. I think, this was a typo. If not please change back.

Issue #1631, opened 2022-09-09
PDS ScanCode Make Timeout Configurable


Jeeppler:
# Problem

In case, one scans a large file and it takes longer than the 120 seconds the scan will timeout.
For more information have a look at: https://github.com/nexB/scancode-toolkit/issues/641

This can happen, if the scanned file contains a large number of licenses.

# Solution

There is a command line option to increase the timeout of 120 seconds:

~~~
--timeout FLOAT

    Stop scanning a file if scanning takes longer than a timeout in seconds. [Default: 120]
~~~

source: https://scancode-toolkit.readthedocs.io/en/stable/cli-reference/list-options.html#all-core-scan-options

Make this timeout configurable.

PR #1634, opened 2022-09-09
Make timeout parameter configurable #1631

Jeeppler:
- add timeout parameter to pds-config #1631
- add timeout parameter to scancode.sh #1631
- refactor scancode config files

closes: #1631


sven-dmlr (approved):
Looks good

PR #1654, opened 2022-09-16
Added git attribute file with auto crlf for all #1654

### LINKED ISSUE #1653: Introduce `.gitattributes` and setup auto crlf as default
## Situation
When a new developer does checkout the project the first time and does not read the GIT setup guide, it can happen, that commits with CR (windows) could be inside a PR and maybe not mentioned by the reviewer.

## Solution
With `.gitattributes` - see https://git-scm.com/docs/gitattributes - it is possible to define auto crlf automatically on checkout:

> If you want to ensure that text files that any contributor introduces to the repository have their line endings normalized, you can set the text attribute to "auto" for all files.
> 
> ` *	text=auto`
> 
>

de-jcup:
This PR
- closes #1653

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*   text=auto


sven-dmlr (approved):
:+1: Let's give it a try ("auto")

PR #1656, opened 2022-09-16
Feature 1655 improve checkmarx cli wrapper

### LINKED ISSUE #1655: Checkmarx CLI wrapper does not use correct mapping and fails on missing optional presetId
## Situation
While writing the PDS solution for checkmarx, we came up with some problems
- preset id is optional , but key must be defined as empty json, because while mapping resolution done, the jackson deserialization failed because of empty string
- test bash script for checkmarx.sh does no longer work
- the mapping for team id done by a executor config file did not work - the mapping was always null

## Wanted
The problems shall no longer estist

## Solution
- [x] change code for preset id handling: The name pattern id provider must be able to handle null or empty strings with
       an empty json `{}` as fallback.
- [x] fix test script
- [x] write a test and try to reproduce the problem with the wrong mapping (+ fix it)

de-jcup:
This PR
- closes #1655

### DIFF
--- a/sechub-integrationtest/src/test/java/com/mercedesbenz/sechub/integrationtest/scenario6/DirectPDSAPICheckmarxWrapperScenario6IntTest.java
+++ b/sechub-integrationtest/src/test/java/com/mercedesbenz/sechub/integrationtest/scenario6/DirectPDSAPICheckmarxWrapperScenario6IntTest.java
@@ -31,7 +31,7 @@ public class DirectPDSAPICheckmarxWrapperScenario6IntTest {
     public Timeout timeOut = Timeout.seconds(600);
 
     @Test
-    public void pds_techuser_can_start_checkmarx_scan_with_job_configuration_and_is_able_to_fetch_result() {
+    public void pds_techuser_can_start_checkmarx_scan_with_job_configuration_and_PDS_job_is_done() {
         /* @formatter:off */
         /* prepare */
         String json =TestFileReader.loadTextFile("src/test/resources/pds/checkmarx-wrapper/checkmarx-pds-job1.json");
--- a/sechub-integrationtest/src/test/java/com/mercedesbenz/sechub/integrationtest/scenario6/DirectPDSAPICheckmarxWrapperScenario6IntTest.java
+++ b/sechub-integrationtest/src/test/java/com/mercedesbenz/sechub/integrationtest/scenario6/DirectPDSAPICheckmarxWrapperScenario6IntTest.java
@@ -1,0 +1,60 @@
+// SPDX-License-Identifier: MIT
+package com.mercedesbenz.sechub.integrationtest.scenario6;
+
+import static com.mercedesbenz.sechub.integrationtest.api.TestAPI.*;
+import static com.mercedesbenz.sechub.test.TestConstants.*;
+
+import java.util.UUID;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.Timeout;
+
+import com.mercedesbenz.sechub.commons.pds.data.PDSJobStatusState;
+import com.mercedesbenz.sechub.integrationtest.api.IntegrationTestSetup;
+import com.mercedesbenz.sechub.integrationtest.api.TestAPI;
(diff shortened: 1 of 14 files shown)


sechub-integrationtest/src/test/java/com/mercedesbenz/sechub/integrationtest/scenario6/DirectPDSAPICheckmarxWrapperScenario6IntTest.java
```diff
@@ -0,0 +1,60 @@
...
+    @Rule
+    public Timeout timeOut = Timeout.seconds(600);
+
+    @Test
+    public void pds_techuser_can_start_checkmarx_scan_with_job_configuration_and_PDS_job_is_done() {
+        /* @formatter:off */
+        /* prepare */
+        String json =TestFileReader.loadTextFile("src/test/resources/pds/checkmarx-wrapper/checkmarx-pds-job1.json");
+
+        String createResult = asPDSUser(PDS_ADMIN).
```
Jeeppler, line 39: The name of the test is about the `pds_techuser` executing the job. Here the `PDS_ADMIN` is used. Using the PDS techuser account would be better, because that is what SecHub usually uses to communicate to the PDS.

sechub-integrationtest/src/test/resources/pds/checkmarx-wrapper/checkmarx-pds-job1.json
```diff
@@ -0,0 +1,36 @@
+{
+    "apiVersion" : "1.0",
+    "sechubJobUUID": "288607bf-ac81-4088-842c-005d5702a9e9", 
+    "productId": "PDS_CHECKMARX_INTEGRATIONTEST",
+    "parameters": [
+        {
+            "key": "pds.scan.configuration",
+            "value": "{\n  \"projectId\": \"sechub_evaluation\",\n  \"codeScan\": {\n    \"fileSystem\": {\n      \"folders\": [\n        \"weaknesses/\"\n      ]\n    }\n  }\n}\n"
```
Jeeppler, line 8: I am not sure, whether there is already a job using the new data section format. In case there is no test, having an integration test with the new data section format would be beneficial.

Jeeppler (changes requested):
Looks good overall. 

I added two comments, which I would like you to review.

sechub-integrationtest/src/test/resources/pds/checkmarx-wrapper/checkmarx-pds-job1.json
de-jcup: ## Current testing situation
There are already integration tests using the new data section format - For example: 

**Example1:**
`PDSCheckmarxIntegrationScenario17IntTest` does it by using the `IntegrationTestTemplateFile` enumeration.
The used Checkmarx mock adapter will inspect the uploaded source code folder path and return a dedicated mock.
The uploaded content is not inspected and does not contain a result. So only path resolving is tested here.

**Example2:**
`PDSBinaryAndSourceCodeHandlingScenario14IntTest` 
does upload source code zip file, extract it and send the results back as findings dynamically.

Example 2 does check the general auto extraction behavior for PDS and new data section
Example 1 does check the file structure given to PDS (checkmarx).

The PDS extraction behaviour is always same - and tested multiple times. In combination with the fact, that

the wrapper relies on the PDS default environment variables wrong usage is nearly impossible (no parameters, so only `java -jar -Dspring.profiles.active=defaul $TOOL_FOLDER/sechub-pds-wrapper-checkmarx.jar  ` is necessary).

## Possibility
If would want to test checkmarx cli wrapper in dynamic way, we would need to change the mock adapter behaviour to provide multiple results - e.g. by a special marker keyword in a reference-id. Than we could change override the checkmarx mock behavior  in this case and use a checkmarx-result.xml file inside the upload sources (for example) and return this.

But... This would 
- change the current mock implementation (I tried to change nothing inside the checkmarx adapter implementation at all) and only use the current implementation in PDS
- makes some effort
- could be done later if really necessary

## Conclusion
I think this is too much effort and the existing tests are enough.

PR #1665, opened 2022-09-26
Update ZAP Extensions #1664

### LINKED ISSUE #1664: Update OWASP ZAP Extensions
There were a lot of versions of all kinds of OWASP ZAP extensions. We should update the list inside the PDS-Solution.

winzj:
- updated ZAP extensions and added the openApi extension explicitly, to make sure it is installed. (But should be installed on default.)
- closes #1664


Jeeppler (approved):
Looks good.

PR #1666, opened 2022-09-26
No longer empty files by archive support ZIP compression #1663

### LINKED ISSUE #1663: Archive support compression creates only empty files
### Situation
For 
- #1415  

the archive support was enhanced with a compression method to "recompress" the received data (makes it possible to filter files before sending to checkmarx).

While implementing #1467 Checkmarx complained about missing sources.

After analyzing the problem, it became clear that the compression mechanism is not correct: The files do exist inside the zip file, but the files are all empty!

(Remark: The wrapper does use the `ArchiveSupport` from commons-archive)

### Wanted
Compress shall compress the files correctly

### Solution
Change the compress implementation and add some tests

de-jcup:
This PR 
- closes #1663


Jeeppler (approved):
Looks good.

PR #1675, opened 2022-09-30
Set CSP header for SecHub server and PDS

### LINKED ISSUE #1672: Set CSP header for SecHub server and PDS
### Situation
Currently there is no CSP header set for the SecHub server or pds.

### Wanted
Set the CSP header for sechub server and PDS.

### Solution
Test different settings/restrictions and set the CSP header inside `AbstractAllowSecHubAPISecurityConfiguration.java`

winzj:
Successfully tested local scan with sechub and pds-gosec. Also verified that the header is set as expected on the sechub server and the PDS.

Alternatively we could set `default-src 'self'` instead of `default-src 'none'`.

- closes #1672


de-jcup (approved):
Looks good for me :-)

Issue #1676, opened 2022-09-30
pds-gosec build fails


sven-dmlr:
# Problem
pds-gosec build fails with:
```
failed to solve with frontend dockerfile.v0: failed to create LLB definition: base name (${BASE_IMAGE}) should not be blank
```

PR #1677, opened 2022-09-30
fix docker build args #1676

sven-dmlr:


Jeeppler (approved):
Looks good. Thanks for fixing the bug.

PR #1693, opened 2022-10-13
Fix PDS upload problems

### LINKED ISSUE #1510: Unable to upload binaries to PDS without shared storage
# Problem

Uploading binaries to SecHub works. However, uploading the same binaries from SecHub to PDS does not work without a shared storage.

# Solution

Implement binary upload from SecHub server to PDS, without using a shared storage.

### LINKED ISSUE #1687: Change PDS configuration key for upload max size
## Situation
With #1351 we changed the upload mechanism for PDS to the new introduced binary upload mechanism of SecHub scheduler.

In SecHub itself, we have two different kind of uploads (2 different REST API methods!) which uses two different upload services:
`SchedulerSourcecodeUploadService` and `SchedulerBinariesUploadService`.

But for PDS we have only one REST API and only one upload service: `PDSFileUploadJobService`

The situation happens when
-  somebody directly uploads to `PDS` by REST API or
-  `pds.config.use.sechub.storage` for a PDS executor is set to `false` . In the latter situation SecHub will download the former uploaded parts from storage, uploads it via REST again to `PDS` when scan starts

### Problems

- `UploadSizeConfiguration` has the entry: `pds.upload.binaries.maximum.bytes` but it is not only for binaries, but for all content
- `PDSFileUploadJobService` exceptions, variable names and also log messages are often containing the term "binaries" inside (which is not correct, because it is for any upload!)

**The Spring Boot standard parameters for multipart handlinge are not used any longer after #1351 had been implemented.**.
E.g.
- spring.servlet.multipart.max-file-size
- spring.servlet.multipart.max-request-size

## Wanted
It shall be clear that there is only ONE configuration possibility for the uploads (source and binary) and that the file size header field "x-file-size" must be always set

## Solution

- [x] change PDSFileUploadJobService , so variables, methods, logs and exceptions are clearly interpreted as general and not only for binary
- [x] change key inside `UploadSizeConfiguration` to `pds.upload.maximum.bytes`
- [x] inspect the existing documentation if there are non generated parts which point to the old `pds.upload.binaries.maximum.bytes` configuration keys change it

- [x] inspect `PDS solutions` if there are side effects/usage of the old configuration keys. If so change it to the new ones

### LINKED ISSUE #1689: PDS Adapter V1 does not set file size header for PDS upload
## Situation
We have introduced `CommonConstants.FILE_SIZE_HEADER_FIELD_NAME` ("x-file-size") and use this to handle
upload big files to S3 storage (see #1471)

We use this information also inside `PDSFileUploadService` for every upload (it is for binary and source, key will be changed see #1687) 

## Problem
Inside PDSAdapterV1 we do not set the additional file header, so the upload mechanism will not have access to the user file size and when sechub storage is not reused by PDS, this will lead to problems with direct uploads to PDS done by SecHub. 

## Solution
- When SecHub upload to PDS is necessary, the field `CommonConstants.FILE_SIZE_HEADER_FIELD_NAME` ("x-file-size") shall always be set
- To have this information available without reading the file from sechub storage we must also introduce two new additional files inside storage : `sourcecode.zip.filesize` and `binaries.tar.filesize`. Such files must be automatically created by scheduler when files are uploaded.

### LINKED ISSUE #1690: Wiremock test for PDS adapter does not really check upload URL calling
## Situation
To have a red test before implementing #1689 I changed the wiremock test and tried to check for the `x-file-size` ... but .. the test did not fail.
After some analyze it became clear that the wiremock stubbing did work, but there was no automated verification (as it was in the past ?).

## Wanted
- Ensure upload really happens - but without code duplication in tests etc.
- Ensure all other PDS adapter v1 calls do really happen as expected

## Solution
- Implement a common usable and maintainable verification for testing wire mocks stubs has been called (or something else) inside sechub testing framework
- use it for upload mocking of PDS adapter wiremock tests
- use it for other simulated parts

### LINKED ISSUE #1692: PDSUploadSupport does always use sourcecode zip file name - even for binaries
## Situation 
`PDSUploadSupport` does always use sourcecode zip file name - even for binaries

## Solution
Change implemantation

de-jcup:
This PR 
- closes #1689 
- closes #1690 
- closes #1692 
- closes #1687
- closes #1510


winzj (changes requested):
Looks good to me, 
I just found 2 occurences of the old env variable `PDS_UPLOAD_BINARIES_MAXIMUM_BYTES`.

Inside the hidden files (starting with a dot) `sechub/sechub-pds-solutions/gosec/.env-single` and 
`sechub/sechub-pds-solutions/findsecuritybugs/.env`, the env variable is still defined as `PDS_UPLOAD_BINARIES_MAXIMUM_BYTES`

de-jcup:
I think both mentioned files (.env and .env-single) are only local because of starts at your machines
- see https://github.com/mercedes-benz/sechub/tree/develop/sechub-pds-solutions/findsecuritybugs 
- see https://github.com/mercedes-benz/sechub/tree/develop/sechub-pds-solutions/gosec

winzj (approved):
You are correct it was my mistake those are generated files.

Issue #1632, opened 2022-09-09
Add Message Script to PDS Base Image


Jeeppler:
Add the message [script](https://mercedes-benz.github.io/sechub/latest/sechub-product-delegation-server.html#example-shared-messaging) to the PDS Base Image as library which pds-solution scripts can import with `source`.

PR #1729, opened 2022-11-07
Added message.sh script to helper folder

Jeeppler:
Added message.sh script to helper folder

closes: #1632


sechub-pds-solutions/pds-base/docker/PDS-Debian.dockerfile
```diff
@@ -176,6 +177,9 @@ COPY run.sh /run.sh
 # Copy the additional "hook" script into the container
 COPY run_additional.sh /run_additional.sh
 
+# Copy run script into the container
+COPY /helper "$HELPER_FOLDER"
```
sven-dmlr, line 181: This would copy the absolute path. I'd guess we want the relative path:
```suggestion
COPY helper/ "$HELPER_FOLDER"
```

sven-dmlr (changes requested):
Looks good.
Please check my comment regarding absolute path when copying into the container.

sechub-pds-solutions/pds-base/docker/PDS-Debian.dockerfile
Jeeppler: Thanks

sven-dmlr (approved):

PR #1730, opened 2022-11-07
Feature 1728 update zaproxy version in pds solution

### LINKED ISSUE #1728: Update PDS OWASP ZAP to ZAP version 2.12.0
### Situation
ZAPROXY version 2.12.0 was released. This might require the update of the java client api to new version 1.11.0.

### Additional information
The official releases are not done very often. This means there were a lot of changes made in comparison to the currently used version 2.11.1. It might require some changes on the PDS solution or the OWASP ZAP wrapper so the PDS solution still works after this update. Also some extensions might need to be updated.

### Wanted
Use the new ZAPROXY version 2.12.0 inside the PDS solution and make sure the PDS solution still works as intended.

winzj:
- closes #1728


sechub-pds-solutions/owasp-zap/docker/run.sh
```diff
@@ -50,7 +50,7 @@ localserver () {
         echo " * Accesskey: $S3_ACCESSKEY"
     fi
 
-    nohup owasp-zap -daemon -host 127.0.0.1 -port 8080 -config api.key="$ZAP_API_KEY" >> /dev/null &
+    nohup owasp-zap -daemon -silent -nostdout -host 127.0.0.1 -port 8080 -config api.key="$ZAP_API_KEY" &
```
Jeeppler, line 53: Please add a comment regarding the `-silent` option. The `-silent` option disables the telemetry collection of the call home addon: https://www.zaproxy.org/docs/desktop/addons/call-home/

That is an important topic.

Jeeppler (changes requested):
See comment.

Jeeppler (approved):
Good job 🥇 

Go ahead merge it.

PR #1740, opened 2022-11-11
When empty zip file uploaded a HTTP 400 is sent #32

### LINKED ISSUE #32: Handle empty source uploads in dedicated way
Checkmarx adapter does currently fail when source files are empty.

E.g. : 
```
Caused by: com.daimler.sechub.adapter.AdapterException: __[SECHUB-UID:18b606d1-6729-486f-83d1-50ea0c7634ed]__ CheckmarxAdapterV1:The queuing has failed:Failed to start scanning due to one of following reasons: source folder is empty, all source files are of an unsupported language or file format
```
On response we got only one finding with text:
"description" : "Security product 'CHECKMARX' failed, so cannot give a correct answer.",

We could handle this in future in one of the following variants:

1. Show a dedicated finding (warning) that no sources where uploaded 
    and do not call any product . So no matter which product this would be always same.
2. Let it as is - means being product specific response.... not nice
3. SecHub accepts this always as "green" - no source, no cry...
So no matter which product this would be always same.

de-jcup:
This PR 
- closes #32


sven-dmlr (changes requested):
We need to re-think this solution.
Please see my comment in the issue: https://github.com/mercedes-benz/sechub/issues/32#issuecomment-1311919822

sven-dmlr (approved):
After discussion: approving changes

PR #1742, opened 2022-11-11
Added `pds.debug.enabled` as default job parameter #1741

### LINKED ISSUE #1741: Provide `pds.debug.enabled` as standard job parameter for PDS
## Situation
We have introduced `PDS_DEBUG_ENABLED` while implementing checkmarx PDS wrapper. But currently the variable is only usable when a PDS instance is fresh deployed - but admins cannot change the value per runtime as a job parameter

## Wanted
As an administrator I want to set `pds.debug.enabled` inside my job parameters.

## Solution
Add the `pds.debug.enabled` as a job parameter which is always sent to `PDS`

de-jcup:
This PR
- closes #1741


Jeeppler (approved):
Looks good. Small, but useful feature 💯

PR #1743, opened 2022-11-11
Feature 1723 findsecuritybugs use pds base image

### LINKED ISSUE #1723: FindSecurityBugs+PDS use PDS+Base Image
Use the PDS Base Image for the FindSecurityBugs+PDS image.

sven-dmlr:
- made pds-findsecuritybugs use the pds base image
- made pds-findsecuritybugs mocked results working
- small adaptions for pds-gosec, pds-scancode to have similar code (pds-heartbeat handling)
- numerous whitespace changes (from IDE's json/yaml plugin)

closes #1723


sechub-pds-solutions/findsecuritybugs/docker/FindSecurityBugs-Debian.dockerfile
```diff
@@ -7,69 +7,44 @@ FROM ${BASE_IMAGE}
...
 
-# Update image and install dependencies
-ENV DEBIAN_FRONTEND noninteractive
+USER root
 
-RUN apt-get update && \
+RUN export DEBIAN_FRONTEND=noninteractive && \
+    apt-get update && \
     apt-get --assume-yes upgrade  && \
     apt-get --assume-yes install dos2unix unzip wget openjdk-17-jre-headless libxml2-utils tree && \
```
Jeeppler, line 27: We now have two openjdk's installed. I am not sure what will be used for FindSecBugs. On the other hand, we could have two PDS-Base Image builds, one with Java 11 and one with Java 17 (both are LTS versions). 

FindSecBugs depends on Java 17, otherwise it cannot analyze code compiled with Java 17. However, FindBugs running on Java 17 can analyze code compiled with Java 11.

The `tree` command is already included in the base image.
 
```suggestion
    apt-get --assume-yes install dos2unix unzip wget libxml2-utils && \
```

Jeeppler (changes requested):
See comment about Java 17.

Jeeppler (approved):

sechub-pds-solutions/findsecuritybugs/docker/FindSecurityBugs-Debian.dockerfile
sven-dmlr: Good point.

Since openjdk-17-jre-headless is installed as last java version, it becomes the default via /etc/alternatives/java
```
pds@pds-findsecuritybugs-66fc486486-xlk9n:/workspace$ java -version
openjdk version "17.0.4" 2022-07-19
OpenJDK Runtime Environment (build 17.0.4+8-Debian-1deb11u1)
OpenJDK 64-Bit Server VM (build 17.0.4+8-Debian-1deb11u1, mixed mode, sharing)
```
So we are using Java 17 as desired.

Removed `tree` package here.

PR #1773, opened 2022-11-22
Improve PDS auto ZIP/TAR handling

### LINKED ISSUE #1172: Provide posiblity to define if sources and/or binaries data type is supported by PDS
## Situation
This is a kind of sub issue of #729,  #1098 but mainly of #1166

At the moment, every code scan will insist on fetching `SOURCE.zip` from s3.
Because we introduce `BINARY.tar` for binary uploads, we must have a possibility to define which one shall be handled

## Wanted
We introduce new configuration entries

## Solution

- ~Introduce new enum `StorageType` with SOURCE and BINARY inside~
- ~Introduce `StorageSupport` class which will provide logic to fetch streams automatically and also to 
  handle missing mandatory storages (depending on type)~

- PDS shall use the the runtime product executor configuration to determine the supported data storage types
- At PDS configuration we introduce `pds.config.supported.datatype=$typedefinition`.  
  The type definition can be either `binary` ,  `source` or 'none'. Also a combination as comma separated entries - e.g. `binary,source` is possible . It can be either optional or mandatory - it does not matter. PDS solutions should have a default value here. If nothing is defined (neither in executor configuration or by a default value inside the PDS configuration) there will be a fallback to the current situation (binary AND source are downloaded + a WARNING will be logged about a missing configuration, but only one time at PDS instance start)
- We will reuse `SecHubDataConfigurationType` enum here
- When a data type is NOT supported, than we do NOT extract or download those files
- When a data type is supported we try to extract and download the file

### LINKED ISSUE #1744: PDS shall inspect ZIP/TAR content and call launcher script only when necessary data available
## Situation
Currently with #32 we only check that there is no empty file upload. That's okay, we do not want to unpack the ZIP at SecHub side and at least empty zip files are handled. But ZIP files (or TAR files) being empty, but containing folder structures etc. is still an issue.

Also filtering at PDS side is an issue: For example: If a PDS solution has filtered the files while unzipping and there is no data after the  unpacking 

## Wanted
The PDS shall only call the launcher script when necessary data is available.

## Solution 
With #1172 we will have the possibility to check by code, if there is a need for source code and/or binaries.
Having this meta information available at PDS side, we can simply check if the necessary parts are available by 
checking the dedicated unzip/untar folders after filtration has been done.

If no content found inside the necessary folders, the launcher script will just not be called and we return OK, but with an empty result.

### LINKED ISSUE #1774: Drop "extractUploads" configuration from PDS config file
## Situation
With #1172 in combination with #1773 we want to 
- download TAR/ZIP
- extract TAR/ZIP
- call launcher script

only when after filtering etc. the corresponding folder is NOT empty.
So we ALWAYS need the extraction (and it not really made sense to have the ZIP/TAR still only available

## Wanted
The configuration shall be removed

## Solution
Change implementation, ensure documentation does not mention it any longer

de-jcup:
This PR
- closes #1172 
- closes #1744 
- closes #1774


sechub-doc/src/docs/asciidoc/documents/pds/product_delegation_server_config_example1.json
```diff
@@ -25,7 +25,10 @@
                     {
                         "key" : "product1.add.tipoftheday",
                         "description" : "boolean as string, when 'true' we add always a tip of the day as info to results"
-                    }  
+                    }, {
+                        "key" : "pds.config.supported.datatype", //<11>
```
winzj, line 29: This should be plural form:
```suggestion
                        "key" : "pds.config.supported.datatypes", //<11>
```

sechub-doc/src/docs/asciidoc/documents/pds/product_delegation_server_config_example1.json
```diff
@@ -34,8 +37,15 @@
         {
             "id" : "PRODUCT_2",
             "path" : "/srv/security/scanner2.sh",
-            "scanType" : "infraScan" ,
-            "extractUploads" : false //<11>
+            "scanType" : "infraScan",
+            "parameters" : {
+                "mandatory" : [
+                    {
+                        "key" : "pds.config.supported.datatype", //<12>
```
winzj, line 44: Please see the comment above above:
```suggestion
                        "key" : "pds.config.supported.datatypes", //<12>
```

winzj (changes requested):
This can improve performance of the PDS solutions. Looks very good. 🚀 

I just found two things to change inside the documentation. Please have a look at the comments.

winzj (approved):
Looks good to me 👍

PR #1784, opened 2022-11-25
Feature 1757 start owasp zap for every scan

### LINKED ISSUE #1757: PDS-OWASP-ZAP: owasp process in separate container
## Situation
pds-owasp-zap image:
Currently we have a Kubernetes POD with one container where both, the PDS _and_ the OWASP-ZAP processes run.

## Problem
We do not want this. The Kubernetes rule of thumb is: "only one process per container".

In this case, Kubernetes wouldn't notice when the OWASP-ZAP process dies.

## Solution
Create a POD with two containers:
* container1: PDS server (small with few ressources)
* container2: OWASP-ZAP server process (what alive check can we use here for Kubernetes?)

winzj:
- closes #1757


sechub-pds-solutions/owasp-zap/docker/scripts/owasp-zap.sh
```diff
@@ -1,11 +1,32 @@
...
+if [ $? -ne 0 ]
+then
+    echo "OWASP-ZAP did not start after waiting for 2 minutes"
+    shutdownZAP
+    exit 1
+fi
+echo "OWASP-ZAP started"
+
+# Execute scan
+sechub_job_uuid=$(echo "$PDS_JOB_WORKSPACE_LOCATION" | cut -d "/" -f 4)
```
Jeeppler, line 29: The sechub_job_uuid is available as variable inside the PDS. The variable is called `SECHUB_JOB_UUID`.

See: https://mercedes-benz.github.io/sechub/latest/sechub-product-delegation-server.html#generated-variables

Jeeppler (changes requested):
See comments.

Jeeppler (approved):
Good job 👍 . Let's get it merged

PR #1809, opened 2022-12-08
Extend CSP header settings to allow CSS

### LINKED ISSUE #1808: HTML report downloaded in browser shows no trafficlight or any other style information
## Situation
While testing #1807 I became aware, that the HTML report does no longer show any style information and looks very ugly when directly downloaded from server (e.g. when using DAUI and downloading a report directly via HTTPS with Browser)
When storing the file locally and showing it inside the browser everything works fine!

## Wanted
HTML report shall have styling gain

## Analyze
With #1672 we have set a CSP - maybe this was the reason.

## Solution
Setup CSP definition inside `AbstractAllowSecHubAPISecurityConfiguration.java` in a way where the HTML report can be downloaded and shown directly from server.

winzj:
- closes #1808


de-jcup (approved):
LGTM

PR #1810, opened 2022-12-09
Rename field uri to url for the webscan config

### LINKED ISSUE #1793: Rename uri to url inside the sechub webscan json config
### Situation
Currently the the sechub webscan json config looks like this:
```json
{
    "apiVersion": "1.0",
    "webScan": {
        "uri": "https://productfailure.demo.example.org",
        "login": {
            "url": "https://productfailure.demo.example.org/login",
           .....
        }
    }
}
```
The target to scan is entered into a field called `uri` but the for the login the field is called `url`.
This can lead to confusion for users.

## Wanted
It should be as easy as possible for users to understand the configuration and make as few mistakes as possible.

### Idea
We could use a Jackson annotation to enable both names for both fields, like `@JsonAlias({"url"})` and `@JsonAlias({"uri"})` or  `@JsonAlias({"uri", "url"})`

winzj:
- closes #1793


de-jcup (approved):
LGTM

PR #1826, opened 2022-12-16
workerThreadCount is pinned to 1 #1825

### LINKED ISSUE #1825: pds-owaspzap: only one scan at a time
## Situation
Currently the number of parallel scans in pds-owaspzap is configurable.

## Problem
Only one scan session at a time is possible due to the central use of Firefox (Session cookies etc.).

## Solution
Remove the configurability of parallelism out of pds-owaspzap's Helm chart.
(`pds.config.execute.workerThreadCount`)

sven-dmlr:
closes #1825


winzj (approved):
Looks good, a fixed value definitely makes more sense here. 👍

PR #1869, opened 2023-01-17
Adjust scripts to MacOS

### LINKED ISSUE #1859: Some scripts cause issues on MacOS (Bash 3.2)
# Problem
MacOS by default uses an older version of Bash (in my case it's `3.2.57(1)-release`). Some of the scripts in this repository are not compatible with that. Additionally, the `SecHubClientExecutor` class only knows how to handle Windows and Linux when deciding which SecHub binary to use.

# Solution
1. Update scripts to use features that are backward compatible between Bash 3.X, 4.X, and 5.X.
2. Update `SecHubClientExecutor` to detect other operating systems and architectures.

jplewa:
In this PR I fixed the following issues:
1. `head -n -1` - negative values are not accepted as arguments of `head -n` in Bash 3. I replaced it with `sed \$d` as per [this SO thread](https://stackoverflow.com/a/18127797/12464594).
2. `${package_split[-1]}` - negative array indices are not supported in Bash 3. I changed it to `${package_split[*]: -1}` as per [this SO thread](https://stackoverflow.com/a/16109755/12464594).
3. `local PROJECT_ID="${1,,}"` - this doesn't work in Bash 3 either. I changed it to `"$(tr [A-Z] [a-z] <<< "$1")"` as per [this SO thread](https://stackoverflow.com/a/2264435/12464594).
4. `rm $PDS_JOB_USER_MESSAGES_FOLDER -rf` - providing options after arguments does not seem to work in Bash 3. I changed it to `rm -rf $PDS_JOB_USER_MESSAGES_FOLDER`.
5. I expanded `SecHubClientExecutor` to handle multiple architectures. Instead of a custom implementation I added an Apache Commons Lang3 dependency. **Note**: Apache Commons Lang3 has a `ArchUtils` class, but this class doesn't know how to handle `aarch64` (which is what I have) or other ARM architectures.

closes #1859


Jeeppler:
@jplewa Thanks for your contribution.

What is the purpose of adding the Apache Commons Lang3 dependency if it does not know how to handle aarch64 (ARM 64-bit)?

jplewa:
> What is the purpose of adding the Apache Commons Lang3 dependency if it does not know how to handle aarch64 (ARM 64-bit)?

Detecting the OS :) The `SystemUtils` class is used to replace your custom `OperationSystem` implementation that only detected Windows

sven-dmlr (changes requested):
@jplewa
Welcome as a first time contributor to SecHub! :tada: 
We appreciate your help, especially as currently none of the core-contributors is working with MacOS.

 1 Interesting: Apple seems not to use GNU `head`.
 2+3 Bash 3 is quite old, but anyway it's great to have support for it. (still works for Bash 5)
 4 OK
 5 Cool!

Please check my comments.

jplewa:
> 2+3 Bash 3 is quite old, but anyway it's great to have support for it.

For some reason it's still the default version available on MacOS :/

> still works for Bash 5

I verified all the commands on my personal Linux Mint computer as well :)

sven-dmlr (approved):
After discussion with the core team, it's fine to add Apache Commons Lang3 library.

:+1: Thanks for your contribution

PR #1929, opened 2023-02-08
container image improvements #1928

### LINKED ISSUE #1928: pds-scancode: improve run_additional.sh script behavior
## Problem
`run_additional.sh` has been introduced with https://github.com/mercedes-benz/sechub/pull/1924.
If the `scancode --version` call fails, then an exit 1 is done which kills the container and k8s will try to launch a new one.
Problem fixing is difficult because the container is gone.

## Solution
Do no exit in `run_additional.sh`. Just print the versions.

sven-dmlr:
- removed exit command from run_additional.sh
- re-arranged some parts in the dockerfile (deb packages should be updated more often than scripts and templates)
- typo in variable name fixed

closes #1928


Jeeppler (approved):
Looks good.

PR #1933, opened 2023-02-10
sechub solution: make helmchart work #1881

### LINKED ISSUE #1881: Make Helm chart work with Docker image from sechub-solution
## Problem
As discussed [here](https://github.com/mercedes-benz/sechub/issues/1567#issuecomment-1397963332):
* Currently there is no Docker image uploaded to ghcr.io/mercedes-benz/sechub/server. So the Helm Chart does not work with the default values.yaml.
* When building the Docker image from `sechub-solution/docker/`, the deployment with the Helm chart fails.

## ToDo
* update `run.sh` script so it is deployable with the Helm chart.
* Provide a Docker image of current SecHub server on ghcr.io/mercedes-benz/sechub/server

sven-dmlr:
- scripts created for
  - build Docker image
  - push Docker image
- run.sh reworked (changed to Bash because of 
- Dockerfile improved
- Helm chart improved
- moved common scripts to `sechub-solutions-shared/scripts`

closes #1881


haerter-tss (approved):

sechub-solution/docker/run.sh
```diff
@@ -1,119 +1,175 @@
...
-check_variable () {
-    value="$1"
-    name="$2"
+init_scheduler_settings() {
+  if [ -z "$SECHUB_CONFIG_TRIGGER_NEXTJOB_DELAY" ] ; then
+    export SECHUB_CONFIG_TRIGGER_NEXTJOB_DELAY="10000"
+    echo "---> SECHUB_CONFIG_TRIGGER_NEXTJOB_DELAY is undefined. Falling back to $SECHUB_CONFIG_TRIGGER_NEXTJOB_DELAY"
+  fi
+
+  # Randomize initial delay used for trigger next job
```
Jeeppler, line 37: I neither understand the comment nor really the code. What is the main point of all of this?

sechub-solution/docker/run.sh
```diff
@@ -1,119 +1,175 @@
...
+- administration url: $SECHUB_SERVER_ADMINISTRATION_BASEURL
+- Upload source code maximum size: $SPRING_SERVLET_MULTIPART_MAX_FILE_SIZE
+- Upload binaries maximum bytes: $SECHUB_UPLOAD_BINARIES_MAXIMUM_BYTES
+- Job scheduling is activated every ${SECHUB_CONFIG_TRIGGER_NEXTJOB_DELAY}ms
+- Job scheduling initial delay: ${SECHUB_CONFIG_TRIGGER_NEXTJOB_INITIALDELAY}ms
+
+Starting up SecHub server
+EOF
+  java $java_debug_options $storage_options \
+    -Dfile.encoding=UTF-8 -Djava.security.egd=file:/dev/urandom -XX:InitialRAMPercentage=50 -XX:MaxRAMPercentage=80 -XshowSettings:vm \
```
Jeeppler, line 135: "Typically, if we're running Java 8 or later and are running on Linux, then our JVM will use file:/dev/urandom by default."(source: https://www.baeldung.com/java-security-egd)

```suggestion
    -Dfile.encoding=UTF-8 -XX:InitialRAMPercentage=50 -XX:MaxRAMPercentage=80 -XshowSettings:vm \
```

sechub-solution/helm/sechub-server/templates/deployment.yaml
```diff
@@ -149,7 +149,7 @@ spec:
         - name: SECHUB_SERVER_SSL_KEYSTORE_TYPE
           value: "PKCS12"
         - name: SECHUB_SERVER_SSL_KEYSTORE_LOCATION
-          value: /home/javarun/{{ .Values.server.ssl.keystoreAlias }}.p12
+          value: /sechub/{{ .Values.server.ssl.keystoreAlias }}.p12
```
Jeeppler, line 152: If we want to use certificates in the future, we probably will mount them. For mounting it would be nice to have a directory like `/sechub/certs/xyz.p12` or `/certs/xyz.p12`. The mount point should be clear and it should be difficult to overwrite any existing directory.

Jeeppler (changes requested):
See comments.

sechub-solution/10-create-image.sh
```diff
@@ -0,0 +1,62 @@
...
+  FAILED=true
+fi
+
+if $FAILED ; then
+  usage
+  exit 1
+fi
+
+BUILD_ARGS="--build-arg BASE_IMAGE=$BASE_IMAGE --build-arg SECHUB_VERSION=$SECHUB_SERVER_VERSION"
+echo ">> Base image: $BASE_IMAGE"
```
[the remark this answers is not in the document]
sven-dmlr: Added verbose output of what is done here.

sechub-solution/10-create-image.sh
```diff
@@ -0,0 +1,62 @@
...
+SECHUB_SERVER_VERSION="$4"
+
+usage() {
+  cat - <<EOF
+
+usage: $0 <docker registry> <version tag> <base image> <SecHub server release version>
+Builds a docker image of SecHub server
+for <docker registry> with tag <version tag>.
+Required:
+- <base image> ; for example debian:11-slim
```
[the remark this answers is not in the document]
sven-dmlr: Will wait until https://github.com/mercedes-benz/sechub/pull/1931 is merged an then split it up into two scripts:
* Alpine and
* Debian

sechub-solution/20-push-image.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
sven-dmlr: Done

sechub-solution/docker/SecHub-Debian.dockerfile
```diff
@@ -158,14 +158,17 @@ ENV SECHUB_STORAGE_SHAREDVOLUME_UPLOAD_DIR="/shared_volumes/uploads"
 
 ARG SECHUB_FOLDER="/sechub"
 
-# non-root user
 # using fixed group and user ids
 RUN groupadd --gid "$GID" "$USER" && \
     useradd --uid "$UID" --gid "$GID" --no-log-init --create-home "$USER"
 
-RUN mkdir --parent "$SECHUB_FOLDER" "$SECHUB_STORAGE_SHAREDVOLUME_UPLOAD_DIR"
+RUN mkdir --parent "$SECHUB_FOLDER/secrets" "$SECHUB_STORAGE_SHAREDVOLUME_UPLOAD_DIR"
```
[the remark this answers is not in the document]
sven-dmlr: done

sechub-solution/docker/run.sh
```diff
@@ -1,119 +1,175 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
sven-dmlr: Of course, Alpine also provides a Bash package.
Although it is desirable to use `/bin/sh` when ever possible.

I expected e.g. `export SECHUB_CONFIG_TRIGGER_NEXTJOB_INITIALDELAY=$(( $(shuf -i 1-$amountOn303 -n 1) * 303 ))` not to work properly.
But surprisingly it works. :-)
Changed to sh.

sechub-solution/docker/run.sh
sven-dmlr: Please see documentation: https://mercedes-benz.github.io/sechub/latest/sechub-operations.html#section-general-configuration 
Look into Table 14

sechub-solution/docker/run.sh
sven-dmlr: ok. removed `-Djava.security.egd`

sechub-solution/helm/sechub-server/templates/deployment.yaml
sven-dmlr: SSL certificates are already working.
`/sechub/{{ .Values.server.ssl.keystoreAlias }}.p12` is a symlink which is mounted in line 93f.
See also `run.sh` lines 112ff.

Why a symlink?
Because if the file `$SECHUB_SERVER_SSL_KEYSTORE_LOCATION` is absent, then Spring will create a self signed certificate. This is good if no certificate is provided.

sechub-solution/docker/run.sh
Jeeppler: The point is not the trigger next job mechanism. My problem is, I do not understand why you are calculating with 303. What is so special about 303. Is this just an arbitrary value? What do you want to achieve with this?

In addition, the `shuf` command needs a bit of explanation `shuf -i 1-10 -n 1` creates a random value between 1 and 10.

sven-dmlr: The code stems from @de-jcup ; So I don't know the thoughts about 303.
As I understood it, the code creates a random initial wait interval, so when containers start at the same time, this tries to avoid race conditions regarding e.g. db schema upgrades.

I'll check this and try to simplify if possible.

sven-dmlr: done.

sechub-solution/docker/SecHub-Debian.dockerfile
```diff
@@ -163,6 +163,7 @@ RUN groupadd --gid "$GID" "$USER" && \
     useradd --uid "$UID" --gid "$GID" --no-log-init --create-home "$USER"
 
 RUN mkdir --parent "$SECHUB_FOLDER/secrets" "$SECHUB_STORAGE_SHAREDVOLUME_UPLOAD_DIR"
+# Mounted secret files (like e.g. SSL certificates) go to $SECHUB_FOLDER/secrets. See deployment.yaml file.
```
Jeeppler, line 166: this comment is in the wrong line: `COPY --from=builder "$SECHUB_ARTIFACT_FOLDER" "$SECHUB_FOLDER"` does not do anything with secrets.

Jeeppler (changes requested):
A lot of improvements 💯. 

However, I still found one small issue.

Jeeppler (approved):
No issues found 👍. Please merge it.

PR #1935, opened 2023-02-10
Implement improved connection checks

### LINKED ISSUE #1934: Owasp Zap Wrapper improve target connection checks
## Situation
We ran into the following problem:

The target URL `https://xyz.example.com` is not found HTTP 404. This means the DAST scan is aborted because the target was not reachable.
But the URL `https://xyz.example.com/health` is reachable and shall be scanned.

## Solution
The Owasp Zap wrapper shall use all URLs including the one build using the include  URL, to check if the target is reachable at all.
Improve the target connection checking so the scan is only aborted if none of URLs was reachable.

We send a list of not reachable URLs back as user messages, so the user knows what was going on.

## Example
```json
{
  "apiVersion": "1.0",
  "server"   : "https://sechub.example.org",
  "user"     : "alice",
  "project"  : "gamechanger",
  "webScan"  : {
      "url": "https://www.gamechanger.example.org",
      "includes": [ "/special/include", "/hidden/include" ]
  }
}
```
### Currently
Target connection check looks at the URL provided by the `url` field:
`https://www.gamechanger.example.org`.

If it is not reachable, the scan is aborted.

### Wanted 
With this issue we want to check the connection to all of the possible URLs:
`https://www.gamechanger.example.org`
`https://www.gamechanger.example.org/special/include`
`https://www.gamechanger.example.org/hidden/include`

Only if none of them is reachable the scan is aborted.

winzj:
- closes #1934

### DIFF
--- a/sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/cli/OwaspZapScanExecutor.java
+++ b/sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/cli/OwaspZapScanExecutor.java
@@ -57,15 +57,18 @@ public class OwaspZapScanExecutor {
     }
 
     private boolean isSiteCurrentlyReachable(OwaspZapScanContext scanContext, URL url, int maxNumberOfConnectionRetries, int retryWaittimeInMilliseconds) {
+        if (connectionChecker.isTargetReachable(url, scanContext.getProxyInformation())) {
+            return true;
+        }
+        // retry if the first connection attempt failed
         for (int i = 0; i < maxNumberOfConnectionRetries; i++) {
-            // do not wait on first try
-            if (i > 0) {
-                wait(retryWaittimeInMilliseconds);
-            }
+            wait(retryWaittimeInMilliseconds);
             if (connectionChecker.isTargetReachable(url, scanContext.getProxyInformation())) {
                 return true;
             }
+
         }
+        // write message to the user for each URL that was not reachable
         scanContext.getOwaspZapProductMessageHelper().writeSingleProductMessage(new SecHubMessage(SecHubMessageType.WARNING,
                 "URL " + url + " was not reachable after trying " + maxNumberOfConnectionRetries + " times. Might not be scanned."));
         return false;
--- a/sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/cli/OwaspZapScanExecutor.java
+++ b/sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/cli/OwaspZapScanExecutor.java
@@ -8,6 +8,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.zaproxy.clientapi.core.ClientApi;
 
+import com.mercedesbenz.sechub.commons.model.SecHubMessage;
+import com.mercedesbenz.sechub.commons.model.SecHubMessageType;
 import com.mercedesbenz.sechub.owaspzapwrapper.config.OwaspZapClientApiFactory;
 import com.mercedesbenz.sechub.owaspzapwrapper.config.OwaspZapScanContext;
 import com.mercedesbenz.sechub.owaspzapwrapper.config.ProxyInformation;
@@ -64,6 +66,8 @@ public class OwaspZapScanExecutor {
                 return true;
             }
         }
+        scanContext.getOwaspZapProductMessageHelper().writeSingleProductMessage(new SecHubMessage(SecHubMessageType.WARNING,
+                "URL " + url + " was not reachable after trying " + maxNumberOfConnectionRetries + " times. Might not be scanned."));
         return false;
     }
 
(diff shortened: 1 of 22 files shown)


sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/cli/OwaspZapScanExecutor.java
```diff
@@ -39,15 +38,29 @@ public void execute(OwaspZapScanContext scanContext) throws ZapWrapperRuntimeExc
 
     }
 
+    private void assertTargetReachable(OwaspZapScanContext scanContext) {
+        boolean isReachable = false;
+        for (String url : scanContext.getOwaspZapURLsIncludeList()) {
```
Jeeppler, line 43: Did I understand this, you are iterating over all URLs from the include list? Meaning if I have 500 urls in the list it would iterate until at least one URL is reachable?

sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/cli/OwaspZapScanExecutor.java
```diff
@@ -39,15 +38,29 @@ public void execute(OwaspZapScanContext scanContext) throws ZapWrapperRuntimeExc
...
+            URI uriToCheck = URI.create(url);
```
Jeeppler, line 44: This can go wrong, then you get an exception. I would prefer doing the conversation before the connection check. I would actually expect `getOwaspZapURLsIncludeList` to return URLs and not strings.

sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/cli/OwaspZapScanExecutor.java
```diff
@@ -39,15 +38,29 @@ public void execute(OwaspZapScanContext scanContext) throws ZapWrapperRuntimeExc
...
+            if (connectionChecker.isTargetReachable(uriToCheck, scanContext.getProxyInformation())) {
+                isReachable = true;
+                break;
```
Jeeppler, line 47: I would use a while loop instead of a for loop and put the break condition in the loop statement:

~~~
while( urls.hasNext() && isReachable != true) {
…
}
~~~

sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/config/OwaspZapScanContext.java
```diff
@@ -253,6 +277,9 @@ public OwaspZapScanContext build() {
 
             owaspZapBasicScanConfiguration.apiDefinitionFile = this.apiDefinitionFile;
 
+            owaspZapBasicScanConfiguration.owaspZapURLsIncludeList.addAll(this.owaspZapURLsIncludeList);
```
Jeeppler, line 280: At this point you could convert all the Strings to proper URLs.

sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/cli/OwaspZapScanExecutor.java
```diff
@@ -39,15 +38,29 @@ public void execute(OwaspZapScanContext scanContext) throws ZapWrapperRuntimeExc
 
     }
 
+    private void assertTargetReachable(OwaspZapScanContext scanContext) {
+        boolean isReachable = false;
+        for (String url : scanContext.getOwaspZapURLsIncludeList()) {
+            URI uriToCheck = URI.create(url);
+            if (connectionChecker.isTargetReachable(uriToCheck, scanContext.getProxyInformation())) {
```
Jeeppler, line 45: There should be a retry logic, who knows maybe the URL is only temporarily not reachable (high load, too many requests etc.).

sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/scan/AbstractScan.java
```diff
@@ -354,6 +350,11 @@ protected void loadApiDefinitions() throws ClientApiException {
         }
     }
 
+    protected boolean atLeastOneURLDetected() throws ClientApiException {
```
Jeeppler, line 353: Please add a comment. How does that work? It is really hard to understand based on the calls to the ZAP API call.

Jeeppler (changes requested):
I would prefer to have the URLs converted into a URL type as soon as possible in the application and later use only the URL.

The conversions of Strings to URI to URL is error prone and potentially problematic.

Additionally, includes and excludes should be de-duplicated and added only once. Sets in Java are de-duplicated by default, as sets can only have unique elements.

Furthermore, applications in networks can be unreachable at times. There should be a retry logic, in case a URL is not reachable.

sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/cli/OwaspZapScanExecutor.java
winzj: We shortly discussed this issue and thought of this solution. Of course in case of many includes, that are all not reachable, the connection check can take some time.

Jeeppler (approved):
Looks good to me.

Issue #1940, opened 2023-02-13
Upgrade to Spring Boot 2.7.9


Jeeppler:
Upgrade to [Spring Boot 2.7.9](https://github.com/spring-projects/spring-boot/milestone/287) and upgrade dependencies as well.

Release date is the 23. of Feb.

PR #1991, opened 2023-02-23
Feature 1940 upgrade to spring boot 2.7.9

Jeeppler:
Upgrade to Spring Boot 2.79

- upgrade dependencies
- upgrade Gradle to 8.0.1

closes: #1940


de-jcup:
Spring data jpa has a bug with native queries here. This is the reason why the build fails, see
https://github.com/spring-projects/spring-data-jpa/issues/2812

It will be fixed with spring boot 2.7.10 as mentioned inside
https://github.com/spring-projects/spring-data-jpa/issues/2812#issuecomment-1453933467
(even when I do not find it inside https://github.com/spring-projects/spring-boot/milestone/289)

So either we wait for the upcoming 2.7.10 release or we do insert a gradle workaround to get the newer data bom version.

Jeeppler:
@de-jcup we wait for Spring Boot 2.7.10. It will be released on the 23rd of March.

The one thing we upgrade beforehand is Gradle.

PR #2015, opened 2023-03-06
Fix problem of missing code call hierarchy in reports for scan type `secretScan`

### LINKED ISSUE #2014: Code call hierarchy not shown for scans of type `secretScan`
## Situation
Code call hierarchy not shown for scans of type `secretScan` - in JSON and also in HTML reports

## Wanted
In both reports the information shall be available

de-jcup:
This PR
- closes #2014


winzj (approved):
Looks good to me 👍

PR #2023, opened 2023-03-06
Feature 2018 multi uri target resolver strategies

### LINKED ISSUE #2017: Provide a URI net resolver strategy called "intranet-hostname-starts-with:" prefix
## Situation
Sometimes we have only an IP address range to identify intranet scans or the URI has a specific subdomain name"
Currently we have only the target resolver strategy with prefix "intranet-hostname-ends-with:"

This does not work when we have a test network where we have no names inside.

## Wanted
It shall be possible to define "intranet-hostname-starts-with:10.5." to enable all scans for the complete ipv4 network 10.5.*.*

## Solution
Introduce a new URI target resolver strategy.

### LINKED ISSUE #2018: Provide possibility to define multiple uri target resolver strategies.
## Situation
With #2017 we will have two different strategies for URI target resolving. And there could come even more.

But currently we can use only ONE strategy at the same time

## Wanted
It shall be possible to define zero, one ore many URI target resolver strategies

## Solution
Change implementation inside `TargetResolverService`:

Use the pipe symbol `|` as a divider inside the spring boot values.
For example:

```
-Dsechub.target.resolve.strategy.uri=intranet-hostname-ends-with:internal.acme.org,test.acme.org|intranet-hostname.starts-with:internal.test,10.5.
```

de-jcup:
This PR
- closes #2017 
- closes #2018

### DIFF
--- a/sechub-scan/src/main/java/com/mercedesbenz/sechub/domain/scan/resolve/IntranetStartsWithURITargetResolveStrategy.java
+++ b/sechub-scan/src/main/java/com/mercedesbenz/sechub/domain/scan/resolve/IntranetStartsWithURITargetResolveStrategy.java
@@ -1,0 +1,53 @@
+// SPDX-License-Identifier: MIT
+package com.mercedesbenz.sechub.domain.scan.resolve;
+
+import java.net.URI;
+import java.util.LinkedHashSet;
+import java.util.Set;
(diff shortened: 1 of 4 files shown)


winzj (approved):
Looks all good to me 👍

PR #2044, opened 2023-03-08
HTML report must contain description with call stack

### LINKED ISSUE #2043: HTML report does not contain details and call stack for secrets
## Situation
With #2014 the JSON reports do now contain the code call hierarchy, but the HTML report is not having those information

image

_(remark: the missing cwe-link is okay here, the SARIF data from gitleak does currently not contain CWE information, so no link has to be rendered. But the description is complete empty and this is the problem)_

## Wanted
HTML report shall have the call stack information in same way as for code scans.

de-jcup:
This PR
- closes #2043


winzj (approved):
Looks all good 🚀

PR #2045, opened 2023-03-08
Fix target resolving #2038

### LINKED ISSUE #2038: TargetResolveService does not work as expected
### Situation
If we use `intranet-hostname-ends-with:intranet.example.org,intx.example.com|intranet-hostname-starts-with:10.5`, the second strategy is never used.
I think the problem is in here: https://github.com/mercedes-benz/sechub/blob/728a4f9d5c9891b9eddbb8d818e30b1cddedf13f/sechub-scan/src/main/java/com/mercedesbenz/sechub/domain/scan/resolve/TargetResolverService.java#L79

The implemented strategies never return `null`, so this is always true and the `break` is always executed in the first round of the loop.

winzj:
- closes #2038


de-jcup (approved):
LGTM :+1:

PR #2057, opened 2023-03-09
Providing now all execution states in admin status #2039

### LINKED ISSUE #2039: Number of waiting jobs when calling `server_status` is misleading
## Situation:
When calling `server_status`, one of the returned keys is `status.scheduler.jobs.waiting`.
The value is the SUM of the execution states `INITIALIZING` + `READY_TO_START` (from database table `schedule_sechub_job` ).

## Problem:
I would expect to see only the `READY_TO_START` number.
(In state INITIALIZING, the user can still do uploads etc. So these jobs are imho not "waiting" to be scheduled)

## Solution:
I propose to count only the jobs in execution state `READY_TO_START`.

de-jcup:
This PR
- closes #2039


sven-dmlr (approved):
:+1: Looking forward to use it :-)

PR #2094, opened 2023-03-20
Feature 1303 integrate generated sarif model

### LINKED ISSUE #1303: Evaluate SARIF model generation via JSON schema and integrate generated model into importer
## Situation
We have much effort so support a working SARIF model and it is sometimes difficult to adopt the written specification into code - see #1299 for details.

But instead of #1299 we could simply generate a model by using the official JSON schema at 
https://github.com/oasis-tcs/sarif-spec/blob/master/Documents/CommitteeSpecifications/2.1.0/sarif-schema-2.1.0.json

For generation we could use https://github.com/joelittlejohn/jsonschema2pojo (Apache 2.0 License) which provides jackson annotations as well.

## Wanted 
Check how the generated SARIF model looks like and if it is working.
Also try to generation with Optional instead of null values.

## Solution
- create a new  branch
- integrate the `gradle` plugin for jsonschema2pojo inside `sechub-sarif`
- create a task which generates sources (with Optional instead of `null` if possible)
- generate the sources by `gradle` task in a way that we can provide multiple SARIF formats in future.
- try to integrate the new generated model into existing SARIF importer
- if this works well, create a PR

### LINKED ISSUE #2093: Replace sechub-sarif with sarif-java library
## Situation
Currently we use  our own SARIF library - but this is done manually inside `sechub-sarif`. The problem here is, that it is extreme cumbersome to write the specification manually. 

## Wanted
Use a library which provides the same functionality - but with a generic approach and based on origin JSON schema from OASIS. 

## Solution
With https://mvnrepository.com/artifact/de.jcup.sarif.java/sarif-2.1.0/1.1.0 is available and provides a full replacement for sechub-java.

### Advantages
- For every SARIF schema, there will be an own library available inside it's own package. Currently the library does only provide SARIF 2.1.0, but the concept is designed to make it possible to use multiple libraries at the same time. Means in future we can use tools which provide SARIF 2.1.0 and also tools which use other versions (maybe something like a "SARIF-3.0.0" )
- The library version handling is clear and semantic:. `$sarif-version-$library-version`
- The library is available at maven central, so we can use it also inside other projects (e.g. Eclipse+IntelliJ SecHub plugins). 
- The library is owned by @de-jcup (myself), so it is ensured that  version handling, evolution etc. will be developed further
- The library is also MIT, so if necessary we could adopt the library, if there is any need, into SecHub directly

### Steps to do
- Remove the `sechub-sarif` project
- Add the mentioned gradle dependency to SecHub
- Make changes to SARIF importer implementation

winzj:
- closes #1303 
- closes #2093


de-jcup (approved):
LGTM  - :+1:

PR #2139, opened 2023-04-03
Remove attack vector and evidence from isFalsePositive method #2137

### LINKED ISSUE #2137: Remove attack vector and evidence from false-positive handling
### Situation
Currently handling false-positives for a webScan uses the finding attack vector to determine if a finding is a false-positive. The attack vector can change with different versions of rules or different DAST scanners. This means the attack vector can be rule or product specific.

The evidence of the web response part is used as well for the false-positive handling. The field can be product specific as well.

### Solution
- The attack vector will be removed from false-positive handling.
- The evidence part of the web response might be removed as well.

winzj:
- closes #2137


de-jcup (approved):
LGTM :+1:

PR #2176, opened 2023-04-14
Make sure includes are known to OWASP ZAP #2160

### LINKED ISSUE #2160: PDS-OWASPZAP needs to make sure all includes are visited and can be scanned by the OWASP ZAP
### Situation
Currently if we want to scan an API by using includes instead of an openAPI file, OWASP ZAP will not scan all includes. The includes are only used to provide a filter if a new URL is detected. If the newly detected URL is inside matches one of the includes it will be scanned except it matches an excluded entry.

### Solution
We will make sure the OWASP ZAP is aware of the included URLS by accessing the include list through the OWASP ZAP. This way we can ensure that the OWASP ZAP has all the wanted URLs inside his sites tree.
- This can be done by the owasp zap wrapper using the ZAP API.

winzj:
- closes #2160

### DIFF
--- a/sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/scan/AbstractScan.java
+++ b/sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/scan/AbstractScan.java
@@ -388,26 +388,28 @@ public abstract class AbstractScan implements OwaspZapScan {
     }
 
     private void scanUnsafe() throws ClientApiException {
+        /* OWASP ZAP setup on local machine */
         setupBasicConfiguration();
         deactivateRules();
         setupAdditonalProxyConfiguration();
-
         createContext();
+
+        /* OWASP ZAP setup with access to target */
         addIncludedAndExcludedUrlsToContext();
         loadApiDefinitions();
+
+        /* OWASP ZAP scan */
         if (scanContext.isAjaxSpiderEnabled()) {
             runAjaxSpider();
         }
-
         runSpider();
-
         passiveScan();
-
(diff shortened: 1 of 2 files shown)


sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/scan/AbstractScan.java
```diff
@@ -464,9 +466,19 @@ private void renameReportFileIfFileExtensionIsNotJSON() {
         }
     }
 
+    private void visitInclude(String url) {
+        try {
+            String followredirects = "false";
```
Jeeppler, line 471:
```suggestion
            String followRedirects = "false";
```

sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/owaspzapwrapper/scan/AbstractScan.java
```diff
@@ -464,9 +466,19 @@ private void renameReportFileIfFileExtensionIsNotJSON() {
...
+            clientApi.core.accessUrl(url, followredirects);
+        } catch (ClientApiException e) {
+            LOG.error("Trying to access URL {} produced the error: {}", url, e.getMessage());
```
Jeeppler, line 474: `got…` or `received the error`

```suggestion
            LOG.error("While trying to access URL {} got the error: {}", url, e.getMessage());
```

Jeeppler (changes requested):
Small changes. Otherwise it looks good.

Jeeppler (approved):
Good job. Please merge it.

PR #2225, opened 2023-05-05
Feature 2058 get joblist with labels and ordering

### LINKED ISSUE #1553: Do the SpotlessCheck Early
Currently, the spotless check is executed during the build in the GitHub Actions workflow. However, that is inefficient. The spotlessCheck should be done before building anything. Best would be to execute it directly after the `./gradlew clean` step in the beginning.

### LINKED ISSUE #2058: SecHub API: Get jobs with labels
# Feature

With #2040 the user can provide labels, additional meta data, which are available as meta data in the report as well. Be able to query for jobs with specific labels attached to them (see: [UC_071-User lists jobs for project](https://mercedes-benz.github.io/sechub/latest/sechub-architecture.html#section-usecase-UC_071)).

### LINKED ISSUE #2219: Ensure sort ordering for get job list SecHub API result inside integration test
## Situation
~~Some users mentioned that the REST API call for  [UC_071-User lists jobs for project](https://mercedes-benz.github.io/sechub/latest/sechub-architecture.html#section-usecase-UC_071) does currently show up the latest entry at the end inside the JSON response.~~

The results from job list are already sorted with latest on top, but the integration test did not check this.

## Wanted
~~The job list result shall be sorted by timestamp and the latest job shall be the first JSON node element in content.
This makes it easier for users, administrators but also for automation workflows.~~

The integration tests in `FetchJobInfoForUserIntTest` should also inspect the positions.

### LINKED ISSUE #2221: JobList elements have metadata information when parameter `withMetaData=true`
## Situation
The REST API call for [UC_071-User lists jobs for project](https://mercedes-benz.github.io/sechub/latest/sechub-architecture.html#section-usecase-UC_071) does currently not contain meta data information.

## Wanted
The job list elements shall contain meta data inside.

### Solution
The label information is already available - even without #2058: We have the sechub configuration inside the `ScheduleSecHubJob` entities. `SecHubJobInfoForUser` is an artificial result object which is created by information from the related `ScheduleSecHubJob`  entity.
We simply add the meta data node object into the `SecHubJobInfoForUser` entity.

## Solution
- To be backward compatible and to reduce time and load, we introduce a new optional parameter `withMetaData` forthe REST call. Only when `withMetaData=true`  is set, the meta data note will be added.

### LINKED ISSUE #2239: REST documentation does only contain path parameters but no request parameters
## Situation
The generated documentation for REST APi does not contain request parameters but only path parameters.

## Wanted
The documentation shall also contain request parameters

## Solution
Change generation

de-jcup:
This PR
- [x] closes #2219 
- [x] closes #2221
- [x] closes #2058
- [x] closes #2239
- [x] closes #1553


winzj (changes requested):
Looks very good to me. 🚀 

Just found two very small things. Please look at my comments.

winzj (approved):
Looking good, let's get it merged 👍

PR #2228, opened 2023-05-08
Secret scan FP handling, allow empty CWE IDs and provide CWE 798 as secret scan SARIF import fallback #2226, #2028

### LINKED ISSUE #2016: Provide a fallback CWE id for secret scans where no CWE id is set.
## Situation
Some secret scanner products  (e.g. gitleaks) do not provide a CWE id inside their SARIF output.

But CWE ids are essential for false-positive handling etc. Also it's very suitable for the end users to have a link to MITRE web page which explains the problem (even when it's just about not having a (real) secret/credential inside the sources ... )

## Wanted
When no CWE id is defined, but the scan type is `secretScan` a default  scan type shall be defined as a fallback

## Solution
Change sereco tranformer and handle `secretScan` scan type accordingly.
Fallback shall be CWE id: 259 - see https://cwe.mitre.org/data/definitions/259.html

### LINKED ISSUE #2028: False-Positive marking shall work without CWE
# Problem

Findings can be marked as false-positives in the SecHub report. However, that requires that a CWE-ID, which not every tool produces.

Examples include:

- Gitleaks
- FindSecurityBugs (SARIF only)

# Wanted

It should be possible to mark false-positives without the CWE-ID being included in the report.

### LINKED ISSUE #2226: False positive handling does not work for secret scan type
## Situation
For a secret scan the false positive handling is currently not available  - when a user tries to mark it, a failure message appears that this not implemented

## Wanted
Shall work...

## Solution
Implement a false positive strategy for secret scan type

de-jcup:
This PR
- closes #2226 
- closes #2028
- closes #2016

### DIFF
--- a/sechub-integrationtest/src/main/java/com/mercedesbenz/sechub/integrationtest/api/AssertReport.java
+++ b/sechub-integrationtest/src/main/java/com/mercedesbenz/sechub/integrationtest/api/AssertReport.java
@@ -127,6 +127,11 @@ public class AssertReport {
             return this;
         }
 
+        public AssertFinding hasNotId(int id) {
+            autoDumper.execute(() -> assertNotEquals(id, finding.getId()));
+            return this;
+        }
+
         public AssertFinding hasName(String name) {
             autoDumper.execute(() -> assertEquals(name, finding.getName()));
             return this;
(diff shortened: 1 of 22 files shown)


winzj (approved):
Looks good to me. 🚀

PR #2254, opened 2023-05-23
Custom bad request handling for Jackson exceptions #2250

### LINKED ISSUE #2250: Improve bad request handling on JSON problems
## Situation
Some bad request responses contain unnecessary data (from user perspective) about JSON parsing etc.

## Wanted
Bad request handling shall only provide a reduced set of necessary information.

## Solution
Analyze result: The Spring Boot exception handling does use the message from the Jackson exception. But the
jackson exception message method returns much additional information, including trace information etc.

We will handle Jackson exceptions in a custom way and use the original message + location information.

de-jcup:
This PR
- closes #2250


sven-dmlr (approved):
Well done :+1:

PR #2256, opened 2023-05-24
Feature 723 drop spring batch dependencies

### LINKED ISSUE #723: Drop Spring Boot Batch parts
## Situation
With #701 we moved the synchronous execution outside the running Spring Batch job in another Thread to close its transaction.

The execution process was the last remaining part why we used Spring Batch here. We do not need it any longer and it is just a unnecessary complexity. 

## Wanted
Change call mechanism, so we can completely remove Spring Batch.
This will also avoid situations like #499 which was only because of internal Spring Batch changes.

## Solution
- Instead of starting a Spring Batch job the scheduler shall directly call the `SynchronSecHubJobExecutor` which was introduced with #701.
- After this has been done, remove any Spring Batch part. This includes not only code but also the existing Spring Batch database tables!
- Update architecture diagrams and documentation

de-jcup:
This PR
- closes #723


Jeeppler (approved):
You were able to remove many classes, but at the same time you did not need to add as many. Very surprising to me.

Overall it looks good. Please merge it.

PR #2271, opened 2023-05-30
Changed PDS workspace location handling #2255

### LINKED ISSUE #2255: PDS: improve definition of workspace folder
## Situation
Currently, one can define `sechub.pds.workspace.rootfolder`.

In our case it is on a separate file system (/workspace):
```
        -Dsechub.pds.workspace.rootfolder="/workspace"
```
When the PDS does its first scan, a new folder "workspace" is created, the result is a workspace folder **`/workspace/workspace/`** in the container.

So if I wanted "/workspace" as workspace folder, I had to define "/" as sechub.pds.workspace.rootfolder.
- this feels not logical
- also if "/workspace" is on a separate file system, defining "/" will lead to problems if someone decides to name the PDS's workspace subfolder (e.g. "ws/" instead of "workspace/")

## Wanted
I would like to be able to directly define the workspace folder (instead of the parent folder).

Example:
```
        -Dsechub.pds.workspace.folder="/workspace"
```

de-jcup:
This PR
- closes #2255


sven-dmlr (approved):
Looks good to me :+1:

PR #2272, opened 2023-05-30
Implemented fetch user details by email address #1095

### LINKED ISSUE #1095: Admin: Find userID for an email address
## Problem
When creating a new user, it can happen that the email address is already in use.
Currently we can't find out the corresponding SecHub userID easily (get list of all users and list details for each user -> hundreds of api calls).

## Wanted
Have an api call (admins only) that returns the user details of an email address.

## Solution

We introduce `GET /api/admin/user-by-email/$emailAddress` and return a `UserDetailInformation` object as JSON - means same way as handled in `GET /api/admin/user/$userId`. When no user exists for the given email address, a 404 is returned 

- [x] create new Usecase "Admin shows user details for email address"
- [x] Add new REST controller method (only for admins) inside `admin` domain
- [x] Document new REST method (REST doc tests)
- [x] Write tests
- [x] Create DAUI (developer admin ui) action

de-jcup:
This PR
- closes #1095


sven-dmlr (approved):
Looks good + successfully tested :+1:

PR #2276, opened 2023-05-30
Introduce continuous deployment for PDS #2273

### LINKED ISSUE #2273: PDS: provide possiblity for continous deployment
## Situation
When there are many `PDS` solutions running and an administrator wants to update just ONE solution, currently it is necessary to stop the `SecHub` server scheduling, wait for all jobs to be done and then deploy the new `PDS` solution, and start `SecHub` scheduler afterwards.

This is extremely annoying and takes a lot of time.

## Wanted
As an administrator, I want to 
- send `SIGTERM` to the process directly or
- Update a complete `PDS` solution cluster to K8s by helm deployment
 (K8s does also send `SIGTERM` for graceful pod deletion - see https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination) 

and the running `PDS` jobs shall be restarted automatically as soon as there is another PDS member running. There shall be no influence to running `SecHub` jobs which wait for running PDS jobs.

### PDS shutdown behavior
Reading https://docs.oracle.com/javase/9/troubleshoot/handle-signals-and-exceptions.htm#JSTGD356 

> SIGTERM, SIGINT, SIGHUP | These signals are used to support the shutdown hook mechanism (java.lang.Runtime.addShutdownHook) when the VM is terminated abnormally. (Optional)

we will introduce a JVM shutdown hook to handle `SIGTERM` and do following steps here:

1. stop scheduling (queue)
2. collect all running PDS job uuids
3. mark all those jobs as `READY_TO_START` + log info that PDS job must be started again etc.
4. stop /kill all current PDS processes inside the JVM

### PDS query handling
- We do not need any changes here. We use `READY_TO_START` to reset, so the next running PDS server will start
 the job automatically again

### SecHub server side
The `PDSAdapterV1` does already just wait for job to be done/failed. So the server will just wait (the state of the PDS job is inside PDS database and every running PDS cluster member can fetch the state...)

Means: There is nothing necessary on `SecHub` side.

de-jcup:
This PR
- closes #2273

### DIFF
--- a/sechub-pds/src/main/java/com/mercedesbenz/sechub/pds/execution/PDSExecutionService.java
+++ b/sechub-pds/src/main/java/com/mercedesbenz/sechub/pds/execution/PDSExecutionService.java
@@ -7,10 +7,12 @@ import java.time.LocalDateTime;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
+import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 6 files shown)


sven-dmlr (approved):
Looks good to me

PR #2368, opened 2023-07-05
Feature 2175 implement handling headers owaspzap wrapper

### LINKED ISSUE #2175: OWASP ZAP wrapper implement handling headers of sechub webscan configuration
### Situation
With #2174 we extended the SecHubWebScanConfiguration by introducing list of headers.
The OWASP ZAP wrapper needs to implement this feature for the PDS solution OWASP ZAP.

### Solution
Make sure the Owasp Zap Wrapper can handle the header section and perform the configurations at the OWASP ZAP side.
This means, we need to add rules for the different headers and their scopes, by using the [Replacer](https://www.zaproxy.org/docs/desktop/addons/replacer/) ZAP extension.

### Additional
Since it is necessary to install the [Replacer](https://www.zaproxy.org/docs/desktop/addons/replacer/) extension with this issue, we should update the ZAP extensions, since there were some fixes and updates applied since we last updated our list of installed extensions.

winzj:
- closes #2175


Jeeppler (approved):
Looks good. 

However, no tests. Is that intentional.

Jeeppler (approved):
Looks good 💯 

Please merge it.

PR #2380, opened 2023-07-17
SecHubTomcatServletWebserverFactory can now handle MB,KB and B #2379

### LINKED ISSUE #2379: `SecHubTomcatServletWebserverFactory` shall be able to handle MB, KB and B
## Situation

Currently `SecHubTomcatServletWebserverFactory` is obly able to handle definitions which ends with `MB` or `mb`.

For all other parts the error 
> Cannot customize tomcat swallow size automatically,...
is logged and the mechanism is not installed

## Wanted
Handle not only `MB` but also `KB` and `B`

## Solution
Change implementation

de-jcup:
This PR
- closes #2379


sven-dmlr (approved):
Looks good to me. :+1:

PR #2382, opened 2023-07-17
Added banners for PDS and SecHub and improved PDS summary #2026

### LINKED ISSUE #2026: Improve startup text of PDS and SecHub
## Problem:
There is currently no way of knowing which default parameters for the PDS solutions are set and what the exact default value is. 

## Solution:
When starting a PDS solution it would be nice if all default parameters and their values would be shown.

### Bonus:
It would be cool if there was some sort of PDS logo in ASCII art that will be shown on boot :)

de-jcup:
This PR
- closes #2026


sven-dmlr (approved):
Looks good to me :+1:

PR #2412, opened 2023-07-26
Enlarge PDS product ID + handle alphabetic validation #1273 #2411

### LINKED ISSUE #1273: Increase PDS Product Identifier Length
The current length of the PDS Product Identifier is limited to 30 characters. This should be increased to at least 50. As `PDS_INTTEST_PRODUCT_LICENSESCAN` is too long.

### LINKED ISSUE #2411: PDSProductIdentifierValidation alphabetic not handled correctly
### Situation
With #2129 we changed the SimpleStringUtils method accordingly, but `PDSProductIdentifierValidator` does not use it. It uses an older approach which does allow more characters.

### Solution
PDSProductIdentifierValidator shall use the  SimpleStringUtils method changed with #2129.

winzj:
- closes #1273 
- closes #2411


de-jcup (approved):
Looks good to me :+1:

PR #2443, opened 2023-08-03
Fix HTML escaping + integration test now uses client #2442

### LINKED ISSUE #2442: Fix problems with wildcards inside web scan header configuration `onlyForUrls`
## Situation
With #2174 we introduced a wildcard `<*>` for filtering urls by `onlyForUrls`.

We wrote an integration test, but in real server deployment the communication between SecHub client and SecHub server failed when creating a new SecHub job.

The failure message contained information that it is not a valid url - but the text did not contain `<*>` but instead `lt;*` - means html escaped

## Wanted
Find the problem and fix it.

## Solution
- Enhance / or write new integration test to reprocude the problem (and have a failing build)
- Fix the problem

de-jcup:
This PR 
- closes #2442


winzj (approved):
Looking good to me.

Thanks for implementing this very quickly. 🚀

PR #2460, opened 2023-08-07
made .ZAP writeable for pds user #2459

### LINKED ISSUE #2459: pds-owasp-zap: make ~/.ZAP writeable
## Situation
Currently only `~/.ZAP/session` is mounted and writeable for the pds user.
Recently we observed error messages that ~/.ZAP cannot be written.

## ToDo
Mount volume into `~/.ZAP`

sven-dmlr:
- closes #2459


winzj (approved):
Looking good 👍

PR #2469, opened 2023-08-10
Feature 2467 sechub calls pds only when it makes sense

### LINKED ISSUE #2467: SecHub PDS product executors shall only call PDS when it makes sense
## Situation
An administrator can define inside an executor configuration that the product can handle
- source and binary
- only source
- only binary

But currently even when an upload has not provided any of the handled variants, the `PDS` adapter will still be called - even when it makes no sense.

## Wanted
`PDS$variant$ProductExecutor` shall only call the PDS adapter when it makes sense

With executor definitions ('pds.config.supported.datatype')
- `binary`, but source uploaded -> no PDS call
- `source`, but binary uploaded -> no PDS call
- `source, binary` but nothing uploaded -> no PDS call
- `binary`, and binary uploaded -> PDS call
- `source`, and source uploaded -> PDS call
- `source, binary` and source uploaded -> PDS call
- `source, binary` and binary uploaded -> PDS call
- `none` -> always PDS call
- `none, source ,binary ` -> always PDS call
- not defined -> always PDS call
- odd values -> log entry + always PDS call

de-jcup:
This PR
- closes #2467

### DIFF
--- a/sechub-scan-product-pds/src/main/java/com/mercedesbenz/sechub/domain/scan/product/pds/AbstractPDSProductExecutor.java
+++ b/sechub-scan-product-pds/src/main/java/com/mercedesbenz/sechub/domain/scan/product/pds/AbstractPDSProductExecutor.java
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: MIT
 package com.mercedesbenz.sechub.domain.scan.product.pds;
 
+import java.util.List;
+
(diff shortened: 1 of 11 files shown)


winzj (approved):
Looking good 👍 

Verified it with some test scenarios locally, with a real SecHub Server and PDS product.

Issue #2513, opened 2023-08-22
Replace manually written API calls in WebUI with SecHub Java API Client


Jeeppler:
Replace manually written API calls in WebUI with SecHub Java API Client. The SecHub Java API client was introduced as part of the system testing framework #2092.

de-jcup:
Remark:

Inside this issue following was done
- status page does now use real server connection
- now centralized access via access service (no direct access to webclient possible)
- changed structure. For example all parts which are visible on pages are in dedicated sub packages of `com.mercedesbenz.sechub.webui.page`
- some additional i18 parts were added
- user credential injection via spring value is now done directly into access service
- some refactorings
- added SecHub logo to web ui

Issue #469, opened 2020-12-01
Replace usecase "Admin checks server version" with "Admin fetches server runtime data" and provide JSON object


Jeeppler:
The admin checks server version API returns a plain text result instead of the usual JSON: https://daimler.github.io/sechub/latest/sechub-restapi.html#admin-checks-server-version

The problem with that is, that OpenAPI cannot figure out the type of the returned payload.

---
<sup>Jeremias Eppler <jeremias.eppler@daimler.com>, Daimler TSS GmbH, [imprint](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)</sup>

Jeeppler:
Related #377

de-jcup:
We will do following changes here

- [ ] Introduce `ServerRuntimeData` as Json object which will initially hold `serverVersion` inside, but can be extended for future needs.
- [ ] Change url from `/api/admin/info/version` to `/api/admin/info/server`
- [ ] Rename usecase `@UseCaseAdminChecksServerVersion` to `@UseCaseAdminFetchesServerRuntimeData`
- [ ] Update the URL builder (TestAPI)
- [ ] Update the parse mechanism inside `sechub-api.sh` / `admin-api.sh`

de-jcup:
I marked this issue with a "breaking-change" label (because the REST API changed a bit) . But it is only a real breaking change, if somebody has automated administrative parts by their own. We provide and maintain `admin-api.sh` by our own - so it is not really a breaking change.

PR #2518, opened 2023-08-28
Replace manually written API calls with SecHub Java API calls #2513

Jeeppler:
- Replace status page API calls with SecHub Java API calls

closes: #2513
closes: #469

### DIFF
--- a/sechub-api-java/src/main/java/com/mercedesbenz/sechub/api/SecHubClient.java
+++ b/sechub-api-java/src/main/java/com/mercedesbenz/sechub/api/SecHubClient.java
@@ -10,6 +10,9 @@ import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.TreeMap;
 import java.util.UUID;
 import java.util.concurrent.Callable;
 
@@ -19,6 +22,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.fasterxml.jackson.databind.json.JsonMapper;
+import com.mercedesbenz.sechub.api.SecHubStatus.Jobs;
+import com.mercedesbenz.sechub.api.SecHubStatus.Scheduler;
 import com.mercedesbenz.sechub.api.internal.ApiClientBuilder;
 import com.mercedesbenz.sechub.api.internal.OpenApiSecHubClientConversionHelper;
 import com.mercedesbenz.sechub.api.internal.WorkaroundAdminApi;
@@ -275,16 +280,26 @@ public class SecHubClient {
     /* + ................Status........................... + */
     /* ++++++++++++++++++++++++++++++++++++++++++++++++++++ */
     public SecHubStatus fetchSecHubStatus() throws SecHubClientException {
-        SecHubStatus status = new SecHubStatus();
+        SecHubStatus status = null;
+
+        Map<String, String> statusInformation = new TreeMap<>();
+
         runOrFail(() -> {
             List<OpenApiStatusInformationInner> statusInformationList = adminApi.adminListsStatusInformation();
+
             for (OpenApiStatusInformationInner info : statusInformationList) {
                 String key = info.getKey();
                 if (key != null) {
-                    status.statusInformation.put(key, info.getValue());
+                    statusInformation.put(key, info.getValue());
                 }
             }
+
         }, "Was not able to fetch SecHub status!");
+
+        if (!statusInformation.isEmpty()) {
+            status = convertStatusMapToStatus(statusInformation);
+        }
+
         return status;
     }
 
@@ -468,6 +483,19 @@ public class SecHubClient {
         runOrFail(() -> projectApi.userApprovesJob(projectId, jobUUID.toString()), "Job approve");
     }
 
+    /* +++++++++++++++++++++++++++++++++++++++++++++++++ */
+    /* + ................Version...................... + */
+    /* +++++++++++++++++++++++++++++++++++++++++++++++++ */
+
+    public Optional<String> getServerVersion() throws SecHubClientException {
+        runOrFail(() -> {
+            Optional<String> serverVersion = Optional.of(adminApi.adminChecksServerVersion().getServerVersion());
+            return serverVersion;
+        }, "Get server version");
+
+        return Optional.empty();
+    }
+
     /* ++++++++++++++++++++++++++++++++++++++++++++++++++++ */
     /* + ................Helpers......................... + */
     /* ++++++++++++++++++++++++++++++++++++++++++++++++++++ */
@@ -508,5 +536,21 @@ public class SecHubClient {
 
         public void inform(SecHubClientListener listener);
 
+    }
+
+    private SecHubStatus convertStatusMapToStatus(Map<String, String> statusMap) {
+        var schedulerEnabled = Boolean.valueOf(statusMap.get("status.scheduler.enabled")).booleanValue();
+        var jobsAll = Long.valueOf(statusMap.get("status.scheduler.jobs.all"));
+        var jobsCancelRequested = Long.valueOf(statusMap.get("status.scheduler.jobs.cancel_requested"));
+        var jobsCanceled = Long.valueOf(statusMap.get("status.scheduler.jobs.canceled"));
+        var jobsEnded = Long.valueOf(statusMap.get("status.scheduler.jobs.ended"));
+        var jobsInitializing = Long.valueOf(statusMap.get("status.scheduler.jobs.initializing"));
+        var jobsReadyToStart = Long.valueOf(statusMap.get("status.scheduler.jobs.ready_to_start"));
+        var jobsStarted = Long.valueOf(statusMap.get("status.scheduler.jobs.started"));
+
+        SecHubStatus.Scheduler scheduler = new Scheduler(schedulerEnabled);
+        SecHubStatus.Jobs jobs = new Jobs(jobsAll, jobsCancelRequested, jobsCanceled, jobsEnded, jobsInitializing, jobsReadyToStart, jobsStarted);
+
+        return new SecHubStatus(scheduler, jobs);
     }
 }
--- a/sechub-api-java/src/main/java/com/mercedesbenz/sechub/api/SecHubStatus.java
+++ b/sechub-api-java/src/main/java/com/mercedesbenz/sechub/api/SecHubStatus.java
@@ -1,21 +1,10 @@
-// SPDX-License-Identifier: MIT
 package com.mercedesbenz.sechub.api;
 
-import java.util.Collections;
-import java.util.Map;
-import java.util.TreeMap;
+public record SecHubStatus(Scheduler scheduler, Jobs jobs) {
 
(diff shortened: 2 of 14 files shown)


sechub-api-java/src/main/java/com/mercedesbenz/sechub/api/SecHubClient.java
```diff
@@ -129,4 +129,7 @@ public interface SecHubClient {
      */
     void approveJob(String projectId, UUID jobUUID) throws SecHubClientException;
 
-}
\ No newline at end of file
+    String getServerVersion() throws SecHubClientException;
+
+    void userRequestsNewApiToken(String emailAddress) throws SecHubClientException;
```
de-jcup, line 134: This is not an eventhandler where methods are in style like "what happened" but more what the caller WANTS to do.

Her my suggestion:
```suggestion
    void requestNewApiToken(String emailAddress) throws SecHubClientException;
```

sechub-api-java/src/main/java/com/mercedesbenz/sechub/api/SecHubStatus.java
```diff
@@ -1,21 +1,11 @@
...
-import java.util.TreeMap;
+public record SecHubStatus(Scheduler scheduler, Jobs jobs) {
 
-public class SecHubStatus {
-
-    Map<String, String> statusInformation = new TreeMap<>();
-
-    public Map<String, String> getStatusInformationMap() {
-        return Collections.unmodifiableMap(statusInformation);
+    public record Scheduler(boolean isEnabled) {
```
de-jcup, line 6: As mentioned, i am not happy with the record here.

sechub-api-java/src/test/java/com/mercedesbenz/sechub/api/DefaultSechubClientWireMockTest.java
```diff
@@ -59,8 +89,8 @@ public void fetch_sechub_status_with_basic_auth() throws Exception {
         /* test */
         verify(getRequestedFor(urlEqualTo("/api/admin/status")));
         assertNotNull(status);
-        assertEquals("true", status.getStatusInformationMap().get("status.scheduler.enabled"));
-        assertEquals("2", status.getStatusInformationMap().get("status.scheduler.jobs.all"));
+        assertTrue(status.scheduler().isEnabled());
+        assertEquals(2, status.jobs().all());
```
de-jcup, line 93: You added additional entries inside the test json file - so please test them as well.

sechub-server-core/src/main/java/com/mercedesbenz/sechub/server/core/ServerVersion.java
```diff
@@ -0,0 +1,4 @@
+package com.mercedesbenz.sechub.server.core;
+
+public record ServerVersion(String serverVersion) {
```
de-jcup, line 3: I would prefer:
```suggestion
public record ServerVersion(String versionAsString) {
```

## Why?
We have a following situation:

The current implementation is equivalent to:
```java
class ServerVersion{
   public String getServerVersion() {return "0.1.0"}
}
```

it's like 
```java
class Apple {
    public String getApple(){return "i am an apple"}
}
```

means the object stands for something, but when you ask it about this it returns something else... a server version is either an object or an string...

When we use `versionAsString` we have a getter `getVersionAsString`. If we would change in future to something like having more getters (e.g. getMinor(), getMajor() etc.) we could achieve this easily but it would be still the `SeverVersion` we are asking for the information.

sechub-webui-solution/docker/WebUI-Debian.dockerfile
```diff
@@ -0,0 +1,164 @@
...
+ARG WEBUI_VERSION
+ARG BUILD_TYPE
+
+# possible values: 17
+ARG JAVA_VERSION="17"
+
+# Artifact folder
+ARG WEBUI_ARTIFACT_FOLDER="/artifacts"
+
+#-------------------
```
de-jcup, line 20: Could the next comments be removed?

sechub-webui-solution/docker/clone.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env sh
```
de-jcup, line 1: Shouldn't we have this inside the shared script folder ?
Looks like it could be useful in other solutions as well.

sechub-webui-solution/docker/copy/README.adoc
```diff
@@ -0,0 +1,4 @@
+// SPDX-License-Identifier: MIT
```
de-jcup, line 1: Why do we need a PDS here?

sechub-webui-solution/docker/helper/message.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
```
de-jcup, line 1: the message helper file is only for PDS

sechub-webui-solution/docker/run.sh
```diff
@@ -0,0 +1,110 @@
...
+    PID_JAVA_SERVER=$!
+    wait "$PID_JAVA_SERVER"
+
+    keep_container_alive_or_exit
+}
+
+keep_container_alive_or_exit() {
+    if [ "$KEEP_CONTAINER_ALIVE_AFTER_PDS_CRASHED" = "true" ]
+    then
+        echo "[ERROR] PDS crashed, but keeping the container alive."
```
de-jcup, line 69:
```suggestion
        echo "[ERROR] WebUI crashed, but keeping the container alive."
```

sechub-webui-solution/docker/run.sh
```diff
@@ -0,0 +1,110 @@
...
+    # Get process pid and wait until it ends
+    #   The pid will be needed by function trigger_shutdown() in case we receive a termination signal.
+    PID_JAVA_SERVER=$!
+    wait "$PID_JAVA_SERVER"
+
+    keep_container_alive_or_exit
+}
+
+keep_container_alive_or_exit() {
+    if [ "$KEEP_CONTAINER_ALIVE_AFTER_PDS_CRASHED" = "true" ]
```
de-jcup, line 67:
```suggestion
    if [ "$KEEP_CONTAINER_ALIVE_AFTER_WEBUI_CRASHED" = "true" ]
```

sechub-webui-solution/env-webui
```diff
@@ -0,0 +1,20 @@
...
+
+# User credentials for SecHub server
+SECHUB_USERID="admin"
+SECHUB_APITOKEN='myTop$ecret!'
+
+# Enable Java Debugging
+JAVA_ENABLE_DEBUG=false
+
+# Keep the container alive after it crashed
+KEEP_CONTAINER_ALIVE_AFTER_PDS_CRASHED=false
```
de-jcup, line 20:
```suggestion
KEEP_CONTAINER_ALIVE_AFTER_WEBUI_CRASHED=false
```

sechub-webui/build.gradle
```diff
@@ -12,6 +12,7 @@ plugins {
 
 dependencies {
     implementation project(':sechub-commons-core')
+	implementation project(':sechub-api-java')
```
de-jcup, line 15: indention?

sechub-webui/src/main/java/com/mercedesbenz/sechub/webui/IndexController.java
```diff
@@ -6,14 +6,16 @@
...
 @Controller
 public class IndexController {
     @Autowired
-    SecHubServerAccessService accessService;
+    SecHubAccessService accessService;
 
     @GetMapping("/")
     String index(Model model) {
-        model.addAttribute("sechubServerUrl", accessService.getSecHubServerUrl());
+        model.addAttribute("sechubServerUrl", accessService.getSecHubServerUri());
```
de-jcup, line 18: you changed the getter to "uri" - should the attribute sechubServerUrl" be renamed as well?

sechub-webui/src/main/java/com/mercedesbenz/sechub/webui/NewApiTokenController.java
```diff
@@ -6,22 +6,25 @@
...
-        // TODO: Replace with real user email once proper authentication is added
-        String email = "email@example.org";
+        // TODO Jeremias Eppler, 2023-09-27: Replace with real user email once proper
+        // authentication is added
+        String emailAddress = "email@example.org";
 
-        newApiTokenService.requestNewApiToken(email);
-        model.addAttribute("sechubServerUrl", accessService.getSecHubServerUrl());
-        model.addAttribute("userEmail", email);
+        model.addAttribute("sechubServerUrl", accessService.getSecHubServerUri());
```
de-jcup, line 25: Please introduce (a class with) constants for the model and use the constants everywhere. Means we would have central place and maintenance would be easier.

sechub-webui/src/main/java/com/mercedesbenz/sechub/webui/NewApiTokenService.java
```diff
@@ -26,10 +26,12 @@ public class NewApiTokenService {
...
-    public Mono<String> requestNewApiToken(String email) {
-
-        return webClient.post().uri("api/anonymous/refresh/apitoken/" + email).retrieve()
-                .onStatus(HttpStatus::is4xxClientError, error -> Mono.error(new RuntimeException("API not found")))
-                .onStatus(HttpStatus::is5xxServerError, error -> Mono.error(new RuntimeException("Server is not responding"))).bodyToMono(String.class);
+    public boolean userRequestsNewApiToken(String emailAddress) {
+        try {
+            accessService.getSecHubClient().userRequestsNewApiToken(emailAddress);
+            return true;
+        } catch (SecHubClientException e) {
```
de-jcup, line 33: add logging in error case

sechub-webui/src/main/java/com/mercedesbenz/sechub/webui/ScanController.java
```diff
@@ -1,19 +1,14 @@
...
 
 @Controller
 public class ScanController {
-    @Autowired
-    SecHubServerAccessService accessSerivce;
-
     @GetMapping("/projects/{projectId}/scans")
     String scans(Model model) {
-        model.addAttribute("sechubServerUrl", accessSerivce.getSecHubServerUrl());
         return "scans";
```
de-jcup, line 12: I am no sure if the remove was intendend.

settings.gradle
```diff
@@ -100,8 +100,10 @@ if (buildStage!=null) {
         'sechub-pds-tools', 
         
         // Examples (are below 'sechub-examples' folder)
-        'sechub-examples:example-sechub-api-java' 
+        'sechub-examples:example-sechub-api-java',
     
+		/* WebUI - needs the Java API */
```
de-jcup, line 105: indention?

de-jcup (changes requested):
Good job, but I found some points. Please read my comments.

sechub-webui-solution/docker/WebUI-Debian.dockerfile
Jeeppler: Add a todo comment.

de-jcup (approved):

Issue #2512, opened 2023-08-22
Developer Guide - Run Integration Tests from the Eclipse IDE


Jeeppler:
Explain in the developer guide on how to setup integration tests from the Eclipse IDE to either:

1. Run the integration tests
2. Debug the integration tests

PR #2539, opened 2023-09-09
Explain how to run the integrations tests #2512

Jeeppler:
- describe how to run integration tests from within the IDE
- add Eclipse launch configurations
- add screenshots

closes: #2512


Jeeppler:
@zigfridus have a look at this pull-request, it is the explanation regarding how to run an integration test.

de-jcup (changes requested):
Hello, good job.

But there is one little thing I found:
The link to the `.launch` will not work.

please take a look into 
https://github.com/mercedes-benz/sechub/blob/8409e12151707865f31290d5310f594940be1496/sechub-doc/build.gradle#L228-L233

The task `copyImagesToFinalDir` copies the generated images to the output directory to have it available at github pages, otherwise the images are not vailable!

I see two options here:
1) You create a new task `copyEclipseExamplesToFinalDir` and add as another dependency inside https://github.com/mercedes-benz/sechub/blob/8409e12151707865f31290d5310f594940be1496/sechub-doc/build.gradle#L235-L237

2) You directly link to the eclipse files via branch `develop`. Hier e an example link to an existing plantuml diagram:
`https://raw.githubusercontent.com/mercedes-benz/sechub/develop/sechub-doc/src/docs/asciidoc/diagrams/diagram_archive_support_overview.puml`

If you use 2) You should create a folder `eclipse` below 
https://github.com/mercedes-benz/sechub/tree/develop/sechub-examples
and move the eclipse examples there and use it for the links.

IMO I would prefer 2) because it is easier to implement and people which are searching for examples would find it also outside the documentation by checking the examples in code.

Jeeppler:
@de-jcup I chose the second option.

PR #2540, opened 2023-09-09
Change SecHubClient to interface with two implementations #2533

### LINKED ISSUE #2533: Provide mock possibility for `SecHubClient` in `sechub-api-java`
## Situation
The development on the WebUI is currently always blocked by the development of the SecHub Java Api client (see https://github.com/mercedes-benz/sechub/tree/develop/sechub-api-java)

If the WebUI developers wants to use a new feature (or there is bug inside the Java API client) the Web development is always blocked at the moment - which can be very frustrating.

## Wanted
It shall be possible to develop the WebUI even when the Java Api client has problems or a new but necessary feature on the client side is still in development.

It shall also be possible to test and develop the WebUI application without any running SecHub server or PDS.

### Java API changes
- We extract the methods of `SecHubClient` to an interface and the existing client implementation will be called `DefaultSecHubClient`.
- We separate parts which will be same in mock variant into an abstract base class `AbstractSecHubClient`
- We introduce `MockedSecHubClient` which uses internally only some predefined data and does not really communicate with the SecHub server. Logic shall work, but only in a reduced way that UI development is possible. The mocked client shall have methods to alter its behavior.

### Afterwards
Inside the WebUI a special spring boot profile can be introduced which uses either the default sechub client or a mocked variant - but this shall be done inside another issue.

de-jcup:
This PR closes #2533

### DIFF
--- a/sechub-api-java/src/main/java/com/mercedesbenz/sechub/api/AbstractSecHubClient.java
+++ b/sechub-api-java/src/main/java/com/mercedesbenz/sechub/api/AbstractSecHubClient.java
@@ -1,8 +1,8 @@
 package com.mercedesbenz.sechub.api;
 
 import java.net.URI;
-import java.util.LinkedList;
-import java.util.List;
+import java.util.LinkedHashSet;
+import java.util.Set;
 
 import javax.crypto.SealedObject;
 
@@ -16,10 +16,10 @@ public abstract class AbstractSecHubClient implements SecHubClient {
     private URI serverUri;
     private CryptoAccess<String> apiTokenAccess = new CryptoAccess<>();
 
-    private List<SecHubClientListener> secHubClientListeners;
+    private Set<SecHubClientListener> secHubClientListeners;
 
     public AbstractSecHubClient() {
-        secHubClientListeners = new LinkedList<>();
+        secHubClientListeners = new LinkedHashSet<>();
     }
 
     public void setUsername(String username) {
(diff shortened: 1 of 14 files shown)


sechub-api-java/src/main/java/com/mercedesbenz/sechub/api/User.java
```diff
@@ -0,0 +1,22 @@
+package com.mercedesbenz.sechub.api;
+
+public class User {
```
Jeeppler, line 3: You could use a record for this: https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Record.html

Jeeppler (changes requested):
Found minor issues. See comments.

sechub-api-java/src/main/java/com/mercedesbenz/sechub/api/AbstractSecHubClient.java
```diff
@@ -0,0 +1,99 @@
...
+    /**
+     * Adds an listener to the client. For some action on client side the listener
+     * will be informed. A listener can be added only one time no matter how many
+     * times this method is called.
+     *
+     * @param listener
+     */
+    @Override
+    public void addListener(SecHubClientListener listener) {
+        if (secHubClientListeners.contains(listener)) {
```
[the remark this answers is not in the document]
de-jcup: Good idea, I will use a linked hash set to have the ordering. Easier to debug, test and deterministic.

sechub-api-java/src/main/java/com/mercedesbenz/sechub/api/User.java
de-jcup: Currently it would work with records - but in future we could have additional methods at this class.
Also we would have no `getUserId()` but instead `userId()` - because records do handle this different. Also equals should be implemented in way, that a set cannot contain the same use r with the same user id multiple times. Records would treat same user with different email addresses as different objects.

I will keep the class approach here.

PR #2543, opened 2023-09-13
Improve include/exclude handling #2542

### LINKED ISSUE #2542: ZAP wrapper handles some combinations of webscan includes and excludes incorrectly
### Situation
Some combinations of target URLs and relative include/exclude paths result in errors, while building the URLs. The reason is, that the handling of slashes is not sufficient, since either the target URL could or could not end with a slash or the include/exclude could or could not start with a slash.

### Wanted
Improve the URL building of the ZAP wrapper, so any combination is handled correctly.

winzj:
- closes #2542


Jeeppler (approved):
Looks good 👍

PR #2551, opened 2023-09-18
Limit PDS db connections per default #2406

### LINKED ISSUE #2406: PDS: limit db connections per default
## Situation
Currently when a PDS is started, the Hikari database connection pool will create 10 database connections.
1-2 shared pool connections are sufficient for a PDS.
Additionally: with many PDS solutions, you can easily reach the max connection limit of the databasee.

So in the Helm chart of every PDS solution, we have set this:
```yaml
            # limit database connection pool
            - name: SPRING_DATASOURCE_HIKARI_MINIMUMIDLE
              value: "1"
            - name: SPRING_DATASOURCE_HIKARI_MAXIMUMPOOLSIZE
              value: "2"
```

## Wanted
Set the default in the PDS Spring profile like above.

de-jcup:
This PR
- closes #2406

### DIFF
--- a/sechub-pds/src/main/resources/application.yml
+++ b/sechub-pds/src/main/resources/application.yml
@@ -41,6 +41,9 @@ spring:
           jdbc.lob.non_contextual_creation: true
   datasource:
     initialization-mode: always
+    hikari:
+      minimumidle: 1
+      maximumpoolsize: 2
   main:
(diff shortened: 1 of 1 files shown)


sven-dmlr (approved):
Looks good :+1:

PR #2560, opened 2023-09-21
Changed behavior of java api client to avoid 401 problems #2553

### LINKED ISSUE #2553: Java SecHub client - Authentication 401 problems
## Situation
The current implementation of the Java SecHub client does use the JDK http client.

But when we call methods which needs authentication (and credentials are given to client) , it seems there are multiple connections but not all use the credientials! This leads internally to 401 authentication problems! If this happens multiple times the SecHub java client throws a client exception and stops processing.

## Wanted
The java client shall operate without any authentication problems.

de-jcup:
This PR
- closes #2553


Jeeppler (changes requested):
The code looks good. But, please fix the failing tests.

Jeeppler (approved):
Looks good 🥇 . Let's merge it.

PR #2572, opened 2023-09-28
Extend webscan include/exclude feature #2554

### LINKED ISSUE #2554: Extend includes/excludes for webscans with wildcards
### Situation
Currently we only allow includes/excludes starting with a slash.
```json
{
  "apiVersion": "1.0",
  "webScan": { 
    "url": "https://www.gamechanger.example.org", 
    "includes": [
      "/special/include"
    ], 
    "excludes": [
      "/en/contact",
      "/contact"
    ] 
  }
}
```
This includes/excludes will be interpreted like this: `https://www.gamechanger.example.org/special/include.*`

### Wanted
We want to change this behavior by allowing wildcards:
1.  `/en/contacts`, will match exactly the URL: `https://www.gamechanger.example.org/en/contacts` during the scan
2.  `/en/contacts/<*>`, will match the pattern: `https://www.gamechanger.example.org/en/contacts.*` during the scan
3. `<*>/en/contacts/<*>`, will match the pattern: `https://www.gamechanger.example.org/.*/en/contacts.*` during the scan
4. `en/contacts/<*>`, `en/contacts` or `en/contacts/`,  will be interpreted the same as `3.` during the scan

For includes it might be useful as well if someone only wants to scan some parts of the application that matches the pattern like: `https://www.gamechanger.example.org/special/include.*` the include `/special/include<*>` could be used.

winzj:
- closes #2554


sechub-scan/src/test/java/com/mercedesbenz/sechub/domain/scan/WebConfigBuilderStrategyTest.java
```diff
@@ -195,245 +191,10 @@ public void form_script_login_data_transfered() {
         assertEquals(null, action.getValue());
     }
 
-    @Test
-    public void excludes_no_slash_infront_single() {
-        /* prepare */
-        List<String> excludes = new LinkedList<>();
-        excludes.add("contact.html");
-
-        String json = createExcludesJson(excludes);
```
de-jcup, line 204: It is good that the validation is removed from the builder strategy. 
But the tests here did something which is not inside the validator test:
They use SECHUB_CONFIG.fromJson(json) -> the excludes/includes were tested with a json text.

`SecHubConfigurationModelTest` does do this also, but not with the new asterisks.
Please do me a favor and add one example with asterisk to `SecHubConfigurationModelTest`, so it is clear that the model can be created when this is inside (should always work, but it is better to have it inside the tests as well)

sechub-scan/src/test/java/com/mercedesbenz/sechub/domain/scan/WebConfigBuilderStrategyTest.java
de-jcup, line 1: Please additionally do following:

change `sechub-integrationtest-webscanconfig-all-options.json` to have the includes and excludes have one part with an asterisk as well.

This is tested by integration test `PDSWebScanJobScenario12IntTest`. With the change mentioned before we ensure that the asterisks are also at PDS side at runtime.

de-jcup (changes requested):
Good job, but I found some smaller parts. Please take a look at my comments.

PR #2592, opened 2023-10-11
Enable auto cleanup for old scan reports #2075

### LINKED ISSUE #2075: Auto cleanup shall drop ScanResults + migration SQL
## Situation
With #1874 we do not longer need scan results for statistics and we can remove the workaround inside `ScanAutoCleanupService`.

IMPORTANT: We need a special migration for old scan results, because the auto cleanup will not drop the old ones, because their jobs are already auto cleaned and the information is missing...

## Wanted
- remove the workaround
- provide a SQL script which drops all scan results where job uuid is not inside (this may not be inside flyway because this works only when domains are inside same database, which could change in future). We provide this inside OPS documentation athttps://mercedes-benz.github.io/sechub/latest/sechub-operations.html#migration-steps
- update dokumentation in https://mercedes-benz.github.io/sechub/latest/sechub-operations.html#migration-steps

### LINKED ISSUE #2076: Change logic in false positive service - only load report when necessary
## Situation
With #2075 we will finally remove the old status results, so auto cleanup is also purging old scan results.

This will have impacts to false positive handling:
Some user do always send the "markFalsePositive" list again and again (they provide it by a file inside their repo).
This includes references to old  jobs which could be removed by auto cleanup after #2075

At this point the mark false positive call would fail!

## Wanted
False positive markup shall not fail when data is already known - so users can still do this. Only new markups shall fail, because the information is not available.

## Solution
We change the implementation, so mark and unmark of false positives does not load the report always, but only when it is necessary.

Means: If the false positive definition for "Job A" and "finding X" is already defined, we just ignore it and do not load the report data (because unnecessary).

de-jcup:
This PR
- closes #2075 
- closes #2076


sven-dmlr (approved):
Looks good to me :+1: 

Looking forward to get the old data deleted :-)

PR #2595, opened 2023-10-13
Feature 2504 pds solutions xray integration

### LINKED ISSUE #2504: PDS solutions:  integrate Xray
# Situation
Xray security scanner from JFrog is currently not integrated in SecHub

# Wanted
Add Xray PDS solution to sechub to enable container SCA scans with Xray. Receive SecHub report containing an SBOM with licenses and related vulnerabilities.

# ToDo
- PDS solution for Xray
- API wrapper to communicate with JFrog artifactory
- Process Xray reports

lorriborri:
Pull request contains:
- Xray wrapper for Jfrog artifactory
-  PDS solutions for Xray

closes #2504


Jeeppler (changes requested):
Overall, the structure look good. However, there are many smaller issues. See comments.

Furthermore, a major issue from my perspective is assuming that the secure hash value will be  always in sha256. Please anticipate other algorithms (SHA2, SHA3 etc.) for the future. This is called crypto agility.

sechub-wrapper-xray/src/main/java/com/mercedesbenz/sechub/wrapper/xray/cli/XrayWrapperArtifactoryClientController.java
```diff
@@ -0,0 +1,119 @@
...
+import com.mercedesbenz.sechub.wrapper.xray.config.XrayWrapperConfiguration;
+import com.mercedesbenz.sechub.wrapper.xray.report.XrayWrapperReportException;
+import com.mercedesbenz.sechub.wrapper.xray.report.XrayWrapperReportReader;
+
+public class XrayWrapperArtifactoryClientController {
+
+    private static final Logger LOG = LoggerFactory.getLogger(XrayWrapperArtifactoryClientController.class);
+    private final XrayWrapperConfiguration xrayWrapperConfiguration;
+    private final XrayAPIArtifactoryClient artifactoryClient;
+    private int retries;
```
de-jcup, line 21: retries is stateful here. As long as `waitForScansToFinishAndDownloadReport` is called only one time it is not problematic. But the wait method is public and if somebody would reuse the controller instance, the second call would no work any more (I expect that a new wait would start from 0.

de-jcup (changes requested):
Hello, first of all: a good job.
But there are parts which are not clear for myself / I want to have changes.

I stop now with the review, because it takes a lot of time to write. I make an appointment and we go to the code parts I have found and what is still unclear from my side. When we discuss it it is more easy for both of us.

See you

Jeeppler (changes requested):
See comments.

sechub-wrapper-xray/src/main/java/com/mercedesbenz/sechub/wrapper/xray/util/InputStreamSaver.java
```diff
@@ -2,12 +2,12 @@
...
 import java.io.*;
 
+import com.mercedesbenz.sechub.wrapper.xray.XrayWrapperException;
 import com.mercedesbenz.sechub.wrapper.xray.cli.XrayWrapperExitCode;
-import com.mercedesbenz.sechub.wrapper.xray.cli.XrayWrapperRuntimeException;
 
 public class InputStreamSaver {
 
-    public static void saveInputStreamToZipFile(String filename, InputStream inputStream) throws XrayWrapperRuntimeException {
+    public static void saveInputStreamToZipFile(String filename, InputStream inputStream) throws XrayWrapperException {
```
de-jcup, line 10: Why are all these methods static?
Why not just use an instance (maybe shared/default one).

For testing purposes static methods are most time problematic.

de-jcup, line 10:
```suggestion
    public static void zip(File zipFilename, InputStream inputStream) throws XrayWrapperException {
```

sechub-wrapper-xray/src/main/java/com/mercedesbenz/sechub/wrapper/xray/util/InputStreamSaver.java
```diff
@@ -19,11 +19,11 @@ public static void saveInputStreamToZipFile(String filename, InputStream inputSt
             }
             inputStream.close();
         } catch (IOException e) {
-            throw new XrayWrapperRuntimeException("Could not save https input stream to zip file", e, XrayWrapperExitCode.IO_ERROR);
+            throw new XrayWrapperException("Could not save https input stream to zip file", XrayWrapperExitCode.IO_ERROR, e);
         }
     }
 
-    public static String readInputStreamAsString(InputStream inputStream) throws XrayWrapperRuntimeException {
+    public static String readInputStreamAsString(InputStream inputStream) throws XrayWrapperException {
```
de-jcup, line 26: The class is called `InputStreamSaver`. 
But it does provide a zip method and a readInputAsStream

Maybe you should rename it to IoUtil (if methods are kept static) or to `IOHelper` or `IOSupport` if you want to use instances.

sechub-wrapper-xray/src/main/java/com/mercedesbenz/sechub/wrapper/xray/util/InputStreamSaver.java
```diff
@@ -19,11 +19,11 @@ public static void saveInputStreamToZipFile(String filename, InputStream inputSt
...
         BufferedReader in = new BufferedReader(new InputStreamReader(inputStream));
```
de-jcup, line 27: Please use auto close instead:
```suggestion
        try( BufferedReader in = new BufferedReader(new InputStreamReader(inputStream));
```

sechub-wrapper-xray/src/main/java/com/mercedesbenz/sechub/wrapper/xray/util/XrayAPIAuthenticationHeader.java
```diff
@@ -13,13 +13,13 @@ public class XrayAPIAuthenticationHeader {
...
-    public static String buildAuthHeader() {
+    public static String buildBasicAuthHeader() throws XrayWrapperException {
         EnvironmentVariableReader environmentVariableReader = new EnvironmentVariableReader();
         String username = environmentVariableReader.readEnvAsString(EnvironmentVariableConstants.XRAY_USERNAME_ENV);
-        String pwd = environmentVariableReader.readEnvAsString(EnvironmentVariableConstants.XRAY_PASSWORD_ENV);
-        String auth = (username + ":" + pwd);
-        if (username == null || pwd == null) {
-            throw new XrayWrapperRuntimeException("Authentication not possible because of missing environment variables XRAY_USER and XRAY_PASSWORD",
+        String password = environmentVariableReader.readEnvAsString(EnvironmentVariableConstants.XRAY_PASSWORD_ENV);
+        String auth = (username + ":" + password);
```
de-jcup, line 20:
```suggestion
        String auth = username + ":" + password;
```

sechub-wrapper-xray/src/main/java/com/mercedesbenz/sechub/wrapper/xray/util/XrayAPIAuthenticationHeader.java
```diff
@@ -13,13 +13,13 @@ public class XrayAPIAuthenticationHeader {
...
+        if (username == null || password == null) {
```
de-jcup, line 21: Please move this up, so it is done before auth string is created.

sechub-wrapper-xray/src/main/java/com/mercedesbenz/sechub/wrapper/xray/util/XrayWrapperConfigurationHelper.java
```diff
@@ -19,10 +19,10 @@ public static XrayWrapperConfiguration createXrayConfiguration(XrayWrapperScanTy
...
 
         // get repository according to scan type
-        String repository = "";
+        String registry = "";
         if (scanType.equals(XrayWrapperScanTypes.DOCKER)) {
-            repository = environmentVariableReader.readEnvAsString(EnvironmentVariableConstants.DOCKER_REGISTRY_ENV);
+            registry = environmentVariableReader.readEnvAsString(EnvironmentVariableConstants.DOCKER_REGISTRY_ENV);
         }
-        return XrayWrapperConfiguration.Builder.create(artifactoryUrl, repository, zipDirectory, outputFile).build();
+        return XrayWrapperConfiguration.Builder.builder(artifactoryUrl, registry, zipDirectory, outputFile).build();
```
de-jcup, line 26: is it correct that we accept registry as an empty string here?
shoud we throw instead an `IllegalStateException` if the scan type is not supported (something else than docker) ?

Or at least a log warning here?

sechub-wrapper-xray/src/main/java/com/mercedesbenz/sechub/wrapper/xray/util/ZipFileExtractor.java
```diff
@@ -9,7 +9,6 @@
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipInputStream;
 
-import com.mercedesbenz.sechub.wrapper.xray.cli.XrayWrapperExitCode;
 import com.mercedesbenz.sechub.wrapper.xray.report.XrayWrapperReportException;
```
de-jcup, line 12: We have a `ZipFileExtract` class (which makes absolute sense) but the zip mechanism is inside a complete other class which handles multiple IO things.

Maybe we should create a `ZipFileCreater` or `ZipFileArchiver` class for the zip mechanism?

sechub-wrapper-xray/src/test/java/com/mercedesbenz/sechub/wrapper/xray/api/XrayAPIHTTPUrlConnectionFactoryTest.java
```diff
@@ -13,25 +13,29 @@
...
+
     @BeforeEach
     void beforeEach() {
         url = mock(URL.class);
+        factory = new XrayAPIHTTPUrlConnectionFactory();
     }
 
     @Test
-    void setUpHTTPConnection_GET_request() throws IOException {
+    void setUpHTTPConnection_GET_request() throws IOException, XrayWrapperException {
```
de-jcup, line 31: Please rename test method names 
from setup... to something like create...

sechub-wrapper-xray/src/test/java/com/mercedesbenz/sechub/wrapper/xray/api/XrayAPIResponseFactoryTest.java
```diff
@@ -14,21 +14,24 @@
...
+
     @BeforeEach
     void beforeEach() {
         connection = mock(HttpURLConnection.class);
+        xrayAPIResponseFactory = new XrayAPIResponseFactory();
     }
 
     @Test
-    void getHttpResponseFromConnection_get_valid_http_response() throws IOException {
+    void getHttpResponseFromConnection_get_valid_http_response() throws IOException, XrayWrapperException {
```
de-jcup, line 34: Please rename test methods like "getHttpResponse..." to something like "createHttpResponse..." or use a name which does not contain the caller name (but currently the get is irritating if somebody does not know the old method names...)

sechub-wrapper-xray/src/test/java/com/mercedesbenz/sechub/wrapper/xray/api/XrayWrapperArtifactoryClientTest.java
```diff
@@ -34,22 +31,19 @@ class XrayWrapperArtifactoryClientTest {
     void beforeEach() {
         XrayWrapperArtifact artifact = new XrayWrapperArtifact("name", "sha256", "tag", XrayWrapperScanTypes.DOCKER);
         XrayWrapperConfiguration configuration = XrayWrapperConfiguration.Builder
-                .create("http://notmalformed-url-example.com", "example", "zipDirectory", "report").build();
+                .builder("http://notmalformed-url-example.com", "example", "zipDirectory", "report").build();
```
de-jcup, line 34: mentioned before, but maybe on something else:

```suggestion
                .builder().
                     url("http://notmalformed-url-example.com").
                     type("example").
                     zipDirectory("zipDirectory").
                     reportDirectory("report").
                  build();
```

Remark: i just added some method names, do not just use them, the introduced names are just to show up the idea.

A builder should have dedicated methods for the parameters.
A builder is intended to avoid "hell of construction parameters" - so multiple parameters inside builder itself makes not much sense IMO.

sechub-wrapper-xray/src/test/java/com/mercedesbenz/sechub/wrapper/xray/api/XrayWrapperArtifactoryClientTest.java
```diff
@@ -34,22 +31,19 @@ class XrayWrapperArtifactoryClientTest {
...
         xrayAPIArtifactoryClient = new XrayAPIArtifactoryClient(artifact, configuration);
```
de-jcup, line 35:
```suggestion
        clientToTest = new XrayAPIArtifactoryClient(artifact, configuration);
```

sechub-wrapper-xray/src/test/java/com/mercedesbenz/sechub/wrapper/xray/api/XrayWrapperArtifactoryClientTest.java
```diff
@@ -61,40 +55,37 @@ void getXrayVersion_return_version() throws JsonProcessingException {
...
 
     @Test
-    void getScanStatus_return_status_scanned() throws JsonProcessingException {
+    void getScanStatus_return_status_scanned() throws XrayWrapperException {
         /* prepare */
-        ObjectMapper mapper = new ObjectMapper();
         String jsonString = "{\"status\": \"scanned\"}";
-        response = XrayAPIResponse.Builder.create(200, headers).setBody(jsonString).build();
-        JsonNode node = mapper.readTree(jsonString);
+        response = XrayAPIResponse.Builder.builder(200, headers).addResponseBody(jsonString).build();
```
de-jcup, line 84: Response implementation shall not be tested inside this class, but the artifactory client. Using the builder here tests also foreign code... couldn't we do this via mockito as done in other test methods or is this just to much effort?

sechub-wrapper-xray/src/test/java/com/mercedesbenz/sechub/wrapper/xray/api/XrayWrapperArtifactoryClientTest.java
```diff
@@ -61,40 +55,37 @@ void getXrayVersion_return_version() throws JsonProcessingException {
...
         Mockito.doReturn(response).when(xrayAPIArtifactoryClientSpy).send(any());
-        Mockito.doReturn(node).when(xrayAPIArtifactoryClientSpy).getBodyAsNode(any());
 
         /* execute */
-        String status = xrayAPIArtifactoryClientSpy.getScanStatus();
+        String status = xrayAPIArtifactoryClientSpy.getScanStatus().getStatus();
```
de-jcup, line 88: It took a while to understand the test mechanism here.

But ...
You are using a spy object to change the implemenation behavior of the origin part you want to test!

Instead of using
```suggestion
        String status = clientToTest.getScanStatus().getStatus();
```
You are using a spy which overrides the send implementation and return always a dedicated response.

Means: All parts from the send implemenation are overriden and never tested!

IMO the spy approach is not good here. We should discuss a better solution.

sechub-wrapper-xray/src/test/java/com/mercedesbenz/sechub/wrapper/xray/cli/XrayWrapperArtifactoryClientSupportTest.java
```diff
@@ -9,13 +9,14 @@
...
+import com.mercedesbenz.sechub.wrapper.xray.XrayWrapperException;
 import com.mercedesbenz.sechub.wrapper.xray.api.XrayAPIArtifactoryClient;
 import com.mercedesbenz.sechub.wrapper.xray.config.XrayWrapperArtifact;
 import com.mercedesbenz.sechub.wrapper.xray.config.XrayWrapperConfiguration;
 
-class XrayWrapperArtifactoryClientControllerTest {
+class XrayWrapperArtifactoryClientSupportTest {
 
-    XrayWrapperArtifactoryClientController controller;
+    XrayWrapperArtifactoryClientSupport controller;
```
de-jcup, line 19:
```suggestion
    XrayWrapperArtifactoryClientSupport support;
```

sechub-wrapper-xray/src/test/java/com/mercedesbenz/sechub/wrapper/xray/report/CycloneDXVulnerabilityHelperTest.java
```diff
@@ -20,7 +21,7 @@ class CycloneDXVulnerabilityHelperTest {
 
     @BeforeEach
     void beforeEach() {
-        mapper = new ObjectMapper();
+        mapper = XrayWrapperJSONConverter.get().getMapper();
         vulnerability = new CycloneDXVulnerabilityHelper();
```
de-jcup, line 25:
```suggestion
        helperToTest = new CycloneDXVulnerabilityHelper();
```

sechub-wrapper-xray/src/test/java/com/mercedesbenz/sechub/wrapper/xray/report/XrayWrapperReportParserTest.java
```diff
@@ -32,11 +33,11 @@ void getRootDataNode_throws_xrayWrapperReportException() {
         /* prepare */
         File file = new File("src/test/resources/invalid-json-examples/invalid_json.json");
         /* execute + test */
-        Assert.assertThrows(XrayWrapperReportException.class, () -> xrayWrapperReportParser.getRootDataNode(file));
+        Assert.assertThrows(XrayWrapperException.class, () -> xrayWrapperReportParser.getRootDataNode(file));
```
de-jcup, line 36: please do an static import for "Assert.*" and use 
```suggestion
        assertThrows(XrayWrapperException.class, () -> xrayWrapperReportParser.getRootDataNode(file));
```

sechub-wrapper-xray/src/test/java/com/mercedesbenz/sechub/wrapper/xray/report/XrayWrapperReportReaderTest.java
```diff
@@ -24,24 +26,24 @@ class XrayWrapperReportReaderTest {
...
         /* execute */
-        reportReader.getReportFiles(source, resultFile);
+        reportReader.findXrayReportsInArchive(source, resultFile);
 
         /* test */
         String cycloneName = reportReader.cycloneReport.getName();
         String securityName = reportReader.securityReport.getName();
-        String sechubReport = reportReader.sechubReport.getName();
+        String sechubReport = reportReader.xrayPdsReport.getName();
         assertEquals(resultFile, sechubReport);
```
de-jcup, line 47: Normally I am a fan of local variables, but here I think that 
```suggestion
        assertEquals("resultfile", sechubReport);
        assertEquals("Docker_SBOM_Export_CycloneDX.json", cycloneName);
        assertEquals("Docker_Security_Export.json", securityName);
```

would be easier to read/maintain.

sechub-wrapper-xray/src/test/java/com/mercedesbenz/sechub/wrapper/xray/report/XrayWrapperReportReaderTest.java
```diff
@@ -24,24 +26,24 @@ class XrayWrapperReportReaderTest {
     @BeforeEach
     void beforeEach() {
         reportReader = new XrayWrapperReportReader();
-        mapper = new ObjectMapper();
+        mapper = XrayWrapperJSONConverter.get().getMapper();
     }
 
     @Test
-    void getReportFiles_get_valid_reports() {
+    void getReportFiles_get_valid_reports() throws XrayWrapperReportException {
```
de-jcup, line 33: Please rename test method

sechub-wrapper-xray/src/test/java/com/mercedesbenz/sechub/wrapper/xray/report/XrayWrapperReportReaderTest.java
```diff
@@ -24,24 +26,24 @@ class XrayWrapperReportReaderTest {
...
         /* prepare */
         String source = "src/test/resources/xray-report-examples";
         String resultFile = "resultfile";
         String cycloneReport = "Docker_SBOM_Export_CycloneDX.json";
         String securityReport = "Docker_Security_Export.json";
 
         /* execute */
-        reportReader.getReportFiles(source, resultFile);
+        reportReader.findXrayReportsInArchive(source, resultFile);
```
de-jcup, line 41: this is a find method... normally I would expect something like this.

``` java
finder = new MyFinder();

MyFinderResult result = finder. findXyz(params)

assertEquals("expected", result.getCycloneName())
// etc. etc.
```

Your current implementation is stateful and would never be thread safe.
It is also difficult to understand why a find... does not return the find result. Means: "find... what?"

sechub-wrapper-xray/src/test/java/com/mercedesbenz/sechub/wrapper/xray/report/XrayWrapperReportVulnerabilityMapperTest.java
```diff
@@ -49,7 +49,7 @@ void mapVulnerabilities_throws_illegalArgumentException() {
     }
 
     @Test
-    void mapVulnerabilities_throws_() {
+    void mapVulnerabilities_throws_xrayWrapperReportException() {
         /* prepare */
         HashMap<String, CycloneDXVulnerabilityHelper> vulnerabilityHashMap = new HashMap<>();
```
de-jcup, line 54:
```suggestion
        Map<String, CycloneDXVulnerabilityHelper> vulnerabilityHashMap = new HashMap<>();
```
It is always a good idea to use interfaces on the left side and hide the details - it reduces time for refactoring, changes and also hides implementation details.

de-jcup (changes requested):
I still found some parts.

CLA-Mercedes-Benz:
[CLA assistant check](https://cla.opensource.mercedes-benz.com/mercedes-benz/sechub?pullRequest=2595) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.mercedes-benz.com/mercedes-benz/sechub?pullRequest=2595) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.mercedes-benz.com/check/mercedes-benz/sechub?pullRequest=2595) it.</sub>

sechub-wrapper-xray/src/main/java/com/mercedesbenz/sechub/wrapper/xray/api/XrayAPIResponseFactory.java
```diff
@@ -0,0 +1,87 @@
...
+            throw new XrayWrapperException("Could not read Response Message from HTTP connection.", XrayWrapperExitCode.IO_ERROR, e);
+        }
+
+        // read headers
+        Map<String, List<String>> header = httpURLConnection.getHeaderFields();
+
+        // get input stream from response, client accept status codes 3xx
+        if (responseCode > 399) {
+            try (InputStream inputStream = httpURLConnection.getErrorStream()) {
+                if (inputStream != null) {
```
de-jcup, line 51: Hm... when the input (in this case error stream) is null, we have no conent and the last call will be done line 70. Was this intended? 

What about simply doing this:
```java
String content = "No error stream information available.";
if (inputStream!=null){
    content = saveOrReadHTTPContent(httpURLConnection, zipArchive, inputStream);
}
return XrayAPIResponse.Builder.builder().statusCode(responseCode).headers(header).addResponseBody(content)
                            .addResponseMessage(responseMessage).build();
```
and next lines similar.

After reading the next lines, maybe even doing the complete operation would recude the build creation to only one builder call if "content" would be used at top and for all of them -> only one 
```java
 return XrayAPIResponse.Builder.builder().statusCode(responseCode).headers(header).addResponseMessage(responseMessage).build();
```

sechub-wrapper-xray/src/main/java/com/mercedesbenz/sechub/wrapper/xray/api/XrayAPIResponseFactory.java
```diff
@@ -0,0 +1,87 @@
...
+                        .build();
+            } catch (IOException e) {
+                throw new XrayWrapperException("Could not save https input stream to zip file", XrayWrapperExitCode.IO_ERROR, e);
+            }
+        }
+
+        return XrayAPIResponse.Builder.builder().statusCode(responseCode).headers(header).addResponseMessage(responseMessage).build();
+    }
+
+    private String saveOrReadHTTPContent(HttpURLConnection httpURLConnection, String zipArchive, InputStream inputStream) throws XrayWrapperException {
```
de-jcup, line 73: I am sorry that I did not see this before, but here we have method with logic inside where the caller has no clue if there shall be string returned or not.

Would it not make more sense to have two methods: "saveHTTPContent" and "readHTTPContentAsString"? 
At line 81 always an empty string is returned - which makes no sense for callers to handle it in any way. I think IMHO that it should be clear at caller side if there is need of storage or a need to fetch a string

sechub-wrapper-xray/src/main/java/com/mercedesbenz/sechub/wrapper/xray/cli/XrayWrapperArtifactoryClientSupport.java
```diff
@@ -0,0 +1,237 @@
...
+            return false;
+        }
+        case SCANNED -> {
+            return true;
+        }
+        }
+        throw new XrayWrapperException("Received unexpected scan status", XrayWrapperExitCode.UNSUPPORTED_API_REQUEST);
+    }
+
+    private static void waitSeconds(int val) throws XrayWrapperException {
```
de-jcup, line 148: I think we can remove the static part here - its is only private and used internally, so access should be always possible.
```suggestion
    private void waitSeconds(int val) throws XrayWrapperException {
```

sechub-wrapper-xray/src/main/java/com/mercedesbenz/sechub/wrapper/xray/cli/XrayWrapperArtifactoryClientSupport.java
```diff
@@ -0,0 +1,237 @@
...
+        try {
+            TimeUnit.SECONDS.sleep(val);
+        } catch (InterruptedException e) {
+            throw new XrayWrapperException("Thread interrupted while wait for next scan status request", XrayWrapperExitCode.THREAD_INTERRUPTION, e);
```
de-jcup, line 152: Please read https://www.baeldung.com/java-interrupted-exception#2-restore-the-interrupt.

Maybe we should think about restoring the interruption state to thread:
```suggestion
        } catch (InterruptedException e) {
            Thread.currentThread().interrupt();  //set the flag back to true
            throw new XrayWrapperException("Thread interrupted while wait for next scan status request", XrayWrapperExitCode.THREAD_INTERRUPTION, e);
```

sechub-wrapper-xray/src/main/java/com/mercedesbenz/sechub/wrapper/xray/report/XrayWrapperReportSupport.java
```diff
@@ -0,0 +1,86 @@
...
+            // folder with reports is zipped
+            String zipArchive = unzippedArchive + ".zip";
+            if (zipFileExtractor.fileExists(zipArchive)) {
+                zipFileExtractor.unzipFile(Paths.get(zipArchive), Paths.get(unzippedArchive));
+            } else {
+                throw new XrayWrapperReportException("File with reports does not exist");
+            }
+        }
+
+        ArrayList<Path> cyclones = getFilesByName(unzippedArchive, "CycloneDX");
```
de-jcup, line 40: I do not think we need the information that this is an ArrayList (and also not in next line)
```suggestion
        List<Path> cyclones = getFilesByName(unzippedArchive, "CycloneDX");
```

sechub-wrapper-xray/src/main/java/com/mercedesbenz/sechub/wrapper/xray/report/XrayWrapperReportSupport.java
```diff
@@ -0,0 +1,86 @@
...
+        return vulnerabilityMapper.mapVulnerabilities(cycloneReport, cycloneDXVulnerabilityMap);
+    }
+
+    public void writeReport(Bom sbom, File xrayPdsReport) throws XrayWrapperReportException {
+        XrayWrapperReportWriter reportWriter = new XrayWrapperReportWriter();
+        reportWriter.writeReport(sbom, xrayPdsReport);
+    }
+
+    private ArrayList<Path> getFilesByName(String dir, String filename) throws XrayWrapperReportException {
+        ArrayList<Path> paths = new ArrayList<>();
```
de-jcup, line 74:
```suggestion
        List<Path> paths = new ArrayList<>();
```

sechub-wrapper-xray/src/main/java/com/mercedesbenz/sechub/wrapper/xray/report/XrayWrapperReportTransformer.java
```diff
@@ -0,0 +1,226 @@
...
+
+        rating.setMethod(Vulnerability.Rating.Method.fromString(ratingRecord.method()));
+
+        rating.setVector(ratingRecord.vector());
+
+        vulnerability.addRating(rating);
+    }
+
+    private void addDescriptionToVulnerability(JsonNode node, Vulnerability vulnerability) {
+        String description = node.get("component_versions").get("more_details").get("description").asText();
```
de-jcup, line 164: Strings like "component_versions" etc. should IMO be constants and the constants could be inside `XrayWrapperReportConstants`. Reason - some parts are used multiple times and this would reduce typo possiblity.

But its only a suggestion and not really necessary.

sechub-wrapper-xray/src/main/java/com/mercedesbenz/sechub/wrapper/xray/util/ZipFileCreator.java
```diff
@@ -0,0 +1,23 @@
...
+
+import com.mercedesbenz.sechub.wrapper.xray.XrayWrapperException;
+import com.mercedesbenz.sechub.wrapper.xray.cli.XrayWrapperExitCode;
+
+public class ZipFileCreator {
+    public void zip(File file, InputStream inputStream) throws XrayWrapperException {
+        int read;
+        byte[] buffer = new byte[1024];
+        FileOutputStream fileOutputStream = null;
+        try {
```
de-jcup, line 13: Please use auto close:
```suggestion
        try (FileOutputStream fileOutputStream = new FileOutputStream(file){
        
```
You do not need to handle inputstream close again and it does also handle closing of the stream in case of an io exception

sechub-wrapper-xray/src/main/java/com/mercedesbenz/sechub/wrapper/xray/util/ZipFileExtractor.java
```diff
@@ -0,0 +1,78 @@
...
+        } catch (IOException e) {
+            throw new XrayWrapperReportException("Could not extract zip file.", XrayWrapperExitCode.IO_ERROR, e);
+        }
+    }
+
+    /*
+     * Please see: <a href=https://security.snyk.io/research/zip-slip-vulnerability
+     * /a>
+     */
+    private static Path zipSlipProtect(ZipEntry zipEntry, Path targetDir) throws XrayWrapperReportException {
```
de-jcup, line 66:
```suggestion
    private Path zipSlipProtect(ZipEntry zipEntry, Path targetDir) throws XrayWrapperReportException {
```

sechub-wrapper-xray/src/test/java/com/mercedesbenz/sechub/wrapper/xray/cli/XrayWrapperArtifactoryClientSupportTest.java
```diff
@@ -0,0 +1,53 @@
...
+            when(mock.getScanStatus()).thenReturn(XrayWrapperArtifactoryClientSupport.ScanStatus.SCANNED);
+            when(mock.requestScanReports()).thenReturn(false);
+        })) {
+            clientSupportToTest = new XrayWrapperArtifactoryClientSupport(configuration, artifact);
+            clientSupportToTest.waitForScansToFinishAndDownloadReport();
+        }
+    }
+
+    @Test
+    void waitForScansToFinishAndDownloadReport_throws_XrayWrapperException() {
```
de-jcup, line 46: I do not exactly get the point why the exception is thrown.

looking at the test before there are 4 mock calls/definitions to get it working - what is here missing to bring up the exception? Or do we have 4 situation were the exception could happen? In this case we should have 4 test methods where only one of the parameters is missing to test the combinations.

sechub-wrapper-xray/src/test/java/com/mercedesbenz/sechub/wrapper/xray/util/ZipFileExtractorTest.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    @Test
+    void unzipFile_throws_rayWrapperReportException_with_invalid_file() {
+        /* prepare */
+        File file = new File("file");
+
+        /* execute + test */
+        assertThrows(XrayWrapperReportException.class, () -> zipFileExtractorToTest.unzipFile(file.toPath(), file.toPath()));
+    }
```
de-jcup, line 34: What about the "positive way" of the unzip method call? there is no check that the unzip mechanism works?

de-jcup (changes requested):
Very good job. But I still found some parts. Please read my comments.

sechub-wrapper-xray/src/test/java/com/mercedesbenz/sechub/wrapper/xray/api/XrayAPIRequestTest.java
```diff
@@ -0,0 +1,37 @@
...
+
+        /* test */
+        assertEquals(url, request.getUrl());
+        assertEquals(auth, request.isAuthenticationNeeded());
+        assertEquals(data, request.getData());
+        assertEquals(XrayAPIRequest.RequestMethodEnum.GET, request.getRequestMethodEnum());
+    }
+
+    @Test
+    void xrayRequest_throws_rayRequest_builder_throws_xrayWrapperException_when_parameters_not_defined() {
```
Jeeppler, line 33:
```suggestion
    void xrayRequest_throws_xrayRequest_builder_throws_xrayWrapperException_when_parameters_not_defined() {
```

sechub-wrapper-xray/src/test/java/com/mercedesbenz/sechub/wrapper/xray/api/XrayWrapperArtifactoryClientTest.java
```diff
@@ -0,0 +1,201 @@
...
+        String jsonString = "{\"xray_version\": \"1";
+        response = XrayAPIResponse.Builder.builder().statusCode(200).headers(headers).addResponseBody(jsonString).addResponseMessage("Message").build();
+        when(clientToTest.xrayAPIResponseFactory.createHttpResponseFromConnection(any(), any())).thenReturn(response);
+
+        /* execute + test */
+        assertThrows(XrayWrapperException.class, () -> clientToTest.requestXrayVersion());
+    }
+
+    @Test
+    void checkArtifactoryUpload_return_true() throws XrayWrapperException {
```
Jeeppler, line 73:
```suggestion
    void checkArtifactoryUpload_returns_true() throws XrayWrapperException {
```

sechub-wrapper-xray/src/test/java/com/mercedesbenz/sechub/wrapper/xray/cli/XrayWrapperCLITest.java
```diff
@@ -0,0 +1,43 @@
...
+
+    XrayWrapperCLI cliToTest;
+
+    @BeforeEach
+    void beforeEach() {
+        cliToTest = new XrayWrapperCLI();
+    }
+
+    @Test
+    void start_with_null_args_throws_nullPointerException() {
```
Jeeppler, line 24:
```suggestion
    void start_with_null_throws_nullPointerException() {
```

sechub-wrapper-xray/src/test/java/com/mercedesbenz/sechub/wrapper/xray/cli/XrayWrapperCommandLineParserTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @BeforeEach
+    void beforeEach() {
+        parserToTest = new XrayWrapperCommandLineParser();
+    }
+
+    @Test
+    void parseCommandLineArgs_valid_arguments() throws XrayWrapperCommandLineParserException, XrayWrapperException {
+        /* prepare */
+        String[] args = { "--name", "myname", "--checksum", "sha256:123", "--scantype", "docker", "--outputfile", "outfile" };
```
Jeeppler, line 23: `sha256:123` -> should throw an exception, because `123` is not a sha256 checksum (too short).

sechub-wrapper-xray/src/test/java/com/mercedesbenz/sechub/wrapper/xray/report/XrayWrapperReportTransformerTest.java
```diff
@@ -0,0 +1,45 @@
...
+    void beforeEach() {
+        xrayWrapperReportTransformerToTest = new XrayWrapperReportTransformer();
+    }
+
+    @Test
+    void transformSecurityReport_number_of_vulnerabilities() throws XrayWrapperException {
+        /* prepare */
+        File file = new File("src/test/resources/xray-report-examples/Docker_Security_Export.json");
+        JsonNode node = XrayWrapperJSONConverter.get().readJSONFromFile(file).get("data");
+        int numberOfVulnerabilities = 25;
```
Jeeppler, line 31:
```suggestion
        int expectedNumberOfVulnerabilities = 25;
```

sechub-wrapper-xray/src/test/java/com/mercedesbenz/sechub/wrapper/xray/report/XrayWrapperReportVulnerabilityMapperTest.java
```diff
@@ -0,0 +1,69 @@
...
+        Vulnerability vulnerability02 = new Vulnerability();
+        vulnerability02.setId("CVE-2019-17594");
+        vulnerability02.setBomRef("testBomRef02");
+        vulnerabilityMap.put("CVE-2019-17594", vulnerability02);
+
+        /* execute */
+        Bom bom = vulnerabilityMapperToTest.mapVulnerabilities(file, vulnerabilityMap);
+
+        /* test */
+        List<Vulnerability> vulnerabilities = bom.getVulnerabilities();
```
Jeeppler, line 44: please add `assertEquals(2, vulnerability.size)`

Jeeppler (changes requested):
It is getting closer to the finish line. However, I found a number of issues. Please have a look at the comments.

sechub-wrapper-xray/src/main/java/com/mercedesbenz/sechub/wrapper/xray/cli/XrayWrapperArtifactoryClientSupport.java
```diff
@@ -0,0 +1,255 @@
...
+        context.setStatus(artifactoryClient.getScanStatus());
+        LOG.debug("Artifact status is: " + context.getStatus().getStatusValue());
+        if (context.isTimeoutReached()) {
+            throw new XrayWrapperException("Reached maximum scan timeout of " + xrayWrapperConfiguration.getMaxScanDurationHours() + " hours. "
+                    + "Started scan at: " + context.startTime, XrayWrapperExitCode.TIMEOUT_REACHED);
+        }
+        switch (context.getStatus()) {
+        case NOT_SCANNED -> {
+            waitSeconds(xrayWrapperConfiguration.getWaitUntilRetrySeconds());
+            if (context.isMaximumRetriesReached()) {
```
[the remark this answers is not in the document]
lorriborri: reset ony called once (in not scanned)

Jeeppler (approved):
Looks good. Can be merged.

sechub-wrapper-xray/src/test/java/com/mercedesbenz/sechub/wrapper/xray/util/IOHelperTest.java
```diff
@@ -0,0 +1,50 @@
...
+        assertThrows(NullPointerException.class, () -> IOHelperToTest.readInputStreamAsString(null));
+    }
+
+    @Test
+    void readInputStreamAsString_throws_xrayWrapperException() {
+        /* prepare */
+        InputStream mockedIs = mock(InputStream.class);
+
+        /* execute + test */
+        assertThrows(XrayWrapperException.class, () -> IOHelperToTest.readInputStreamAsString(mockedIs));
```
de-jcup, line 48:
```suggestion
        XrayWrapperException exception = assertThrows(XrayWrapperException.class, () -> IOHelperToTest.readInputStreamAsString(mockedIs));
        assertEquals("expected message", exception.getMessage());
```

sechub-wrapper-xray/src/test/java/com/mercedesbenz/sechub/wrapper/xray/util/IOHelperTest.java
```diff
@@ -0,0 +1,50 @@
...
+        String content = IOHelperToTest.readInputStreamAsString(inputStream);
+
+        /* test */
+        assertEquals(testinputString, content);
+    }
+
+    @Test
+    void readInputStreamAsString_throws_nullPointerException() {
+        /* execute + test */
+        assertThrows(NullPointerException.class, () -> IOHelperToTest.readInputStreamAsString(null));
```
de-jcup, line 39: Either just no test, or we throw here an illegal argument exception. Testing a null exception is ... not very useful IMO

sechub-wrapper-xray/src/main/java/com/mercedesbenz/sechub/wrapper/xray/cli/XrayWrapperArtifactoryClientSupport.java
```diff
@@ -0,0 +1,257 @@
...
+import com.mercedesbenz.sechub.wrapper.xray.config.XrayWrapperConfiguration;
+import com.mercedesbenz.sechub.wrapper.xray.report.XrayWrapperReportSupport;
+
+public class XrayWrapperArtifactoryClientSupport {
+
+    private static final Logger LOG = LoggerFactory.getLogger(XrayWrapperArtifactoryClientSupport.class);
+    private final XrayWrapperConfiguration xrayWrapperConfiguration;
+    private final XrayAPIArtifactoryClient artifactoryClient;
+
+    XrayWrapperReportSupport reportReader;
```
de-jcup, line 25:
```suggestion
    XrayWrapperReportSupport reportSupport;
```

sechub-wrapper-xray/src/main/java/com/mercedesbenz/sechub/wrapper/xray/cli/XrayWrapperArtifactoryClientSupport.java
```diff
@@ -0,0 +1,257 @@
...
+            readAndConvertReports();
+        }
+
+        // delete artifact from artifactory
+        artifactoryClient.deleteArtifact();
+        artifactoryClient.deleteUploads();
+    }
+
+    private void readAndConvertReports() throws XrayWrapperException {
+        XrayWrapperReportData xrayWrapperReportData = new XrayWrapperReportData();
```
de-jcup, line 102: Just a thought:
For simpler testing and maybe also for maintainance, extracting the 
read and convert reports method to its own class could have been helpful.

Also a return value or the context object as parameter could have been helpful - at least for testing.
It would be possible to check for the result file, compare it etc.
Also for the logic this could have been interesting - at least to log the report creation.

Another thing: IMO I would have maybe returned the report file as

sechub-wrapper-xray/src/test/java/com/mercedesbenz/sechub/wrapper/xray/util/IOHelperTest.java
```diff
@@ -0,0 +1,50 @@
...
+import java.io.InputStream;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import com.mercedesbenz.sechub.wrapper.xray.XrayWrapperException;
+
+public class IOHelperTest {
+
+    final String testinputString = "test data";
```
de-jcup, line 17: IMHO it would be better to define the content of "testInputString" inside the @beforeEach method.
Reasons:
- mainly because every default setup is done in @BeforeEach method, I did not see at the first glimpse, that there was a field declaration as well. To have defaults centrally defined is a good way.
- you could get rid of the "final" part
- you could change the test inputString inside a test 

Another option would be to change it to constant : 

```suggestion
    private final static String TEST_INPUT_STRING= "test data";
```
-> it would be clear that the data
- cannot be changed
- is constant
- initialized as field

sechub-wrapper-xray/src/test/java/com/mercedesbenz/sechub/wrapper/xray/util/XrayWrapperConfigurationHelperTest.java
```diff
@@ -0,0 +1,50 @@
...
+    @Test
+    void createXrayConfiguration_with_valid_parameters() throws XrayWrapperException {
+        /* prepare */
+        XrayWrapperConfiguration xrayWrapperConfiguration;
+
+        /* execute + test */
+        // for valid parameters the artifactory and the registry can not be null or
+        // empty
+        try (MockedConstruction<EnvironmentVariableReader> mocked = mockConstruction(EnvironmentVariableReader.class, (mock, context) -> {
+            when(mock.readEnvAsString(EnvironmentVariableConstants.ARTIFACTORY_ENV)).thenReturn("artifactoryMock");
```
de-jcup, line 34: - please define comments for prepare, execute and test at correct locations

sechub-wrapper-xray/src/main/java/com/mercedesbenz/sechub/wrapper/xray/XrayWrapperException.java
```diff
@@ -0,0 +1,22 @@
+package com.mercedesbenz.sechub.wrapper.xray;
+
+import com.mercedesbenz.sechub.wrapper.xray.cli.XrayWrapperExitCode;
+
+public class XrayWrapperException extends Exception {
+
```
de-jcup, line 6: To avoid warnings about missing serial version uids, please add


```suggestion
private static final long serialVersionUID = 1L, 
```

sechub-wrapper-xray/src/main/java/com/mercedesbenz/sechub/wrapper/xray/api/XrayAPIArtifactoryClient.java
```diff
@@ -0,0 +1,116 @@
...
+        this.artifact = artifact;
+        this.xrayWrapperConfiguration = xrayWrapperConfiguration;
+    }
+
+    public String requestXrayVersion() throws XrayWrapperException {
+        XrayAPIRequest request = XrayAPIRequestFactory.createGetXrayVersionRequest(xrayWrapperConfiguration.getArtifactory());
+        XrayAPIResponse response = send(request);
+        assertNoError(response, "Could not get Xray Version from Artifactory");
+        JsonNode node = XrayWrapperJSONConverter.get().readJSONFromString(response.getBody());
+        return node.get("xray_version").asText();
```
de-jcup, line 38: We introduced constants - would this one also be a constant part?

sechub-wrapper-xray/src/main/java/com/mercedesbenz/sechub/wrapper/xray/api/XrayAPIArtifactoryClient.java
```diff
@@ -0,0 +1,116 @@
...
+
+    public XrayAPIArtifactoryClient(XrayWrapperArtifact artifact, XrayWrapperConfiguration xrayWrapperConfiguration) {
+        this.artifact = artifact;
+        this.xrayWrapperConfiguration = xrayWrapperConfiguration;
+    }
+
+    public String requestXrayVersion() throws XrayWrapperException {
+        XrayAPIRequest request = XrayAPIRequestFactory.createGetXrayVersionRequest(xrayWrapperConfiguration.getArtifactory());
+        XrayAPIResponse response = send(request);
+        assertNoError(response, "Could not get Xray Version from Artifactory");
```
de-jcup, line 36: Just a suggestion: Sometimes a blank line between code parts make it more readable
```suggestion

        assertNoError(response, "Could not get Xray Version from Artifactory");
        
```

sechub-wrapper-xray/src/main/java/com/mercedesbenz/sechub/wrapper/xray/api/XrayAPIArtifactoryClient.java
```diff
@@ -0,0 +1,116 @@
...
+        return true;
+    }
+
+    public void startArtifactScan() throws XrayWrapperException {
+        XrayAPIRequest request = XrayAPIRequestFactory.createScanArtifactRequest(xrayWrapperConfiguration.getArtifactory(), artifact,
+                xrayWrapperConfiguration.getRegistry());
+        XrayAPIResponse response = send(request);
+        assertNoError(response, "Could not start external Xray scan");
+        JsonNode node = XrayWrapperJSONConverter.get().readJSONFromString(response.getBody());
+        String result = node.get("info").asText();
```
de-jcup, line 71: info as a constant?

sechub-wrapper-xray/src/main/java/com/mercedesbenz/sechub/wrapper/xray/api/XrayAPIArtifactoryClient.java
```diff
@@ -0,0 +1,116 @@
...
+    private XrayAPIResponse send(XrayAPIRequest request) throws XrayWrapperException {
+        HttpURLConnection con = apiHttpUrlConnectionFactory.create(request);
+        return xrayAPIResponseFactory.createHttpResponseFromConnection(con, xrayWrapperConfiguration.getZipDirectory());
+    }
+
+    private boolean isErrorResponse(XrayAPIResponse response) {
+        int statusCode = response.getHttpStatusCode();
+        LOG.debug("Response Code: {}", statusCode);
+        LOG.debug("Response Message: {}", response.getResponseMessage());
+        LOG.debug("Response Headers: {}", response.getHeaders().toString());
```
de-jcup, line 106: When you call the toString() method in this way, it is always called, even when debug is not enabled. Also it could be possible that headers is null - in this case we had a NPE even for a non-wanted debug message in production.

On the other side: If you give an object to slf4j which is not wellknown, slf4j will always call the `toString` method automatically!

So you can simple change this to:
```suggestion
        LOG.debug("Response Headers: {}", response.getHeaders());
```

The benefit here: toString is only called by slf4j when debugging is enabled. Also null is no problem, because it is handled by slf4j.

#### Just as an information

If you would need something else then `toString` but also only when debugging is enabled, you have to use the `isDebugEnabled` method in this case, to avoid unnecessary object creation. The following code is an example for such handling (but only nonsense output)

```suggestion
        if (LOG.isDebugEnabled()) {
           var headers = response.getHeaders();
           if (headers==null){
                LOG.warn("Headers null, cannot provide debug logging!");
           }else{
                LOG.debug("Response Headers fancy output: {}", headers.createFancyFullDescription());
           }

        }
        LOG.debug("Response Headers: {}", response.getHeaders());
```

sechub-wrapper-xray/src/test/java/com/mercedesbenz/sechub/wrapper/xray/util/XrayWrapperConfigurationHelperTest.java
```diff
@@ -0,0 +1,50 @@
...
+        })) {
+            xrayWrapperConfiguration = helperToTest.createXrayConfiguration(XrayWrapperScanTypes.DOCKER, "output", "workspace");
+            assertEquals("output", xrayWrapperConfiguration.getXrayPdsReport());
+            assertEquals("https://artifactoryMock", xrayWrapperConfiguration.getArtifactory());
+            assertEquals("registerMock", xrayWrapperConfiguration.getRegistry());
+        }
+    }
+
+    @Test
+    void createXrayConfiguration_with_null_parameters_throws_xrayWrapperException() {
```
de-jcup, line 46: either remove the test or handle it different.
You are testing 3 null values at once and that (one of them) it throws an NPE.

NPEs are unexpected - the test here is IMO useless.

If we would change to an IllegalArgumentException inside the implematation and if there would be 4 different tests it would make more sense:

Tests:
- `xray_config_creation_throws_illegal_argument_exc_when_first_parameter_is_null`
- `xray_config_creation_throws_illegal_argument_exc_when_second_parameter_is_null`
- `xray_config_creation_throws_illegal_argument_exc_when_third_parameter_is_null`
- `xray_config_creation_works_when_all_parameters_are_valid`

sechub-wrapper-xray/src/test/java/com/mercedesbenz/sechub/wrapper/xray/util/ZipFileExtractorTest.java
```diff
@@ -0,0 +1,84 @@
...
+        }
+        mockFileInput.close();
+    }
+
+    @Test
+    void unzipFile_with_zip_files() throws IOException, XrayWrapperReportException {
+        /* prepare */
+        Path target = createTempDirectoryInBuildFolder("xray-zipFileExtractor");
+        Path source = Paths.get("src/test/resources/xray-zip-test/xray-zip-test.zip");
+        Path expectedFile01 = Paths.get(target + "/xray-testfile-01.txt");
```
de-jcup, line 65: The string concatenation does call internally the toString() method and could be result in errors (e.g. when forgetting the following slash on next string)

I think you should use the path.resolve(..) method here as convenient way:
```suggestion
        Path expectedFile01 = target.resolve("xray-testfile-01.txt");
```
this could be applied in the complete test setup.

sechub-wrapper-xray/src/test/java/com/mercedesbenz/sechub/wrapper/xray/util/ZipFileExtractorTest.java
```diff
@@ -0,0 +1,84 @@
...
+        Path expectedFile02 = Paths.get(target + "/xray-testfile-02.txt");
+        Path expectedFile03 = Paths.get(target + "/xray-test-zip-folder/xray-testfile-03.txt");
+        Path expectedFileSomeFile = Paths.get(target + "/xray-test-zip-folder/.somefile");
+
+        /* execute */
+        zipFileExtractorToTest.unzipFile(source, target);
+
+        /* test */
+        assertTrue(Files.exists(expectedFile01));
```
de-jcup, line 74: What about a
```suggestion
     assertFileExists(expectedFile)
```
in combination with
```suggestion
private void assertFileExists(Path file){
     if (Files.exists(file)){
        return;
     }
     fail("The expected file does not exist: "+file);
}

```
Benefit here: When your junit test fails, you have already the path output inside the failure message and it is clear what the problem is. Maybe you see also the problem inside the path output itself.

sechub-wrapper-xray/src/test/java/com/mercedesbenz/sechub/wrapper/xray/util/ZipFileCreatorTest.java
```diff
@@ -0,0 +1,23 @@
...
+    ZipFileCreator creatorToTest;
+
+    @BeforeEach
+    void beforeEach() {
+        creatorToTest = new ZipFileCreator();
+    }
+
+    @Test
+    void saveInputStreamToZipFile_throws_nullPointerException() {
+        /* execute + test */
```
de-jcup, line 19: Please remove the NPE test or provide illegal argument exception and check messages etc.

de-jcup (changes requested):
Good work - but I still found some issues. Please look at my comments.

de-jcup (approved):
Looks good to me - good job :+1:

PR #2612, opened 2023-10-25
Changed content type for full scan log REST to "application/zip" #2608

### LINKED ISSUE #2608: Produce definition for fetching full scan log files is not `application/zip`
## Situation
While implementing #2606 and enhancing the Sechub java api client, the download for @UseCaseAdminDownloadsFullScanDataForJob did not work.

The reason is that the Spring annotation for the rest controller defines that it produces "application/json"  - which is wrong and leads to client exception.

## Wanted
Shall work as expected

## Solution
Change to "application/zip" at produces method

de-jcup:
This PR
- closes #2608


winzj (approved):
Looks good 👍

PR #2651, opened 2023-11-08
Add client certificate option to sechub web scan config #2524

### LINKED ISSUE #2524: Extend SecHub webscan config by certificate part to enable client certificate authentication like mTLS
### Situation
Currently we do not support client certificate based authentication for our web scan module. 

### Situation
Extend the Sechub webscan configuration with a client certificate part.
We need entries for:
- a data reference for the client certificate file which shall be uploaded
- the password the certificate is protected with

Inside the configuration file it could look like this, referencing a single file like for the openApi file definition we already use:
```json
{
  "apiVersion" : "1.0",
  "project" : "example_project",
  "data" : {
    "sources" : [ {
      "name" : "client-certificate-file-reference",
      "fileSystem" : {
        "files" : [ "backend-cert.p12" ]
      }
    } ]
  },
  "webScan" : {
    "url" : "https://my-app.com",
    "clientCertificate" : {
      "password" : "{{ .CERT_PASSWORD }}",
      "use" : [ "client-certificate-file-reference" ]
    }
  }
}
```

### Important notice
Our current DAST solution with the Zaproxy can only configure `PKCS #12` via the java client api.
The Zaproxy itself supports `PKCS #11` as well, but after the first evaluation it looks like it cannot be configured via the java client api.

winzj:
- closes #2524


de-jcup (approved):
Good job - looks good to me :+1:

PR #2655, opened 2023-11-10
Added Rust and Lua to supported fileextension

### LINKED ISSUE #2654: Add file extensions for Lua and Rust to the client
## Situation:
The SecHub client filters for known file extensions and does not include anything else. 

## Wanted:
The languages Lua and Rust should be included in the supported file extensions list.

## To Do:
The following file extensions should be added:

```
Rust: .rs .rlib
Lua: .lua
```

haerter-tss:
The file extensions for Lua and Rust has been added.

closes #2654


sven-dmlr (approved):
Looks good :+1:

PR #2681, opened 2023-11-17
Storing now test reports for webscan, secretscan and codescan #2680

### LINKED ISSUE #2680: Store reports of integration tests which download reports inside dedicated build folder
## Wanted
We have some integration tests which do already download a HTML report or a json report file

## Wanted
If would be nice to have these reports automatically as a example outputs after a build.

## Solution
Enhance TestAPI similar to this

```java
public class TestApi

private static File testReportStorageFolder = new File("./build/sechub-test-reports");
private static TextFileWriter writer = new TextFileWriter();

public static void storeTestReport(String fileName, String text){
        writer.save(new File(folder, fileName, report, true);
}

```
and integrate the calls where a report in html or in json is already done by existing
integration tests.

de-jcup:
This PR
- closes #2680

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -116,6 +116,14 @@ jobs:
           path: sechub-cli/build/go
           retention-days: 14
 
+      - name: Archive sechub integration test report artifacts
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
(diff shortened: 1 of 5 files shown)


sven-dmlr (approved):
Looks good to me :+1:

PR #2699, opened 2023-11-24
client: support certificates in scans for secrets #2698

### LINKED ISSUE #2698: client: support certificates in scans for secrets
## Situation
When uploading sources, only known file extensions for programming languages/source code are added by default.

When doing a scan for secrets, it's desirable to have files containing certificates also uploaded per default so the secrets scanner has a chance to detect them.

## ToDo
Add these extensions to the list of source code file extensions:
`.pem .der .crt .cer .csr .pfx .p12 .p7b .p7c`

sven-dmlr:
- closes #2698


winzj (approved):
Looking good 👍

PR #2716, opened 2023-11-29
Systemtest sechub config file is now written direct after creation #2715

### LINKED ISSUE #2715: Ensure in systemtests a  `sechub-config.json` is written even when a Job cannot be created
## Situation

Inside as system test, the created artifacts will be stored inside the workspace, including the generated `sechub-config.json` :

image

But currently the `sechub-config.json` is written to disk before the upload is done - which means after the job creation.

This is a problem, when it comes to the situation that a job creation fails - inside a systemtest from console, we have currently not the chance to check the generated `sechub-config.json` because it is not stored at this moment:

image

## Wanted
The `sechub-config.json` shall be stored as soon as possible, so it is available, even if job creation fails.

de-jcup:
This PR
- closes #2715


lorriborri (approved):
Looks good to me! Good Work! :fairy: :star2:

PR #2719, opened 2023-11-29
Introduced and used JobInformationListEntry for lists #2717

### LINKED ISSUE #2717: JobInformation list shall not contain full job information
## Situation
This is a sub issue of #2366 

Currently when a administrator fetches the running jobs, the configuration for each job is returned as well. This is much unnecessary data, but also can contain test credentials etc.

## Wanted
We want to reduce the job information list entries returned on usecase `UC_ADMIN_LISTS_ALL_RUNNING_JOBS`

The returned data shall not contain the configuration any more or unnecessary details.

## Solution

We only support following information inside list entries:
- jobUUID;
- since;
- status;
- projectId (this is still interesting, because an admin can see if the jobs belong to same project)

### LINKED ISSUE #2718: Introduce usecase `UC_ADMIN_SHOWS_JOB_DETAILS`
## Situation
With #2717 we removed the configuration information from the REST result of usecase `UC_ADMIN_LISTS_ALL_RUNNING_JOBS`.

But sometimes it can be necessary that an administrator checks the configuration - e.g. when users have a problems to get their jobs running.

(Remark: this is also a sub issue of #2366)

## Wanted
An administrator shall have the possibility to still get details for a job - when necessary.

## Solution

- [ ] Introduce new usecase `UC_ADMIN_SHOWS_JOB_DETAILS`
- [ ] Implement audit logging for this case, to log that this happend
- [ ] Implement REST controller
- [ ] Write a RESTdoc test (for documentation + open api file+ api generation)

de-jcup:
This PR
- closes #2717 
- closes #2718

### DIFF
--- a/sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/job/JobAdministrationRestController.java
+++ b/sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/job/JobAdministrationRestController.java
@@ -56,7 +56,7 @@ public class JobAdministrationRestController {
 				description="Administrator lists all running jobs by calling rest api"))
 	@RequestMapping(path = AdministrationAPIConstants.API_LIST_JOBS_RUNNING, method = RequestMethod.GET, produces= {MediaType.APPLICATION_JSON_VALUE})
 	@ResponseStatus(HttpStatus.OK)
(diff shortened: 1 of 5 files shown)


de-jcup:
Outdated - with #3254 we have now full encrypted entries and also removed REST endpoints. So closing this PR.

PR #2751, opened 2023-12-13
Provide test reports from PDS solution mocks #2750

### LINKED ISSUE #2750: Provide test reports from PDS solution mocks in `sechub-integration-test-reports.zip `
## Situation
We already have `sechub-integration-test-reports.zip` file in every github CI/CD build.

But only for some special integration test with mocked data.
This is very inconvenient - e.g. when developing plugins (we need example report data).

## Wanted
`sechub-integration-test-reports.zip` shall containalso results from real products

## Solution 
In our existing PDS solutions we have always a mock file with real product data (for "mock mode").

We add test products in our PDS integration test server. Here we use simple scripts which do just return the mock files (but from real products) back to SecHub.

In new integration tests we start the scans, fetch HTML and json reorts and dump them as reports to disk by TestAPI.
In case of scan type: licenseScan, we download also the SPDX report via SecHub.

All reports will be available in `sechub-integration-test-reports.zip`

de-jcup:
This PR
- closes #2750


Jeeppler (changes requested):
Overall looks good. I found a few minor issues. Have a look at the comments.

Jeeppler (approved):
Looks good 👍 . Can be merged.

PR #2755, opened 2023-12-18
Improved false-positive handling for missing end elements #2754

### LINKED ISSUE #2754: FalsePositive location strategy problem on missing end elements
## Situation

The false positive handling is done by loading following data:

- `FalsePositiveMeta` is fetched from database (it contains relevant information from a Secub report which was used by false positive marking via job uuid and finding id)
- `SecHubFinding` the current SERECO findings is returned by products (contains all findings, no matter if marked as false positive or not)

After this a `FalsePositiveStrategy` is applied and locations CweId etc. are compared. In case of locations the start code element and the end element are inspected both and compared between SERECO finding and an false positive element. When start and end elements have same locations  (and other checks are fine as well), the strategy will accept this as an existing false positive handle and the report will not show up this one.

But...
What happens when meta data does contain an end element but the SecHub fiding from Sereco has only a start element?
Same the other way: What if an existing false positive meta data has only a start element but the SecHub finding has start and end?

## Wanted
False positive finding shall also work when only one code entry element is resolved in one product

## Solution
Change `AbstractSourceCodeBasedFalsePositiveStrategy` and write tests to ensure expected behavior.
When no end element is available, the start element shall be always the fallback.

de-jcup:
This PR
- closes #2754


sven-dmlr (approved):
Looks good to me.
Change successfully tested.
You can go ahead and merge the PR. :-)

PR #2761, opened 2023-12-20
Introduce gradle task buildJavaApiAll

### LINKED ISSUE #2740: Provide gradle task `buildJavaApiAll` for plugin development
## Situation
To update the existing java based IDE plugins (Eclipse, Jetbrains IDEA) we need an api client.

The existing old build task to generate fat jars was removed.

## Wanted
Developer can call 
`./gradlew buildJavaApiAll -Dsechub.build.stage=all` 

After this `/sechub-api-java/build/libs/sechub-java-api-all-0.0.0.jar` is available and can be used.
The jar contains all other dependencies.

de-jcup:
This PR
- closes #2740

### DIFF
--- a/sechub-api-java/build.gradle
+++ b/sechub-api-java/build.gradle
@@ -164,3 +164,18 @@ tasks.named("eclipse") {
     dependsOn("assemble")
 }
 
+/**
+ * Custom gradle task to build a "fatJar".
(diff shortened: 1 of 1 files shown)


winzj (approved):
Looking good 👍

PR #2805, opened 2024-01-05
Proper docs versioning in build + release job  #2798

### LINKED ISSUE #2798: Docs: Versioning works not as expected in release job
## Situation
Job: `.github/workflows/release-client-server-pds.yml`
The latest version of a not released product cannot be obtained.
Example from last release:
The client version should remain at v1.3.0 instead of v0.0.0
```html
<meta http-equiv="refresh" content="0; URL=https://github.com/mercedes-benz/sechub/releases/download/v0.0.0-client/sechub-cli-0.0.0.zip">
```

## Cause
The git tag history is not checked out

## ToDo
Check out the git tag history

sven-dmlr:
- Actions workflow now fetches git tag history so latest versions can be obtained for documentation versioning
- Extended the `VersionData` Groovy class so the versions for docs can also be stored
- Added the same to the Gradle files
- Tested successfully with `./gradlew documentation-with-pages`
- closes #2798

### DIFF
--- a/.github/workflows/release-client-server-pds.yml
+++ b/.github/workflows/release-client-server-pds.yml
@@ -69,6 +69,8 @@ jobs:
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
         with:
           ref: master
+          fetch-tags: true
+          fetch-depth: 0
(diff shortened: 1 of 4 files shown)


winzj (approved):
Looking good to me 👍

PR #2812, opened 2024-01-09
Fix ZAP configuration order #2811

### LINKED ISSUE #2811: Fix ZAP wrapper configuration call order
### Situation
Currently the client certificate is loaded after the openapi definitions. This can cause issues if someone requires both.

### Solution
Make sure the client certificate is loaded before the openapi definitions.

winzj:
- closes #2811 
- fix order to load client certificate before openApi definitions
- improve error handling if the client certificate file was not extracted correctly
- update test cases to match improved error handling

### DIFF
--- a/sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/zapwrapper/scan/ZapScanner.java
+++ b/sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/zapwrapper/scan/ZapScanner.java
@@ -92,9 +92,9 @@ public class ZapScanner implements ZapScan {
             addReplacerRulesForHeaders();
 
             /* ZAP setup with access to target */
+            importClientCertificate();
             addIncludedAndExcludedUrlsToContext();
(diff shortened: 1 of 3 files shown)


sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/zapwrapper/scan/ZapScanner.java
```diff
@@ -92,9 +92,9 @@ public void scan() throws ZapWrapperRuntimeException {
             addReplacerRulesForHeaders();
 
             /* ZAP setup with access to target */
+            importClientCertificate();
```
Jeeppler, line 95: Please add a comment explaining that the order of the calls is important and especially that the client certificate needs to be loaded at the very first.

Jeeppler (changes requested):
Overall looks good. However, I found one issue.

Jeeppler (approved):
Looks good. Can be merged 💯

PR #2860, opened 2024-01-24
client: defineFalsePositives feature #2077

### LINKED ISSUE #2077: Introduce sechub client action `defineFalsePositives`
## Situation
SecHub is able to 
- mark false positives by REST API
  (see  https://mercedes-benz.github.io/sechub/latest/sechub-restapi.html#user-marks-false-positives-for-finished-sechub-job) 
- unmark false positives by REST API
  (see https://mercedes-benz.github.io/sechub/latest/sechub-restapi.html#user-unmarks-existing-false-positive-definitons)

In both cases, we only provide the new information as JSON content for REST.

Here an example:
- job "A" has been started and returns a finding number 12 which is a false positive and shall be marked
- user "alice" marks finding 12 for job "A" as a false positive and next scans will ignore it
- if user "bob" marks finding 12 for job "A" again, this will be just ignored

But we are asked from many people that they want the possibility to maintain only one single `sechub-false-positives.json` file where they define their false positives at a single point. 

If a false positive is removed from the file, the existing false positive shall be removed as well. If a new one is defined, a new false positive shall be marked.

## Wanted
It shall be possible to maintain easily false positives by a  `sechub-false-positives.json` file and a new client action `defineFalsePositives`.

## Solution
- we introduce a new action for client : `defineFalsePositives`
- we need no change on server side here
- the client will just do following here:
  1. load the information from `sechub-false-positives.json`
  2. load the existing false positive list for sechub-project via existing REST API
  3. calculate the delta for i and ii
  4. all (on server side) existing false positives which are no longer mentioned inside ii will be unmarked
  5. all (on server side) non existing false positives are marked by client
- defaults to `sechub-false-positives.json` in root folder but we should have the possibility to define another file location via parameter

### LINKED ISSUE #2862: client: remove redundant code when reading file
## Situation
We have redundant code to read the content of `context.config.file` into `context.inputForContentProcessing`:
```go
	jsonFile, err := os.Open(context.config.file)
	defer jsonFile.Close()
	failed := sechubUtil.HandleIOError(err)

	context.inputForContentProcessing, err = ioutil.ReadAll(jsonFile)
	failed = sechubUtil.HandleIOError(err)

	if failed {
		showHelpHint()
		os.Exit(ExitCodeIOError)
	}

```

## ToDo
Create a function which does the same (plus more error checking) and call it to replace the redundant code.

sven-dmlr:
- closes #2077 
- closes #2862

### DIFF
--- a/sechub-cli/src/mercedes-benz.com/sechub/cli/config.go
+++ b/sechub-cli/src/mercedes-benz.com/sechub/cli/config.go
@@ -225,7 +225,7 @@ func assertValidConfig(context *Context) {
 		getReportAction:                       {"server", "user", "apiToken", "projectID", "secHubJobUUID"},
 		getFalsePositivesAction:               {"server", "user", "apiToken", "projectID"},
 		listJobsAction:                        {"server", "user", "apiToken", "projectID"},
-		defineFalsePositivesAction:            {"server", "user", "apiToken", "projectID", "file"},
+		defineFalsePositivesAction:            {"server", "user", "apiToken", "projectID"},
 		markFalsePositivesAction:              {"server", "user", "apiToken", "projectID", "file"},
 		unmarkFalsePositivesAction:            {"server", "user", "apiToken", "projectID", "file"},
 		interactiveMarkFalsePositivesAction:   {"server", "user", "apiToken", "projectID"},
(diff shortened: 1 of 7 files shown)


winzj (approved):
Looking good, let's get it merged 🚀

PR #2865, opened 2024-01-26
username case correction works with getReport #2861

### LINKED ISSUE #2861: Client: Add case correction in getReport
## Situation
As a convenience function, the SecHub client converts IDs (projectID, userID) to lowercase.

When calling `getReport` without providing a jobUUID, you get a 401 error when using a userID with wrong case.

Example:
```bash
> sechub -user TestUser -configfile ../sechub.json getreport
 _____           _   _       _     
/  ___|         | | | |     | |    
\ `--.  ___  ___| |_| |_   _| |__  
 `--. \/ _ \/ __|  _  | | | | '_ \ 
/\__/ /  __/ (__| | | | |_| | |_) |
\____/ \___|\___\_| |_/\__,_|_.__/ Client Version 1.3.0-20231124124949

ERROR: The SecHub server responded with HTTP status code '401'
body={"status":401,"error":"Unauthorized","message":"Unauthorized","details":[],"timeStamp":"Thu Jan 25 08:12:06 GMT 2024"}
Hint: Unauthorized - Please check if your SecHub credentials (userID and API-token) are valid.
```

## Wanted
Above example shall work.

## Solution
The conversion must happen before the first api call to SecHub server.

sven-dmlr:
- used to throw a 401; now fixed
- test added for the same
- closes #2861

### DIFF
--- a/sechub-cli/src/mercedes-benz.com/sechub/cli/config_test.go
+++ b/sechub-cli/src/mercedes-benz.com/sechub/cli/config_test.go
@@ -137,7 +137,7 @@ func Example_validateRequestedReportFormatMakesLowercase1() {
 	// TEST
 	fmt.Println(config.reportFormat)
 	// Output:
-	// NOTICE: Converted requested report format 'HTML' to lowercase. Because it contained uppercase characters, which are not accepted by SecHub server.
+	// NOTICE: Converted requested report format 'HTML' to lowercase because it contained uppercase characters, which are not accepted by SecHub server.
 	// html
 }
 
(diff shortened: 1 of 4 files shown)


sechub-cli/src/mercedes-benz.com/sechub/cli/config_test.go
```diff
@@ -504,6 +504,43 @@ func Example_will_reportfile_be_found_in_current_dir() {
...
+
+	// EXECUTE
+	assertValidConfig(context)
+
+	// TEST
+	// Restore original arguments
+	os.Args = originalArgs
+
+	// Output:
+	// NOTICE: Converted user id 'TESTUSER' to lowercase. Because it contained uppercase characters, which are not accepted by SecHub server.
```
haerter-tss, line 540:
```suggestion
	// NOTICE: Converted user id 'TESTUSER' to lowercase because it contained uppercase characters, which are not accepted by SecHub server.
```

haerter-tss (approved):

sechub-cli/src/mercedes-benz.com/sechub/cli/config_test.go
sven-dmlr: This is the client's output. If I change it only here, the test will fail.

But I will change the client's output because you suggestion is better english. :+1:

PR #2914, opened 2024-02-21
Feature 2900 SecHub Prepare Phase

### LINKED ISSUE #2900: Extend SecHub server for prepare phase
## Wanted
SecHub Server should be able to handle remote data

## Situation
Currently only "real" data (source code or binaries) can be handled 

## Solution
A prepare phase needs to be implemented before the analyze phase in the Server 

related to: #2722

### LINKED ISSUE #2921: Special handling for prepare phase
## Situation
The analytics and the prepare phase should not 

## Wanted

- [x] SecHub user messages from preparation phase shall appear inside the report
- [x] "No results from a security product available for this job!" shall appear in reports when preparation phase is done, but no results from security products are available
- [x] When at least one preparation execution fails, the complete SecHub job must stop and fail without executing other parts

### LINKED ISSUE #2926: Non deterministic behavior in SecHub report messages when PDS has internal problems
## Situation
While implementing #2921 we had the situation that an Integration test, where a PDS script does exit with code 5 in script, failed always on github actions, but never locally.

## Wanted
Clear and unique behavior

## Solution
- Analyze
- Fix the problem

lorriborri:
- [x] closes #2900 
- [x] closes #2921
- [x] closes #2926


de-jcup (approved):
Looks good to me

PR #2929, opened 2024-02-23
Username now always extracted from BasicAuth added to security log #2907

### LINKED ISSUE #2907: Add username to 401 Unauthorized log message
# Situation
If someone tries to access SecHub with a wrong user and / or wrong API token a log message of the type "WRONG_USAGE" and "401 UNAUTHORIZED" is created. Missing in this message is the name of the user that was tried. This would be helpful for various troubleshooting scenarios.

Example:

image

# Wanted
The user name should be added to the log messages 

# Solution
- Implement extraction for basic auth user
- add to security log data

## Extraction
- returns `info.no-basic-auth-defined` when not basic auth info available (e.g. when anonymous)
- returns `error.unsupported-format` when its not a basic auth information
- returns `error.too-small` when "Basic" found inside authorization header, but not long enough
- returns `error.colon-count-wrong:$colon`
- returns $username from basic auth otherwise - when empty username, the return value is empty

### LINKED ISSUE #2928: Sort security log data output
## Wanted
Security log data output shall be sorted

de-jcup:
This PR

- [x] closes #2907 
- [x] closes #2928

### DIFF
--- a/sechub-shared-kernel/src/main/java/com/mercedesbenz/sechub/sharedkernel/logging/BasicAuthUserExtraction.java
+++ b/sechub-shared-kernel/src/main/java/com/mercedesbenz/sechub/sharedkernel/logging/BasicAuthUserExtraction.java
@@ -1,0 +1,57 @@
+package com.mercedesbenz.sechub.sharedkernel.logging;
+
+import java.util.Base64;
+
+import org.slf4j.Logger;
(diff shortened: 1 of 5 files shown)


sven-dmlr (approved):
Looks good to me.
Also successfully tested :+1:

PR #2931, opened 2024-02-23
Sanitize parsed basic auth information #2930

### LINKED ISSUE #2930: Sanitize extracted basic auth information in Security Log
## Situation
With #2907 the basic auth information is now parsed from header for every security log entry.
But the implementation did not sanitize the output - all other parts are sanitized.

## Wanted
We should sanitize this value

## Solution
Use log sanitizer inside the implementation and reduce to 100 characters only .
Log forging shall not be done here.

de-jcup:
This PR

- [ ] closes #2930


sven-dmlr (approved):
Looks good :+1:

PR #2954, opened 2024-03-04
Provide SarifImportProductWorkaround interface and provide implementation for gitleaks

### LINKED ISSUE #2952: Provide SarifImportProductWorkaround interface and provide implementation for gitleaks
## Situation
When import gitleak reports by SARIF, the git leak sarif does not contain the rule name inside the report but instead the regular expression is shown. We want to fix this upstream with #2951

But in the mean time we need here a workaround

## Wanted
Until fixed upstream, our sarif imports for git leaks shall fix this automatically. The workaround
shall be easy to remove when fixed upstream.

## Solution
We introduce interface `SarifImportProductWorkaround`.
```java

public interface SarifImportProductWorkaround {

     public default String resolveType(ReportingDescriptor rule, Run run) {
          /* per default product workaround does not handle it */
          return null;
     }
}

```

```java
@Component
public class SarifImportProductWorkaroundComponent {

     @Autowired
     List<SarifImportProductWorkaround> workarounds;

     public String String resolveType(ReportingDescriptor rule, Run run) {
        // iterate over workarounds, first non null result -> return
   
     } 

}
```

and add an implementation: `GitLeaksSarifImportWorkaround` which checks if run is from tool gitleaks and in this case handle the type resolution different. The implementation is marked as a spring boot component -> auto wired.

`SarifV1JSONImporter` shall use the workaround component inside `resolveType(...)` -> if type null, use default (old implementation)

### Benefits
If we have other special needs for other products, it easy to adopt those changes in own implementations. If there are additional parts which shall be handled, the interface will just be extended (and with default implementation returning null)

Also it is very easy to remove the workaround when a product changes. Just remove the annotation or delete the class.

winzj:
- closes #2952

### DIFF
--- a/sechub-sereco/src/main/java/com/mercedesbenz/sechub/sereco/importer/GitleaksSarifImportWorkaround.java
+++ b/sechub-sereco/src/main/java/com/mercedesbenz/sechub/sereco/importer/GitleaksSarifImportWorkaround.java
@@ -5,6 +5,7 @@ import org.springframework.stereotype.Component;
 
 import de.jcup.sarif_2_1_0.model.ReportingDescriptor;
 import de.jcup.sarif_2_1_0.model.Run;
+import de.jcup.sarif_2_1_0.model.Tool;
 import de.jcup.sarif_2_1_0.model.ToolComponent;
 
 @Component
@@ -25,13 +26,14 @@ public class GitleaksSarifImportWorkaround implements SarifImportProductWorkarou
         if (run == null) {
             return false;
         }
-        if (run.getTool() == null) {
+        Tool tool = run.getTool();
+        if (tool == null) {
             return false;
         }
-        ToolComponent driver = run.getTool().getDriver();
+        ToolComponent driver = tool.getDriver();
         if (driver == null) {
             return false;
         }
-        return "gitleaks".equals(driver.getName());
+        return "gitleaks".equalsIgnoreCase(driver.getName());
     }
 }
--- a/sechub-sereco/src/main/java/com/mercedesbenz/sechub/sereco/importer/GitleaksSarifImportWorkaround.java
+++ b/sechub-sereco/src/main/java/com/mercedesbenz/sechub/sereco/importer/GitleaksSarifImportWorkaround.java
@@ -1,0 +1,37 @@
+// SPDX-License-Identifier: MIT
+package com.mercedesbenz.sechub.sereco.importer;
+
+import org.springframework.stereotype.Component;
+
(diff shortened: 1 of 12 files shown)


sechub-sereco/src/main/java/com/mercedesbenz/sechub/sereco/importer/GitleaksSarifImportWorkaround.java
```diff
@@ -0,0 +1,37 @@
...
+
+    private boolean isGitleaksRun(Run run) {
+        if (run == null) {
+            return false;
+        }
+        if (run.getTool() == null) {
+            return false;
+        }
+        ToolComponent driver = run.getTool().getDriver();
+        if (driver == null) {
```
de-jcup, line 32:
```suggestion
        /* only pseudo class - TheToolClass...*/
        TheToolClass tool = run.getTool()
        if (tool == null) {
            return false;
        }
        ToolComponent driver = tool.getDriver();
        if (driver == null) {
```

I would prefer a local variable here.

sechub-sereco/src/main/java/com/mercedesbenz/sechub/sereco/importer/GitleaksSarifImportWorkaround.java
```diff
@@ -0,0 +1,37 @@
...
+            return false;
+        }
+        return "gitleaks".equals(driver.getName());
```
de-jcup, line 35:
```suggestion
        return "gitleaks".equalsIgnoreCase(driver.getName());
```

sechub-sereco/src/main/java/com/mercedesbenz/sechub/sereco/importer/SarifImportProductWorkaroundSupport.java
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * Resolve type from SARIF rule and SARIF run.
+     *
+     * @param rule
+     * @param run
+     * @return Resolved type or <code>null</code> if type could not be resolved by
+     *         any available workaround.
+     */
+    public String resolveType(ReportingDescriptor rule, Run run) {
+        // iterate over workarounds, first non null result will be returned
```
de-jcup, line 31: Normally we use // for comments to remove - but okay.

I  think we should remove the comment - the javadoc is description enough and the code is also very clear.

de-jcup (approved):
Good job. I found some minor things - see my comments - but approved anyway. Please decide by yourself if you want to change the mentioned parts or not.

PR #2999, opened 2024-03-12
Obfuscate secret scan findings in reports

### LINKED ISSUE #2891: Secret scan `source` shall be obfuscated
## Situation
Inside a SecHub report for a secret scan, the found secret is shown inside source field.

Here a HTML example:

image

Even when a secret shall be treated as leaked when committed to a SCM, there are some reasons why the found sensitive data should not appear inside the report

## Wanted
It shall be possible to obfuscate secret scan source results inside the reports (HTML, JSON).
This shall be done product independently (means for every secret scan).

## Solution
Location: `SerecoProductResultTransformer` convert method 
where code call stack elements are created. Method must be enhanced with boolean flag (obfuscate) + implementation

- We use a trailing replacement like `*****` for hidden parts

- We introduce spring boot property:
`sechub.secretscan.source.visible.length` with default 5

With setting to 0 the result will only be
origin: pwd12345 --> `*****`

With setting to 5 the result will only be
origin: pwd12345 --> `pwd123*****`

With setting to 15 the result will  be
origin: pwd12345 --> `pwd12345`

lorriborri:
- closes #2891

### DIFF
--- a/sechub-commons-core/src/main/java/com/mercedesbenz/sechub/commons/core/util/SimpleStringUtils.java
+++ b/sechub-commons-core/src/main/java/com/mercedesbenz/sechub/commons/core/util/SimpleStringUtils.java
@@ -119,6 +119,14 @@ public class SimpleStringUtils {
             return null;
         }
 
+        if (string.isEmpty()) {
+            return string;
+        }
+
+        if (nonObfuscatedCharacters < 0) {
+            return string;
+        }
+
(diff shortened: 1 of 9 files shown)


de-jcup (changes requested):
Good job, but I still found some issues. Please read my comments

de-jcup (approved):
Looks good to me.

PR #3004, opened 2024-03-14
Enable administrator to pass JVM ARGS to ZAP #3003

### LINKED ISSUE #3003: Introduce options to provide JVM ARGS for ZAP
### Situation
We currently cannot influence the JVM arguments of the ZAP inside the PDS solution. This is a problem if the ZAP is running inside a container/cluster where the resources are limited.

### Wanted
For an administrator it shall be possible to provide JVM arguments for the ZAP process via the executor configuration, like:
`ZAP_JVM_ARGS=-Xmx4096m`

winzj:
- closes #3003


haerter-tss (approved):
Looks good!

PR #3036, opened 2024-03-19
Remote credential model and credential support

### LINKED ISSUE #3025: Prepare phase credential container and support
## Situation
To access remote data, the PDS prepare needs credentials. As there exist different remote platforms (eg. github) it is necessary to store credentials for each platform and select the correct ones for the configured remote location.

## Wanted
A credential container model (containing user, password, remote-pattern and maybe type) and a credential support, that can read the secret credentials and choose the correct ones for the given location. (eg. "location": "github.com/my_example/example_project" -> credentials for github.com)

## Solution
1. Implement credentials-container model in sechub.commons.model, for example:
```json
 {
 "credentials" : [
      {
         "user" : "example-user",
  	   "password" : "example-api-token-or-password",
  	   "remotePattern" : ".*github.com/*",
  	   "type" : "example-type"
  	}
]
}
```

2. Implement CredentialSupport class, that is able to:
-  read the credentials from an ENV (PDS_PREPARE_REMOTE_CREDENTIAL_LIST)
- implements the pattern matching (choose correct credentials) by location and by type

lorriborri:
- closes #3025


sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/model/SecHubRemoteCredentialSupport.java
```diff
@@ -0,0 +1,65 @@
...
+            LOG.debug("Could not match any credentials for configured location {}", location);
+        }
+        return matchedCredentials;
+    }
+
+    public SecHubRemoteCredentialConfiguration getRemoteCredentialConfigurationFromJSONString(String json) {
+        return SecHubRemoteCredentialConfiguration.fromJSONString(json);
+    }
+
+    public String readRemoteCredentialsFromEnv() {
```
de-jcup, line 62: Please remove this method completely - this PDS specific. But we want to have the support (it is in commons) to be useful for any location were we have mutltiple remote credentials configured.

Please adopt the GitHub issue as well, I think the environments are mentioned there also.

sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/model/TypeFilter.java
```diff
@@ -0,0 +1,39 @@
...
+public interface TypeFilter {
+    public static class AcceptAllTypeFilter implements TypeFilter {
+
+        @Override
+        public boolean isTypeAccepted(String type) {
+            return true;
+        }
+    }
+
+    public static class AcceptTypeFilter implements TypeFilter {
```
de-jcup, line 12:
```suggestion
    public static class AcceptGivenCommaSeparatedTypesFilter implements TypeFilter {
```

sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/model/TypeFilter.java
```diff
@@ -0,0 +1,39 @@
...
+
+        private String type;
+
+        public AcceptTypeFilter(String type) {
+            this.type = type;
+        }
+
+        @Override
+        public boolean isTypeAccepted(String type) {
```
de-jcup, line 21:
```suggestion
        public boolean isTypeAccepted(String commaSeparatedTypes) {
```

sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/model/TypeFilter.java
```diff
@@ -0,0 +1,39 @@
...
+            if (type == null) {
+                return false;
+            }
+            // more than one type can be defined eg. docker,git,binary
+            final String[] splitTypes = type.split(",");
+            for (String splitType : splitTypes) {
```
de-jcup, line 27: You do not handle situations where a setup contains whitespaces inside, e.g. "docker, git, binary"

Please add
- trim mechanism
- unit test which has a comma separated list inside which contains whitespaces
```suggestion
            for (String splitType : splitTypes) {
                 String typeToCheck = splitType.trim();
```

sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/model/TypeFilter.java
```diff
@@ -0,0 +1,39 @@
...
+                if (this.type.equals(splitType)) {
+                    return true;
+                }
+            }
+            return false;
+        }
+    }
+
+    public static final AcceptAllTypeFilter ACCEPT_ALL = new AcceptAllTypeFilter();
```
de-jcup, line 36: Please move static parts always on top of main class - here at the interface.

We have always following order

1. fields
2. methods
3. internal classes

And to separated this again:
1. static parts before "normal" parts
4. constructors before other methods
5. public before protected, before package private, before private

sechub-commons-model/src/test/java/com/mercedesbenz/sechub/commons/model/SecHubRemoteCredentialSupportTest.java
```diff
@@ -0,0 +1,181 @@
...
+
+        /* test */
+        List<SecHubRemoteCredentialContainer> credentials = configuration.getCredentials();
+        assertEquals(1, credentials.size());
+
+        SecHubRemoteCredentialContainer container = credentials.get(0);
+        assertNotNull(container);
+        assertEquals("example-user", container.getUser());
+        assertEquals("example-api-token-or-password", container.getPassword());
+        assertEquals(".*github.com/*", container.getRemotePattern());
```
de-jcup, line 41: It is just an example, but it must be clear that we have here regular expression here. So it should be:
```suggestion
        assertEquals(".*github.com/.*", container.getRemotePattern());
```

sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/model/SecHubRemoteCredentialContainer.java
```diff
@@ -0,0 +1,36 @@
+package com.mercedesbenz.sechub.commons.model;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class SecHubRemoteCredentialContainer {
```
de-jcup, line 6: Hmm... "container" means normally something "big" which hold child data etc. and gives the possibility for something like "getChildren()" or any additional methods. Oft containers acts like a list.

Here we have only data for credential situation which itself is always given back inside a list by the support.

I would suggest here a name change:

```suggestion
public class SecHubRemoteCredentialData {
```
or

```suggestion
public class SecHubRemoteCredentialEntry {
```

Means: The SecHubRemoteCredentialConfiguration has "credentials" field which contains data/entry elements.

sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/model/SecHubRemoteCredentialContainer.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public String getUser() {
+        return user;
+    }
+
+    public String getPassword() {
+        return password;
+    }
+
+    public String getRemotePattern() {
```
de-jcup, line 29: Please add javadoc that this is a regular expression

sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/model/SecHubRemoteCredentialSupport.java
```diff
@@ -0,0 +1,65 @@
...
+    private final String PDS_PREPARE_REMOTE_CREDENTIAL_LIST = "PDS_PREPARE_REMOTE_CREDENTIAL_LIST";
+    private static final Logger LOG = LoggerFactory.getLogger(SecHubRemoteCredentialSupport.class);
+
+    public List<SecHubRemoteCredentialContainer> resolveCredentialsForLocation(SecHubRemoteCredentialConfiguration configuration, String location) {
+        return matchRemoteCredentialsFromLocationConfiguration(configuration, location, null);
+    }
+
+    public List<SecHubRemoteCredentialContainer> resolveCredentialsForLocation(SecHubRemoteCredentialConfiguration configuration, String location,
+            String type) {
+        TypeFilter.AcceptTypeFilter acceptTypeFilter = new TypeFilter.AcceptTypeFilter(type);
```
de-jcup, line 21: You can import the subclass in imports directly and use it without the "TypeFilter." prefix. 
It would be more readable and also easier for refactoring (if the class would be moved to its own location)

sechub-commons-model/src/test/java/com/mercedesbenz/sechub/commons/model/SecHubRemoteCredentialSupportTest.java
```diff
@@ -0,0 +1,181 @@
...
+
+    private SecHubRemoteCredentialSupport supportToTest;
+
+    @BeforeEach
+    void beforeEach() {
+        supportToTest = new SecHubRemoteCredentialSupport();
+    }
+
+    @Test
+    void read_simple_credential_configuration_from_json_string() {
```
de-jcup, line 26: Looking into the configuration file and this test at the end, "simple" means here it is just one entry. 
```suggestion
    void read_credential_configuration_with_one_entry_only_from_json_string() {
```

"Simple" is ... not so clear at first sigh - please rename the file and also the method name, so it is clear that there is just one entry inside. Would make it easier to understand.

sechub-commons-model/src/test/java/com/mercedesbenz/sechub/commons/model/SecHubRemoteCredentialSupportTest.java
```diff
@@ -0,0 +1,181 @@
...
+        assertNotNull(container2b);
+        assertEquals("user3", container2b.getUser());
+
+        assertTrue(credentials3.isEmpty());
+    }
+
+    @ParameterizedTest
+    @EmptySource
+    @NullSource
+    void match_remote_credential_pattern_by_type_with_empty_or_null_type(String string) {
```
de-jcup, line 155:
```suggestion
    void match_remote_credential_pattern_by_type_with_empty_or_null_type(String type) {
```

sechub-commons-model/src/test/java/com/mercedesbenz/sechub/commons/model/SecHubRemoteCredentialSupportTest.java
```diff
@@ -0,0 +1,181 @@
...
+        List<SecHubRemoteCredentialContainer> credentials = supportToTest.resolveCredentialsForLocation(configuration, string);
+
+        /* test */
+        assertTrue(credentials.isEmpty());
+    }
+
+    @ParameterizedTest
+    @EmptySource
+    @NullSource
+    void match_remote_credential_pattern_by_location_with_empty_or_null_location(String string) {
```
de-jcup, line 170:
```suggestion
    void match_remote_credential_pattern_by_location_with_empty_or_null_location(String location) {
```

sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/model/SecHubRemoteCredentialSupport.java
```diff
@@ -0,0 +1,65 @@
...
+                matchedCredentials.add(credentialContainer);
+            }
+        }
+        if (matchedCredentials.isEmpty()) {
+            LOG.debug("Could not match any credentials for configured location {}", location);
+        }
+        return matchedCredentials;
+    }
+
+    public SecHubRemoteCredentialConfiguration getRemoteCredentialConfigurationFromJSONString(String json) {
```
de-jcup, line 58: Please drop this method completely.

SecHubRemoteCredentialConfiguration.fromJSONString(json); is enough, we need no additional method which just calls this.

sechub-commons-model/src/test/java/com/mercedesbenz/sechub/commons/model/SecHubRemoteCredentialSupportTest.java
```diff
@@ -0,0 +1,181 @@
...
+        /* prepare */
+        String json = TestFileReader.loadTextFile(new File("./src/test/resources/sechub_remote_credentials_complex_config.json"));
+        SecHubRemoteCredentialConfiguration configuration = supportToTest.getRemoteCredentialConfigurationFromJSONString(json);
```
de-jcup, line 173:
```suggestion
        SecHubRemoteCredentialConfiguration configuration = SecHubRemoteCredentialConfiguration.fromJSONString(json); 
```
A unit test should not need to call the object which is tested to prepare something.

sechub-commons-model/src/test/java/com/mercedesbenz/sechub/commons/model/SecHubRemoteCredentialSupportTest.java
```diff
@@ -0,0 +1,181 @@
...
+        assertNotNull(container2c);
+        assertEquals("userX", container2c.getUser());
+
+        assertTrue(credentials3.isEmpty());
+    }
+
+    @Test
+    void match_remote_credentials_pattern_by_type() {
+        /* prepare */
+        String json = TestFileReader.loadTextFile(new File("./src/test/resources/sechub_remote_credentials_complex_config.json"));
```
de-jcup, line 120: Normally i am a fan of loading test data, but sometimes it can become difficult to understand whats happening inside test.

Sometimes (when data is not too much) it is just better to use the java 17 feature "text blocks"

and do something like this:

```java
String json = """
{
  "credentials": [
    {
      "user" : "example-user",
      "password" : "example-api-token-or-password",
      "remotePattern" : ".*github.com*",
      "type" : "example-type"
    },
    {
      "user" : "user2",
      "password" : "password2",
      "remotePattern" : ".*example*",
      "type" : "git"
    },
    {
      "user" : "user3",
      "password" : "token3",
      "remotePattern" : ".*example.url.com*",
      "type" : "docker,git"
    },
    {
      "user" : "userX",
      "password" : "tokenX",
      "remotePattern" : ".*example.*",
      "type" : "binary"
    }
  ]
}
"""
```

Doing this, some tests could maybe easier to understand

sechub-commons-model/src/test/java/com/mercedesbenz/sechub/commons/model/SecHubRemoteCredentialSupportTest.java
```diff
@@ -0,0 +1,181 @@
...
+        SecHubRemoteCredentialConfiguration configuration = supportToTest.getRemoteCredentialConfigurationFromJSONString(json);
+
+        /* test */
+        assertEquals(0, configuration.getCredentials().size());
+    }
+
+    @Test
+    void read_incomplete_credential_configuration_and_match_location_and_type_where_possible() {
+        /* prepare */
+        String json = TestFileReader.loadTextFile(new File("./src/test/resources/sechub_remote_credentials_incomplete_config.json"));
```
de-jcup, line 60: As mentioned below, here a text block instead of using json would be easier to understand.
The "incomplete" part is not clear.

When using
```java
String json = """
{
  "credentials": [
    {
      "user" : "example-user",
      "password" : "example-api-token-or-password",
      "remotePattern" : ".*github.com.example*",
      "type" : "git"
    },
    {
      "user" : "user2",
      "password" : "password2",
      "type" : "git"
    },
    {
      "user" : "userX",
      "password" : "tokenX",
      "remotePattern" : ".*example*"
      // type is missing here - meaning "incomplete"
    }
  ]
}
The incomplete part comes visible (I think our jackson Json setup accepts comments, so I used it here)
""";
```

sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/model/SecHubRemoteCredentialSupport.java
```diff
@@ -0,0 +1,65 @@
...
+        for (SecHubRemoteCredentialContainer credentialContainer : configuration.getCredentials()) {
+            String stringPattern = credentialContainer.getRemotePattern();
+            String type = credentialContainer.getType();
+
+            if (stringPattern == null) {
+                LOG.debug("Could not match credentials as configured remote pattern was null.");
+                continue;
+            }
+
+            Pattern pattern = Pattern.compile(stringPattern);
```
de-jcup, line 46: Pattern compile can throw an exception.

Please do following:

1. write a unit test with some data containing illegal regular expressions.
   expect no exception here, but the part with the wrong configuration lead to a missing entry. I would expect here a warn log entry
2. change the implementation

sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/model/SecHubRemoteCredentialSupport.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SecHubRemoteCredentialSupport {
```
de-jcup, line 11: I thought about the pattern compile problem (needs time).

I have following idea:

divide the logic:

CreentialDataContainer container = support -> createCredentialDataContainer(SecHubRemoteCredentialConfiguration configuration)
( can be created on startup of a PDS instance for example)

the create method will compiles all patterns inside and create a "container" which
has the patterns inside at runtime.

at runtime the container is used by logic and provides the resolve methods.

Remark: I speak for container this time - means after the renaming. we would have here
- `SecHubRemoteCredentialConfiguration` contains json structure with `SecHubRemoteCredentialData` inside
- `SecHubRemoteCredentialData` (data)
- `SecHubRemoteCredentialSupport` (support which can create the container) - but we could rename it to `SecHubRemoteCredentialDataFactory` as well, becauseit only creates the container...
- `SecHubRemoteCredentialContainer` (resolve methods)

de-jcup (changes requested):
Good job, but I have found many things we should improve

sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/model/RemoteCredentialContainer.java
```diff
@@ -0,0 +1,98 @@
...
+
+    public List<RemoteCredentialData> resolveCredentialsForLocation(String location) {
+        return resolveCredentialsFromLocation(location, null);
+    }
+
+    public List<RemoteCredentialData> resolveCredentialsForLocation(String location, String type) {
+        return resolveCredentialsFromLocation(location, type == null ? null : t -> t.equals(type));
+    }
+
+    private List<RemoteCredentialData> resolveCredentialsFromLocation(String location, TypeFilter filter) {
```
de-jcup, line 31:
```suggestion
    public List<RemoteCredentialData> resolveCredentialsFromLocation(String location, TypeFilter filter) {
```

sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/model/RemoteCredentialContainer.java
```diff
@@ -0,0 +1,98 @@
...
+        this.configuration = configuration;
+        this.patternMap = patternMap;
+    }
+
+    public List<RemoteCredentialData> resolveCredentialsForLocation(String location) {
+        return resolveCredentialsFromLocation(location, null);
+    }
+
+    public List<RemoteCredentialData> resolveCredentialsForLocation(String location, String type) {
+        return resolveCredentialsFromLocation(location, type == null ? null : t -> t.equals(type));
```
de-jcup, line 28:
```suggestion
        return resolveCredentialsFromLocation(location, type == null ? null : t -> type.equals(t));
```

sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/model/RemoteCredentialContainer.java
```diff
@@ -0,0 +1,98 @@
...
+         */
+        if (TypeFilter.ACCEPT_ALL.equals(filter)) {
+            typeAccepted = true;
+        }
+        for (String type : types) {
+            typeAccepted = filter.isTypeAccepted(type);
+            if (typeAccepted) {
+                break;
+            }
+        }
```
de-jcup, line 69:
```suggestion
        if (types.isEpty()){
          typeAccepted = filter.isTypeAccepted(null);
        }else {
       
          for (String type : types) {
              typeAccepted = filter.isTypeAccepted(type);
              if (typeAccepted) {
                  break;
              }
          }
        }
```

sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/model/RemoteCredentialContainerFactory.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.regex.PatternSyntaxException;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class RemoteCredentialContainerFactory {
+
+    private static final Logger LOG = LoggerFactory.getLogger(RemoteCredentialContainerFactory.class);
+
+    public RemoteCredentialContainer create(RemoteCredentialConfiguration configuration) {
```
de-jcup, line 15: Please add javadoc and describe with @throws IllegalStateException ....

sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/model/RemoteCredentialData.java
```diff
@@ -0,0 +1,43 @@
...
+
+    public String getPassword() {
+        return password;
+    }
+
+    /**
+     *
+     * @return regular expression for the remote location
+     */
+    public String getRemotePattern() {
```
de-jcup, line 36:
```suggestion
    /**
     * Returns the remote location pattern - which must be a valid regular expression
     * @return pattern
     */
    public String getRemotePattern() {
```

de-jcup (approved):
Good job! 

I found some minor parts, but it's up to you if you want to add those changes as well or not.

PR #3045, opened 2024-03-20
Server Helm Chart 2.0.0 #3042

### LINKED ISSUE #3042: SecHub Server Helm chart: make env vars configurable
## Situation
Currently there is no option to set environment variables for containers directly.
Instead there is a construct like this in the deployment.yaml:
```yaml
{{- if .Values.pdsGoSec.enabled }}
# ----------------------------------------------------------------------------------------------#
# - Scope: PDS-GoSec
# --------------------------------------------------------------------------------------------- #
        - name: SECHUB_PDS_GOSEC_USERID
          valueFrom:
            secretKeyRef:
               name: secret-pds-gosec
               key: techuser_username
        - name: SECHUB_PDS_GOSEC_PASSWORD
          valueFrom:
            secretKeyRef:
               name: secret-pds-gosec
               key: techuser_password
{{- end }}
```
If someone creates their own PDS solution and wants to set the secrets via environment variables, then currently this is not possible.

## Wanted
A possibility to define envrionment variables directly via the values.yaml

## Solution
Change the Helm chart so it's possible to define environment vars directly in the values.yaml.
Remove the `{{- if .Values.pdsXYZ.enabled }}` construct.

### LINKED ISSUE #3044: SecHub Server Helm chart: make secrets obfuscation configurable
## Situation
With #2891 now secrets are obfuscated completely by `*****` per default.
But with `sechub.report.sensitivedata.max.nonobfuscated.characters` some of the first characters con be displayed.
(See https://mercedes-benz.github.io/sechub/latest/sechub-techdoc.html)

## Wanted
This parameter shall be configurable within the SecHub Server Helm chart.

## ToDo
Add above to the Helm chart (via environment variable).

sven-dmlr:
- closes #3042
- closes #3044


haerter-tss (approved):
Looks good!

PR #3050, opened 2024-03-22
Feature 3049 add custom rule file to pds gitleaks

### LINKED ISSUE #3049: Add initial file with custom rules to to gitleaks pds solution
### Wanted
We want to add a file like `custom-gitleaks.toml` to the gitleaks PDS solution, to add rules or overwrite default rules.

#### Edit
There is a bug inside gitleaks, where only the custom rules are well defined inside the SARIF report. The default rules are missing:
image

This means for now we have to specify all rules and simply overwrite or add the ones needed, otherwise we always get the SecHub finding type `Undefined` (see image).
This was already reported to gitleaks: https://github.com/gitleaks/gitleaks/issues/1299

### Out of scope for this issue
In the future we want to be able to apply/load the custom rule file during runtime, e.g via Helm Chart.

winzj:
- closes #3049


sven-dmlr (approved):
Looks good to me :+1:

PR #3052, opened 2024-03-26
FindSecurityBugs works with java 21 #2773

### LINKED ISSUE #2773: FindSecurityBugs encounters issues with java 21
## Situation:
When scanning binaries that use Temurin / Java 21 FindSecurityBugs will fail.

## Solution:
The currently used Spotbugs version does not support Temurin / Java 21. Starting from version 4.8.0 they do. That means that Spotbugs needs to be updated and the system version of the openjdk needs to be updated to 21 too.

sven-dmlr:
currently this is a draft PR :-)
- closes #2773


haerter-tss (approved):
Looks good!

Issue #2986, opened 2024-03-07
Implement PDS prepare wrapper application


lorriborri:
## Situation
To download configured remote sources during the prepare phase, a wrapper application is required. 

## Wanted
Application that implements the prepare phase. The wrapper should recognize different remote types and select the correct module (eg. git, skopeo for docker). Further the application needs to access the remote credentials and hand them over to the modules. 

## Solution
Implement prepare application that can handle different remote sources and binaries. 
Maybe  the commons-core ArchiveSupport needs to be extended for zipping and extracting the defined sources.

The wrapper application should:
- assert if remote data is defined/ available (or maybe validator for model) 
- implement PrepareModuleService (Logic in Modules #3027 and #3028)
- notify sechub server if PREPARE was successful or not

Issue #3028, opened 2024-03-18
Implement PrepareModule for Skopeo


lorriborri:
## Situation
The PDS prepare-wrapper needs to execute skopeo commands to download remote sources from a docker repository. This should be implemented as a skopeo module.

## Wanted
Skopeo PrepareModule that implements downloading data from a docker repository.

## Solution
Implement PrepareModule with necessary logic to download docker images.

Issue #3027, opened 2024-03-18
Implement PrepareModule for GIT


lorriborri:
## Situation
The PDS prepare-wrapper needs to execute git or Jgit to download remote sources from a git repository. This should be implemented as a git module.

## Wanted
Git PrepareModule that can implements downloading (cloning) from a git repository.

## Solution
Implement PrepareModule with necessary logic to clone git repositories.

PR #3053, opened 2024-03-26
Prepare-wrapper-application

lorriborri:
- closes: #2986


sechub-wrapper-prepare/src/main/java/com/mercedesbenz/sechub/wrapper/prepare/prepare/PrepareWrapperPreparationService.java
```diff
@@ -0,0 +1,68 @@
...
+    @Autowired
+    PrepareWrapperResultStatus status;
+
+    public AdapterExecutionResult startPreparation() {
+        LOG.debug("Start preparation");
+        PrepareWrapperContext context = factory.create(environment);
+        List<SecHubRemoteDataConfiguration> remoteDataConfigurationList = getRemoteConfigurationsFromSecHubModel(context);
+        if (remoteDataConfigurationList.isEmpty()) {
+            // TODO: 27.03.24 laura is it ok that noe remote data was configured?
+            LOG.debug("No Remote configuration was found");
```
de-jcup, line 36:
```suggestion
           LOG.warn("No Remote configuration entries were found for this PDS instance!");
```

sechub-wrapper-prepare/src/main/java/com/mercedesbenz/sechub/wrapper/prepare/prepare/PrepareWrapperPreparationService.java
```diff
@@ -0,0 +1,68 @@
...
+
+    private List<SecHubRemoteDataConfiguration> getRemoteConfigurationsFromSecHubModel(PrepareWrapperContext context) {
+        List<SecHubRemoteDataConfiguration> result = new ArrayList<>();
+        if (context.getSecHubConfiguration() == null || context.getRemoteCredentialContainer() == null) {
+            throw new IllegalStateException("Context was not initialized correctly. SecHub or credential configuration was null");
+        }
+        if (context.getSecHubConfiguration().getData().isPresent()) {
+            List<SecHubSourceDataConfiguration> sourceDataList = context.getSecHubConfiguration().getData().get().getSources();
+            List<SecHubBinaryDataConfiguration> binaryDataList = context.getSecHubConfiguration().getData().get().getBinaries();
+
```
de-jcup, line 52:
```suggestion
        var dataOpt = context.getSecHubConfiguration().getData();
        if (dataOpt.isPresent()) {
            var data = dataOpt.get();
            List<SecHubSourceDataConfiguration> sourceDataList = data.getSources();
            List<SecHubBinaryDataConfiguration> binaryDataList = data.getBinaries();

```

sechub-wrapper-prepare/src/main/java/com/mercedesbenz/sechub/wrapper/prepare/prepare/PrepareWrapperPreparationService.java
```diff
@@ -0,0 +1,68 @@
...
+            for (SecHubSourceDataConfiguration sourceData : sourceDataList) {
+                if (sourceData.getRemote().isEmpty()) {
+                    continue;
+                }
+                result.add(sourceData.getRemote().get());
```
de-jcup, line 57:
```suggestion
                var remotOpt = sourceData.getRemote();
                if (remotOpt.isPresent()) {
                      result.add(remotOpt.get());
                }

```

sechub-wrapper-prepare/src/main/java/com/mercedesbenz/sechub/wrapper/prepare/prepare/PrepareWrapperPreparationService.java
```diff
@@ -0,0 +1,68 @@
...
+            }
+            for (SecHubBinaryDataConfiguration binaryData : binaryDataList) {
+                if (binaryData.getRemote().isEmpty()) {
```
de-jcup, line 60: same as above

sechub-wrapper-prepare/src/main/java/com/mercedesbenz/sechub/wrapper/prepare/prepare/PrepareWrapperPreparationService.java
```diff
@@ -0,0 +1,68 @@
...
+        LOG.debug("Start preparation");
+        PrepareWrapperContext context = factory.create(environment);
+        List<SecHubRemoteDataConfiguration> remoteDataConfigurationList = getRemoteConfigurationsFromSecHubModel(context);
+        if (remoteDataConfigurationList.isEmpty()) {
+            // TODO: 27.03.24 laura is it ok that noe remote data was configured?
+            LOG.debug("No Remote configuration was found");
+            return new AdapterExecutionResult(status.getSTATUS_OK());
+        }
+
+        // TODO: 26.03.24 laura start GIT and SKOPEO Services with model
```
de-jcup, line 40: Missing: The complete launch mechanism is missing.

I would expect here a something like

```java
@Autowired
List<PreparWrapperModule> modules = new ArrayList<>();

//...

for (PreparWrapperModule module: modules){
    var sechubConfiguration = context.getModel();
    if (module.isAbleToPrepare(sechubConfiguration){
        module.prepare(sechubConfiguration, remoteDataConfigurationList);
    }
}

```
The implementations could be done later, but currently
https://github.com/mercedes-benz/sechub/issues/2986 describes that the 
https://github.com/mercedes-benz/sechub/issues/3027 and https://github.com/mercedes-benz/sechub/issues/3028 should be done inside (but PR does only close 2986)

Remark:
Maybe it would be good to move the remoteDataConfigurationList calculation to a `PrepareWraperRemoteDataConfigurationService` which does this at startup and the modules just could have it injected - could keep the interface smaller...

sechub-wrapper-prepare/build.gradle
```diff
@@ -0,0 +1,47 @@
...
+
+version = versionData.getPrepareWrapperVersion()
+
+bootJar {
+
+    doLast {
+        /* Here we copy the created wrapper jar as "sechub-wrapper-prepare.jar" into the
+         * pds-tools folder inside the sechub-integrationtest gradle subproject.
+         */
+        File integratonTestLaunchFile = new File("${rootProject.projectDir}/sechub-integrationtest/build/pds-tools/sechub-wrapper-prepare.jar")
```
de-jcup, line 35: Okay, but the purpose should be clear: This is copied from checkmarx wrapper parts and the reason why this was done is, that the checkmarx wrapper is integration tested!

When we do this here as well, we could write an integration test for the prepare wrapper as well!

How?

First we write a special class:

```java
@Component
@Profile(PrepareWrapperModuleProfiles.INTEGRATIONTEST)
public class IntegrationTestPrepareWrapperModule implements PrepareWrapperModule

   @Autowired
   StorageService storageService;

   public boolean isAbleToPrepare(...){
     return true;
   }

   public void prepare(SecHubConfiguration model, ... more) throws IOException{
      String someOutputData = """ 
         ... some test data which will be checked inside integration test
         maybe with additional information - e.g. a variable output for the 
         used test credentials etc
      """;
      storageService.write(someOutputData);
   } 
```

Add the PDS wrapper script from the PDS-solution into the integration test PDS configuration file.

After this we write a "normal" SecHub integration test (with a new scenario + profile were a prepare wrapper is configured;

Last but not least: write an integration test which uses the new profile + checks the expected output data

de-jcup (changes requested):
Good job, but I found some parts.

de-jcup (approved):
Well done. 

Remark: I found some small naming issues inside tests - see my comments. But I approved it anyway.

PR #3065, opened 2024-04-10
client accepts remote-data-section #2976

### LINKED ISSUE #2976: Sechub client extend scan preperations for remote data
## Situation
The SecHub client can not deal with the newly introduced remote data section. The client automatically tries to create a zip from the sources or get the tar from the binary if a source or binaries section is defined.

Example remote section for binaries:
```json
{
  "apiVersion": "1.0",
  "data": {
    "binaries": [
      {
        "name": "remote_example_name",
        "remote": {
          "location": "remote_example_location",
          "type": "docker"
        }
      }
    ]
  },
  "codeScan": {
    "use": [
      "remote_example_name"
    ]
  }
}
```

Example remote section for sources:
```json
{
  "apiVersion": "1.0",
  "data": {
    "sources": [
      {
        "name": "git_remote_example_name",
        "remote": {
          "location": "git_remote_example_location",
          "type": "git"
        }
      }
    ]
  },
  "codeScan": {
    "use": [ "git_remote_example_name" ]
  }
}
```

## Wanted
The Client should be able to detect when a remote scan is present. If a remote scan is present, the client does not need to create a zip or binary file for upload. 

## Solution
Integrate remote section behavior in the client prepareScan function.

sven-dmlr:
- closes #2976

### DIFF
--- a/sechub-cli/src/mercedes-benz.com/sechub/cli/context.go
+++ b/sechub-cli/src/mercedes-benz.com/sechub/cli/context.go
@@ -15,20 +15,14 @@ type Context struct {
 	contentToSend             []byte // template output used for communication to server - contains replaced parts from env variables (e.g. password instead of variable)
 	HTTPClient                *http.Client
 	sechubConfig              *SecHubConfig
-	binariesTarFileName       string
 	binariesTarFileChecksum   string
+	binariesTarFileName       string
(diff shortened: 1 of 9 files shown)


lorriborri (approved):
God Job! Looks god to me :star2:

PR #3084, opened 2024-04-22
Feature 3078 client secrets scan switches off filtering

### LINKED ISSUE #3078: Client: On secrets scan: switch off source code filter automatically
## Situation
The SecHub client per default add only source code files which is fine for SAST code scan.
See https://mercedes-benz.github.io/sechub/latest/sechub-client.html#code-scan-accepted-source-code-files 

Since some time SecHub is able to scan for secrets.
For this almost every file should be uploaded for scan.

Currently you have to switch off the source code filters like this:
```bash
export SECHUB_WHITELIST_ALL="true"
export SECHUB_IGNORE_DEFAULT_EXCLUDES="true"
```
and define excludes for everything that is not wanted in the upload like:
```json
{
  "apiVersion": "1.0",
  "secretScan": { "use": [ "src" ] },
  "data": {
    "sources": [
      {
        "name": "src",
        "fileSystem": { "folders": [ "." ] },
        "excludes": [
          "**/node_modules/**",
          "sechub_report_*.json",
          "sechub-false-positives*.json",
          "sourcecode-*.zip"
        ]
      }
    ]
  }
}
```

## Wanted
The SecHub client parses the `sechub.json` and
identifies all "sources" that are used for secret scan
in these cases:
- switch off the source code filter on secrets scans
- automatically filter out SecHub files, libraries, IDE files, images

Documentation :-)

sven-dmlr:
- closes #3078

### DIFF
--- a/sechub-cli/src/mercedes-benz.com/sechub/cli/constants.go
+++ b/sechub-cli/src/mercedes-benz.com/sechub/cli/constants.go
@@ -46,11 +46,29 @@ const MinimalTimeoutInSeconds = 10
 // SizeOfJobList - Number of latest jobs to print
 const SizeOfJobList = 20
 
-// DefaultSourceCodeExcludeDirPatterns - Define directory patterns to exclude from zip file:
-// - code in directories named "test" is not considered to end up in the binary
-// - also ignore ".git" directory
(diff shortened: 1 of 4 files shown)


lorriborri (approved):
Good Work :star2: looks all fine to me

PR #3097, opened 2024-04-24
Add custom rules and update Dockerfile #3056

### LINKED ISSUE #3056: Add additional rules to PDS_GITLEAKS
Add custom rules to the PDS_GITLEAKS solution.

winzj:
- closes #3056


sechub-pds-solutions/gitleaks/docker/Gitleaks.dockerfile
```diff
@@ -5,7 +5,7 @@ ARG BASE_IMAGE
 
 FROM ${BASE_IMAGE}
 
-ARG GITLEAKS_VERSION=8.16.4
+ARG GITLEAKS_VERSION=8.18.2
```
sven-dmlr, line 8: This is just a fallback.
In the image build the real value is taken from `sechub-pds-solutions/gitleaks/env`.
There the version is already 8.18.2
We could think about leaving it empty here(?)

sven-dmlr (approved):
Looks good to me :+1: 

Please see my comment.

PR #3132, opened 2024-05-17
Enlarge project_id database field #3102

### LINKED ISSUE #3102: Enlarge project_id database varchar field
### Situation
We had to enlarge this field in past, because it was to short for our needs.
There might be some use cases `project_id` should be larger than the currently allowed 40 characters.

#### Example
For Example mapping a Github repository name as a SecHub `project_id` or even contain the repository name and the Github organization. Github for example allows:
- organization name: 39 characters
- repository name: 100 characters

This means somehting like `$ORGNAME_$REPONAME` could be to large for the current 40 characters

### Solution
- Find a good length that is sufficient, like for example `255` characters
- Enlarge the  `project_id` varchar database field.

winzj:
- closes #3102

### DIFF
--- a/sechub-server/src/main/resources/db/migration/V28__enlarge_project_id.sql
+++ b/sechub-server/src/main/resources/db/migration/V28__enlarge_project_id.sql
@@ -1,0 +1,22 @@
+-- SPDX-License-Identifier: MIT
+
+-- Update project id : we now accept 765 (3 x 255) see ProjectIdValidation
+ALTER TABLE adm_project ALTER COLUMN project_id TYPE varchar(765);
+ALTER TABLE scan_report ALTER COLUMN project_id TYPE varchar(765);
+ALTER TABLE scan_access ALTER COLUMN project_id TYPE varchar(765);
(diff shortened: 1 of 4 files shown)


sechub-server/src/main/resources/db/migration/V28__enlarge_project_id.sql
```diff
@@ -0,0 +1,22 @@
+-- SPDX-License-Identifier: MIT
+
+-- Update project id : we now accept 765 (3 x 255) see ProjectIdValidation
```
sven-dmlr, line 3:
```suggestion
-- Update project id : we now accept 255 characters see ProjectIdValidation
```
There is no need to triple the value.
When the maximum is 255, then a `varchar(255)` value is sufficient.
(I have tested this successfully with Postgres and H2/Integrationtests)

sven-dmlr (changes requested):
Looks good overall.

Please check my comment.

We should also extend the integration tests with the real client because it may have its own limits which could prevent large project IDs.
-> separate issue

sven-dmlr (approved):
:+1:

PR #3134, opened 2024-05-17
Added validation for remote data configuration

### LINKED ISSUE #3133: Improve remote Data validator
## Situation
1. The Sechub model validator allows multiple remote sections that can be configured and remote and file system can be configured both at once.

2. The sechub validator does not check the remote data section for correct configuration.

## Wanted

1. Check only one configuration
- When using a remote data section it is only possible to define ONE binary or ONE source definition. Means also: It is only possible to define ONE remote data section (see example 1 for wrong configuration).

- When using a remote data section it is not possible to define any file system (see example 2 for wrong configuration).

2. Check if remote configuration is correct
- make sure location is not null or blank

- make sure credentials are configured correctly or not defined 
 
### Details
example 1: duplicated remote sources (forbidden)
```json
{
  "apiVersion": "1.0",
  "data": {
    "sources": [
      {
        "name": "remote_example_name",
        "remote": {
          "location": "remote_example_location",
          "type": "git",
          "credentials": {
            "user": {
              "name": "my-example-user",
              "password": "my-example-password"
            }
          }
        }
      },
      {
       "name": "remote_example_name_two",
       "remote": {
         "location": "remote_example_location",
         "type": "some-other-type",
         "credentials": {
           "user": {
             "name": "my-example-user",
             "password": "my-example-password"
           }
         }
       }
      }
    ]
  },
  "codeScan": {
    "use": [
      "remote_example_name"
    ]
  }
}
```
example 2: mix remote and file system (forbidden)
```json
{
  "apiVersion": "1.0",
  "data": {
    "sources": [
      {
        "name": "remote_example_name",
        "remote": {
          "location": "remote_example_location",
          "type": "git",
          "credentials": {
            "user": {
              "name": "my-example-user",
              "password": "my-example-password"
            }
          }
        }
      },
      {
      "name": "filesystem_example",
        "fileSystem" : {
          "folders" : [ "myProject/build" ]
        }
      }
    ]
  },
  "codeScan": {
    "use": [
      "remote_example_name"
    ]
  }
}
```
## Solution
Implement Validation in `SechubConfigurationModelValidator`

lorriborri:
- closes #3133

### DIFF
--- a/sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/model/SecHubConfigurationModelValidationError.java
+++ b/sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/model/SecHubConfigurationModelValidationError.java
@@ -74,7 +74,7 @@ public enum SecHubConfigurationModelValidationError {
 
     REMOTE_DATA_CONFIGURATION_USER_NAME_NOT_DEFINED("Remote data configuration credentials: user name is not defined."),
 
-    REMOTE_DATA_CONFIGURATION_USER_PASSWORD_NOT_DEFINED("Remote data configuration credentials: user password is not defined."),;
+    REMOTE_DATA_CONFIGURATION_USER_PASSWORD_NOT_DEFINED("Remote data configuration credentials: user password is not defined.");
 
     private String defaultMessage;
 
--- a/sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/model/SecHubConfigurationModelValidationError.java
+++ b/sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/model/SecHubConfigurationModelValidationError.java
@@ -66,7 +66,15 @@ public enum SecHubConfigurationModelValidationError {
 
     METADATA_LABEL_KEY_CONTAINS_ILLEGAL_CHARACTERS("Meta data label key contains illegal characters."),
 
-    ;
+    REMOTE_DATA_MULTI_CONFIGURATION_NOT_ALLOWED("Remote data configuration is not allowed for multiple data configurations."),
+
+    REMOTE_DATA_CONFIGURATION_LOCATION_NOT_DEFINED("Remote data configuration location is not defined."),
+
+    REMOTE_DATA_CONFIGURATION_USER_NOT_DEFINED("Remote data configuration credentials: no user is defined."),
+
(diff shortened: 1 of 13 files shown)


sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/model/SecHubConfigurationModelValidator.java
```diff
@@ -552,6 +554,56 @@ private void validateNameUniqueAndNotNull(InternalValidationContext context, Col
...
+
+            if (optRemoteData.isEmpty()) {
+                // no remote data is configured
+                return;
+            }
+
+            String uniqueName = configurationObject.getUniqueName();
+
+            if (configurationObjects.size() > 1) {
+                // remote data is configured with multiple configurations (filesystem or second
```
de-jcup, line 572: I find this misleading.

Not the remote data is configured with multiple configurations - but we have multiple sources or binaries (configuration objects) and at least one is a remote configuration.

Means if one configurationObject (or "sourceOrBinary") is a remote definition and we have multiple "sourceOrBinary" objects its currently a failure

It is currently not clear if we allow:
a) if one remote defined, no filesystem may be defined
b) if one remote defined, not any other remote maybe defined 
  b1) everywhere for any other source / binary (but handled separatley) (as implemented currently)
  b2) everywhere for any other source + binary (means not two calls but only one call necessary)
c) does the unique id matter in such cases?

sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/model/SecHubConfigurationModelValidationError.java
```diff
@@ -66,7 +66,15 @@ public enum SecHubConfigurationModelValidationError {
 
     METADATA_LABEL_KEY_CONTAINS_ILLEGAL_CHARACTERS("Meta data label key contains illegal characters."),
 
-    ;
+    REMOTE_DATA_MULTI_CONFIGURATION_NOT_ALLOWED("Remote data configuration is not allowed for multiple data configurations."),
```
de-jcup, line 69: Sorry but the text is a little bit hard to understand (at least two parts are tested: remote only once (per unique id?) and no filesystem with remote combined possible

de-jcup (changes requested):
Good job, but there are some parts I want to clarify. Most of all the definition how the multiple configuration is defined and how to handle it + error message.

de-jcup (approved):
Well done

PR #3144, opened 2024-05-23
Introduce PDS environment cleanup for caller scripts #3142

### LINKED ISSUE #3142: Reduce visibility of deployment variables inside PDS caller scripts
## Situation
`PDSExecutionCallable` does currently only add calculated parts to the process environment of the PDS caller script and executes it.

But... the scripts have access to any environment variable of the PDS server.
This is a security flaw.

## Wanted
The scripts shall have no access to sensitive data.

## Solution
- we clear everything from environment map except
  - standard whitelisted environment variables
    - `HOME`
    - `HOSTNAME`
    - `PATH`
    - `PWD`
    - `TERM`
    - `UID`
    - `USER`
  - values inside `PDS_SCRIPT_ENV_WHITELIST` 
    - it contains comma separated values
    - it is possible to define explicit accepted environment variable names
    - it is also possible to use asterisk for accepted variable names - e.g. `PDS_STORAGE_*` would accept `PDS_STORAGE_S3_SOMETHING`, `PDS_STORAGE_S3_OTHER` etc. as white listed (just for convenience when a group of variables shall be accepted)

- we add calculated parts

For some special PDS solutions like `prepare-solution` which need special parts must add those parts to their whitelist. So the responsibility is not inside the script but at `PDS` side and secure per default

de-jcup:
This PR

- [x]  closes #3142


sechub-pds/src/test/java/com/mercedesbenz/sechub/pds/execution/PDSScriptEnvironmentCleanerTest.java
```diff
@@ -0,0 +1,182 @@
...
+        assertEquals(null, environment.get(key3));
+
+    }
+
+    @Test
+    void two_keys_which_are_in_environment_and_are_whitelisted_with_asterisk_will_not_be_removed_but_others() {
+
+        /* prepare */
+        String key1 = "PDS_STORAGE_S3_1";
+        String key2 = "PDS_STORAGE_S3_1";
```
winzj, line 119: This is a typo I think, because `key1` and `key2` are currently the same.
```suggestion
        String key2 = "PDS_STORAGE_S3_2";
```

winzj (approved):
Looking good 👍 

I just found a small issue, please have a look at my comment.

Issue #3143, opened 2024-05-23
Ignore whitespaces when parsing github action scan types


hamidonos:
Situation

When passing scan types as comma separated list with whitepaces the scan type is not correctly recognized because some scan types contain the whitespace

Wanted

scan-types: 'codeScan, secretScan' is parsed as ['codeScan', 'secretScan']

Solution

Sanitize the comma separated string

PR #3152, opened 2024-05-28
sanitize github actions input strings #3143

hamidonos:
- closes:  #3143
 
note: The `createSafeBuilderData` method of the `launcher.ts` component is a candidate for testing, but since it's not exposed as a public function it's not accessible from outside. Testing it through the `launch` method call would be complex for the sake of this pull request. 
We might consider to refactor the `launcher.ts` component in the future.


sven-dmlr (approved):
Looks good to me :+1:

PR #3156, opened 2024-05-29
Feature2835 selectablekicsversion

### LINKED ISSUE #2835: KICS version to be installed should be customizable via environment variable
### Situation:
Currently the PDS-IaC implementation will pull the latest version of KICS from their git when the docker image is built. 

### Wanted:
It should be possible to define an environment variable ($KICS_VERSION) with which you can customize which version of KICS should be installed in the PDS.

### LINKED ISSUE #3155: Adjust CWE workaround for SARIF reports in light of KICS 1.7.13 update
## Situation
In previous versions of KICS (prior to 1.7.12), SARIF reports did not include Common Weakness Enumeration (CWE) identifiers for findings. To address this, a workaround was implemented in SecHub that adds a default CWE to all findings to ensure that reports were compliant with expected standards.

## Wanted
With the release of KICS version 1.7.12, some CWEs are now included for Docker findings. However, the current workaround in SecHub indiscriminately adds a default CWE to all findings, which can lead to inaccuracies in the reports where KICS already provides a CWE.

## Solution
The workaround needs to be refined to check if a CWE is already present in the SARIF report for a finding. If a CWE is present, the workaround should not add a default CWE. This adjustment will ensure that the SARIF reports contain accurate CWE information when available from KICS, and fall back to the default CWE only when necessary.

haerter-tss:
## Overview
This pull request introduces two significant improvements to the PDS-IaC implementation. It addresses the need for customizable KICS version installations and refines the CWE handling in SARIF reports post KICS 1.7.13 update.

### Changes
- **Customizable KICS Version (Closes #2835):**
  - Implemented support for a `$KICS_VERSION` environment variable to specify the version of KICS to be installed during the Docker image build process.
  - The version number is included in the build process so it will later show when queried

- **Refined SARIF CWE Handling (Closes #3155):**
  - Modified the SecHub workaround to conditionally add a default CWE identifier in SARIF reports only when KICS does not provide one.


sechub-pds-solutions/iac/09-compute-image-tag.sh
```diff
@@ -12,6 +12,10 @@ else
   VERSION_TAG="`date +%Y-%m-%d`"
 fi
 
+if [[ -n "$KICS_VERSION" ]] ; then
```
sven-dmlr, line 15: Suggest to leave out this section because in future the container may contain other tools than Kics.

sechub-pds-solutions/iac/10-create-image.sh
```diff
@@ -44,9 +44,9 @@ fi
 BUILD_ARGS="--build-arg BASE_IMAGE=$BASE_IMAGE"
 echo ">> Base image: $BASE_IMAGE"
 
-if [[ ! -z "$IAC_VERSION" ]] ; then
-    echo ">> IaC version: $IAC_VERSION"
-    BUILD_ARGS="$BUILD_ARGS --build-arg IAC_VERSION=$IAC_VERSION"
+if [ -n "$KICS_VERSION" ] ; then
```
sven-dmlr, line 47: Suggestion: `source ./env` if $KICS_VERSION is empty so the version is always set.

sechub-pds-solutions/iac/docker/IaC-Debian.dockerfile
```diff
@@ -22,21 +23,22 @@ FROM ${BASE_IMAGE} AS builder
 # Build args
 ARG GO
 ARG PDS_ARTIFACT_FOLDER
-
 ARG BUILD_FOLDER="/build"
-ARG GIT_URL_KICS="https://github.com/Checkmarx/kics.git"
-ARG GIT_BRANCH_KICS="master"
+ARG KICS_VERSION="2.0.1"
```
sven-dmlr, line 27: KICS_VERSION should be inherited from the base container (line 9)
```suggestion
ARG KICS_VERSION
```

sven-dmlr (changes requested):
Look good!
Please check my comments for some suggestions.

sven-dmlr (approved):
Looks good! :+1:

PR #3179, opened 2024-05-31
fixed licenseScan having an empty zip file #3178

### LINKED ISSUE #3178: Client: Empty .zip file with type "licenseScan"
## Situation
When doing a licenseScan, then the following message appears:
```
...
ERROR: Zipfile has no content! Please check your "sources" section in the config file.
Exiting due to fatal error while creating sources zip file...
```

Example SecHub JSON configfile:
```json
{
  "apiVersion": "1.0",
  "data": {
    "sources": [
      {
        "name": "reference-data-1",
        "fileSystem": { "folders": [ "." ] }
      }
    ]
  },
  "licenseScan": { "use": [ "reference-data-1" ] }
}
```
Among others this makes the integration test of `github-actions/scan` fail.

## ToDo
- On scan type "licenseScan" use the source code patterns for upload (same as "codeScan")
- Add a unit test for "licenseScan"

sven-dmlr:
- closes #3178

### DIFF
--- a/sechub-cli/src/mercedes-benz.com/sechub/cli/sechubconfig.go
+++ b/sechub-cli/src/mercedes-benz.com/sechub/cli/sechubconfig.go
@@ -171,12 +171,9 @@ func adjustSourceFilterPatterns(context *Context) {
 			// Clear upload filter for secrets scans
 			context.sechubConfig.Data.Sources[i].SourceCodePatterns =
 				adjustSourceFilterPatternsWhitelistAll(item.SourceCodePatterns, true)
-		} else if slices.Contains(context.sechubConfig.CodeScan.Use, item.Name) {
+		} else if slices.Contains(context.sechubConfig.CodeScan.Use, item.Name) ||
+			slices.Contains(context.sechubConfig.LicenseScan.Use, item.Name) {
 			// Append default source code patterns for code scans
-			context.sechubConfig.Data.Sources[i].SourceCodePatterns =
-				adjustSourceFilterPatternsWhitelistAll(item.SourceCodePatterns, context.config.whitelistAll)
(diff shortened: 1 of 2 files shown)


sechub-cli/src/mercedes-benz.com/sechub/cli/sechubconfig.go
```diff
@@ -162,13 +168,17 @@ func adjustSourceFilterPatterns(context *Context) {
...
 		if slices.Contains(context.sechubConfig.SecretScan.Use, item.Name) {
-			// Clear all source code patterns for secrets scans
+			// Clear upload filter for secrets scans
 			context.sechubConfig.Data.Sources[i].SourceCodePatterns =
 				adjustSourceFilterPatternsWhitelistAll(item.SourceCodePatterns, true)
 		} else if slices.Contains(context.sechubConfig.CodeScan.Use, item.Name) {
 			// Append default source code patterns for code scans
 			context.sechubConfig.Data.Sources[i].SourceCodePatterns =
 				adjustSourceFilterPatternsWhitelistAll(item.SourceCodePatterns, context.config.whitelistAll)
+		} else if slices.Contains(context.sechubConfig.LicenseScan.Use, item.Name) {
```
hamidonos, line 178: wouldn't it be better to combine this into the previous "else-if" statement
the executed code is the same

sven-dmlr: For now we can do this. Right.
But in future there may be a `adjustLicensefileFilterPatternsWhitelistAll()` function.
Currently, SecHub's only license scan tool is PDS-Xray which consumes only binary files.
In future there may be integrated source code license scanners. Their file pattern may differ.

But anyway: Currently there is no reason for the extra "else if".
-> I'll change this into one "else if" statement.

PR #3185, opened 2024-05-31
use 'latest' version as default for GitHub actions #3101

### LINKED ISSUE #3101: Provide "latest" for client version in GitHub action and use as default
## Situation
Currently the GitHub action default is a dedicated version (e.g. '1.4.0').
It's also mentioned inside 
https://github.com/mercedes-benz/sechub/blob/develop/github-actions/scan/README.adoc

The problem is, that client version updates are not automatically used by the SecHub GitHub action

## Wanted
The default shall be `latest`

## Solution

- [ ] change download implementation
- [ ] change default value to latest
- [ ] change README with example to use latest

### Details: change download implementation
https://github.com/mercedes-benz/sechub/blob/develop/github-actions/scan/src/client-download.ts

must be changed in a way, that when client version is `lastest`, the
```typescript
 const zipDownloadUrl = `https://github.com/mercedes-benz/sechub/releases/download/v${clientVersion}-client/sechub-cli-${clientVersion}.zip`;
```
uses instead (in this case)
```typescript
 zipDownloadUrl = `https://mercedes-benz.github.io/sechub/latest/sechub-client.html`;
```
### Details: change README
`latest` must be the example entry - means people just copying the example will automatically use the latest client version

hamidonos:
- closes #3101


github-actions/scan/package.json
```diff
@@ -20,6 +20,8 @@
     "@actions/core": "^1.10.1",
     "@types/fs-extra": "^11.0.4",
     "axios": "^1.7.2",
+    "axios-mock-adapter": "^1.22.0",
```
de-jcup, line 23: axios-mock-adapter should be moved to devDependencies

github-actions/scan/src/launcher.ts
```diff
@@ -66,6 +68,7 @@ export const LAUNCHER_CONTEXT_DEFAULTS: LaunchContext = {
 
     inputData: INPUT_DATA_DEFAULTS,
     reportFormats: ['json'],
+    clientVersion: '1.6.1',
```
de-jcup, line 71: Why is the initial client version here `1.6.1` and not `latest`?

github-actions/scan/src/launcher.ts
```diff
@@ -84,17 +87,11 @@ export const LAUNCHER_CONTEXT_DEFAULTS: LaunchContext = {
  * Creates the initial launch context
  * @returns launch context
  */
-function createContext(): LaunchContext {
+async function createContext(): Promise<LaunchContext> {
```
de-jcup, line 90: Why did you change to asynchronous? The method is only used internally to create the context - and the caller is already asynchronous.

Was there any special purpose?

de-jcup (changes requested):
Looks very good :rocket: , I found only minor parts/ was unsure about some parts.

github-actions/scan/src/client-version-helper.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+function isValidVersion(version: string): boolean {
+    const regex = /^\d+\.\d+\.\d+$|^latest$/;
+    return regex.test(version);
+}
+
+async function getRedirectUrl(url: string): Promise<string> {
+    const response = await axios.get(url);
+    const $ = load(response.data);
```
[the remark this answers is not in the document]
hamidonos: The $ symbol is often used as a variable name in libraries like jQuery and Cheerio to provide a shorthand for calling functions.

image

We can also name it differently. In JavaScript it's not uncommon to access DOM elements via "$".

github-actions/scan/src/launcher.ts
hamidonos: take a look at line 94
the "getClientVersion" method returns a Promise
We have to wait for the version before continuing the procedure therefore we need to "await"
In JavaScript any "await" keyword has to be surrounded by a "async" function

PR #3190, opened 2024-06-05
Handle deployment without scheduler off #390

### LINKED ISSUE #390: Introduce state PAUSED into scheduler and provide automatic restart of PAUSED ones
We have already possiblity to restart jobs (for recognized zombie Jobs - e.g. when a server has crashed)

For rolling updates we would like to have a very fast deployment - without waiting for running jobs done...

When we introduce new scheduler job state PAUSED (similar to cancel, but without closing the Job!)
We could trigger PAUSED to all Jobs currently running when scheduler has been stopped!

When #389 is implemented, restart of scheduler can be done with server-version given, so only new servers are starting the scheduling.
Scheduling would fetch not only jobs in ready-to-start  state, but first all of in state PAUSED.
PAUSED jobs do normally have meta-data inside, so product restarts could be prevented and existing data reused

### LINKED ISSUE #3422: SecHub architecture documentation shall give only an overview about PDS concept
## Situation
Currently the full PDS concept is inside SecHub architecture documentation.
This is too much and and also irritating, because it looks like it would be the only PDS documentation (because much conent) - but isn't: there is much more in dedicated PDS doc!

## Wanted
- Only the point of view from SecHub side for PDS shall by in SecHub doc
- There shall a be a hint about the dedicated PDS documentation ( too look for details...)

## Solution
- change and divide + extend documentation

### LINKED ISSUE #3460: Architecture and Techdoc shall have same concept content
## Situation
Currently it is cumbersome to find information about the concepts.
Some parts are inside techdoc and some in architecture

## Wanted
The concepts shall be available on both documents

## Solution
Create one concept base asciidoc file which is included from both documents.

winzj:
This PR
- [x] closes #390 
- [x] closes #3422
- [x] close #3460


de-jcup:
I made now the most changes and want @winzj to do the review, so must close this one and create another one

Issue #3146, opened 2024-05-27
SecHub GHA: Add an option `addScmHistory:`


sven-dmlr:
## Situation
For secrets scans the client has been extended to upload scm history (e.g. `,git/` folder).
The option is `-addScmHistory`.
See https://mercedes-benz.github.io/sechub/latest/sechub-client.html#commandline-options

## ToDo
Add an option for this to the SecHub scan action.
`addScmHistory:` which defaults to `false`.

PR #3194, opened 2024-06-06
update action.yml to use 'latest' cli  version and set 'add-scm-history' parameter 

hamidonos:
- #3146


sven-dmlr (approved):
Looks good :+1:

PR #3221, opened 2024-06-12
Replace PDS_SCRIPT_ENV_WHITELIST env entry by PDS config entry 

### LINKED ISSUE #3220: Replace `PDS_SCRIPT_ENV_WHITELIST` env entry by PDS config entry
## Situation
With
- #3142

we kept the script environment as small as possible but have the possibility to add additional env values via `PDS_SCRIPT_ENV_WHITELIST`.

We have here 3 problems:

### Problem 1: Setup for every environment necessary

It works, but there is a need to handle/setup this in any kind of environment again and again
(e.g. local, local testing, builds (github,jenkins), k83s helm, system tests etc. etc.)

After all: too much complicated to configure and so error prone...

### Problem 2: Multiple products on same PDS will have both access
Multiple products on same PDS will have both access to the whitelisted parts - even when not
necessary.

### Problem 3: Not clear at first sight
It is not clear on first glimpse which product has additional script variable access or not.

## Wanted
- It shall be no configuration effort any more
- It shall be clear for an administrator what is whitelisted for a product by first glimpse
- It shall be possible to define a env whitelist setup for each product

## Solution
We change the existing implementation in a way, that no longer a env variable `PDS_SCRIPT_ENV_WHITELIST` is used for the comma separated list, but we move the information to the pds configuration file instead by providing a new (optional) field `envWhitelist`.

Example:

```json
{
  "apiVersion": "1.0",
  "serverId": "XYZ_Cluster",
  "products": [
    {
      "id": "product_1",
      "path": "/pds/scripts/xyz.sh",
      "envWhitelist" : [
            "ALLOWED_EVERY_*",
            "ALLOWED_SPECIAL_ONE"
      ],
      "scanType": "xyz",
      "description": "Runx xyz",
    }
  ]
}
```

Also the PDS startup summary shall show the envWhitelist entries.

### LINKED ISSUE #3222: Wrong test behavior of PDSUseSecHubCentralMappingInJobScenario16IntTest on local IDE execution
## Situation

We test inside `PDSUseSecHubCentralMappingInJobScenario16IntTest` that the white list entries for reduced script environment variable inheritance work as expected.

But when the integration test PDS server is started from an IDE and not from the gradle scripts, ,the test cannot be executed, because the variables to inherit are not defined. 

Currently `pds_script_does_have_the_mappings_from_sechub_injected_as_environment_variables` 
ist just skipped in this case and an error log message is done - but the test is green.

This is extreme irritating (at least for myself and I wrote this test... :flushed: )

## Wanted
Throw an exception with an info about the problem. The details shall be described still in the log output in detail, but the exception shall also show the necessary configuration of the wanted environment variables.

de-jcup:
This PR
- closes #3220
- closes #3222

### DIFF
--- a/sechub-doc/src/docs/asciidoc/documents/pds/pds_config.adoc
+++ b/sechub-doc/src/docs/asciidoc/documents/pds/pds_config.adoc
@@ -140,6 +140,13 @@ include::product_delegation_server_config_example1.json[]
 
 <12> Using `none` will not try to download any ZIP or TAR from storage but does call the caller script even when no data is available.
 
+<13> An optional list of environment variable names which are additionally white listed from PDS script environment cleanup.
+     Those entries can also end with an asterisk, in this case every variable name starting with this entry will 
+     be whitelisted (e.g. `SPECIAL_VAR_*` woud white list `SPECIAL_VAR_CONNECTION_URL` etc.) +
+     The PDS script cleanup process prevents inheritage of environment variables from PDS parent process.
+     There are some also some default variable names which are automatically accepted (e.g. `HOME`, `PATH`, ..).
+     
+
 === Launcher scripts
 
 ==== Generated variables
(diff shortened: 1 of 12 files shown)


winzj (approved):
Looking good 🚀

PR #3231, opened 2024-06-18
Feature 3029 archive support zip extraction security

### LINKED ISSUE #3029: ArchiveSupport ZIP extraction should handle ZIP archives with highly compressed large data more careful
### Situation
A test with a ZIP file with highly compressed large data, which when extracted exceed the PDS solutions workspace size, make a PDS job fail. The PDS will simply fail the job, because an exception happens with the message: `No space left on device`.
The PDS then cleans up the workspace for this job and starts the next job. This could cause problems for other jobs running on the same PDS as well, if someone tries this on purpose e.g. multiple times to create heavy load.

#### Important Insights
- Files are uploaded and extracted
- PDS fails the job if the workspace size is not sufficient for the extracted data size
- ZIP archives are not extracted recursively, which means no recursive ZIP bombs are possible

### Wanted
The `ArchiveSupport` must be able to check the uploaded sources ZIP file before or while extracting it and fail the job without risking to fail other jobs currently running on the PDS solution.

### Discussed solutions
- [ ] `configurable time`: Fail the job before extracting everything, if it takes longer than the configured time value to extract an entry.
- [ ] `configurable size`: Fail the job before extracting everything, if the stream archive entry is to big
- [ ] `compression rate`: Fail the job before extracting everything, if the compression rate is "unusually" high, like 5Kb would extract to like 10GB.

All these options should be configurable via a PDS parameter. This way administrators can adjust these values as needed. A reasonable default value must be set for each parameter.

hamidonos:
- closes #3029


sechub-commons-archive/src/main/java/com/mercedesbenz/sechub/commons/archive/ArchiveSupport.java
```diff
@@ -508,7 +491,7 @@ private ArchiveExtractionResult extract(ArchiveInputStream sourceArchiveInputStr
                     continue;
                 }
                 try (OutputStream outputFileStream = new FileOutputStream(outputFile)) {
-                    IOUtils.copy(sourceArchiveInputStream, outputFileStream);
+                    IOUtils.copy(safeArchiveInputStream, outputFileStream);
```
de-jcup, line 494: ## I thought about following:
The tests do check the implementations of `FileZip`, `SafeArchiveIpustream` etc. All well done tested, but what we do not test is, that extracting with `ArchiveSupport` does always use a `SafeARchiveInputStream` at the end/ the archive support is using the implemented protection against Zip bombs at all.

## We could do this following ways IMO:

a) we write an integration test with a zimb bomb upload via rest directly - and check for upload error without upload - one problem, e.g. too many directories - would  be enough. But this could (maybe) lead to a longer running integrtion test? (and we do not want to increase build time much)

b) we could wrap here the static call via IOUtils - e.g. by a package private field using a `IOUTilsCopyWrapper` (or a better interface name...) with a default implementation which uses IOUtil.copy... Inside the test we could mock it with `Mockito` and check if it is called with a a `SafeArchiveInputStream`.

I think we should implement at least one of the possibilities.

sechub-commons-archive/src/main/java/com/mercedesbenz/sechub/commons/archive/FileSize.java
```diff
@@ -0,0 +1,102 @@
...
+ * System.out.println(fileSize.getSizeBytes()); // Prints: 10737418240
+ * </pre>
+ *
+ * <p>
+ * Note: This class does not support file sizes in terabytes (TB) or higher, or
+ * sizes without a unit.
+ *
+ * @author hamidonos
+ */
+public class FileSize {
```
de-jcup, line 41: Nice class :-)

sechub-commons-archive/src/test/java/com/mercedesbenz/sechub/commons/archive/SafeArchiveInputStreamTest.java
```diff
@@ -0,0 +1,267 @@
...
+        Duration timeout = Duration.ofSeconds(10L);
+
+        ArchiveExtractionContext properties = new ArchiveExtractionContext(maxFileSizeUncompressed, maxEntries, maxDirectoryDepth, timeout);
+        ArchiveInputStream<?> archiveInputStream = createZipArchiveInputStream(maxEntries, entryFileSize, maxDirectoryDepth);
+
+        /* execute */
+        SafeArchiveInputStream result = new SafeArchiveInputStream(archiveInputStream, properties);
+
+        /* test */
+        assertThat(result.getProperties(), is(properties));
```
de-jcup, line 57: Please rename properties variable and getter accordingly after the Context class has been renamed, 

```suggestion
        assertThat(result.getConstraintData(), is(constraintData));
```

sechub-commons-archive/src/main/java/com/mercedesbenz/sechub/commons/archive/SafeArchiveInputStream.java
```diff
@@ -0,0 +1,96 @@
...
+
+        ArchiveEntry entry = inputStream.getNextEntry();
+        if (entry == null) {
+            return null;
+        }
+
+        long maxDirectoryDepth = properties.getMaxDirectoryDepth();
+        long directoryDepth = entry.getName().split(DIRECTORY_DELIMITER).length - 1; // Subtract 1 because split includes the file name
+        if (directoryDepth > maxDirectoryDepth) {
+            throw new IllegalArgumentException("Directory depth exceeds the maximum allowed value of %s".formatted(maxDirectoryDepth));
```
de-jcup, line 84: I would not expect here an Illegal argument exception. In all of the getNextEntry() assertions I would throw an IOException - input cannot be transformed to output because of an error.

I would prefer to introduce here an `ArchiveExtractionException` which would extends IoException or RuntimeException. Or maybe another name, but a dedicated exception for the potential security problem which is an `IOException`.

Alternatively we could use `IllegalStateException` - but `IllegalArgument` sound odd to me, because it is no argument but a corrupted state

de-jcup (changes requested):
I like your implmentation. But I found some things to discuss. Please look into my comments.

sechub-commons-archive/src/main/java/com/mercedesbenz/sechub/commons/archive/ArchiveSupport.java
```diff
@@ -403,18 +384,19 @@ private void compressRecursively(String basePath, ArchiveOutputStream outputStre
...
     private ArchiveExtractionResult extractTar(InputStream sourceInputStream, String sourceLocation, File outputDir,
-            SecHubFileStructureDataProvider fileStructureProvider) throws IOException {
+            SecHubFileStructureDataProvider fileStructureProvider, ArchiveExtractionContext properties) throws IOException {
         try (ArchiveInputStream archiveInputStream = new ArchiveStreamFactory().createArchiveInputStream("tar", sourceInputStream)) {
             if (!(archiveInputStream instanceof TarArchiveInputStream)) {
                 throw new IOException("Cannot extract: " + sourceLocation + " because it is not a tar tar");
             }
-            return extract(archiveInputStream, sourceLocation, outputDir, fileStructureProvider);
+
+            SafeArchiveInputStream safeArchiveInputStream = new SafeArchiveInputStream(archiveInputStream, properties);
```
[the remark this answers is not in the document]
hamidonos: No, this is a custom class and it does not implement `Closable`

de-jcup (approved):
Looks good to me :+1:

PR #3236, opened 2024-06-20
Improve prepare application and PDS solution

### LINKED ISSUE #3233: Improve PDS Solution for Prepare for new environment variable handling
## Situation
PDS solution for prepare has no access to storage variables as they were handed over via spring.

## Wanted
Improve the PDS solution for the prepare wrapper to be able to use the newly implemented environment variable handling.

### LINKED ISSUE #3234: Prepare Wrapper Application add Tag to docker archive
### Situation
The Skopeo module downloads the docker images, but it does not add Tags to the archive (like the docker save command does). To use the PDS Xray tags are necessary to get the image name and tag to be able to upload it to the artifactory. 

### Wanted
the image name should be added as tag to the docker archive

### Solution
Make use of skopeo `-- add-additional-tag <tag>` with the name of the image that we want to scan. The image name can be get from the location defined in the remote data section.

lorriborri:
- closes #3233 
- closes #3234


sechub-wrapper-prepare/src/main/java/com/mercedesbenz/sechub/wrapper/prepare/modules/skopeo/SkopeoLocationConverter.java
```diff
@@ -16,7 +16,7 @@ public String convertLocationForLogin(String location) {
         return removeProtocolPrefix(location);
     }
 
-    private String removeProtocolPrefix(String location) {
+    public String removeProtocolPrefix(String location) {
```
de-jcup, line 19: The class is named SkopeoLocationConverter.

You made the private method here just public. but we use internally the public method again public->private->public ... 

Please do following:
- keep the private method as private.
- introduce a method "convertToAddititionalTag(...)" and ue this inside Skopeo wrapper. You can call the private part from there
- add some unit tests for convertToAddititionalTag
- add javadoc and describe on convertToAddititionalTag if the tag result could ever be empty or null (we add it inside commands list and I am currently not 100% sure what happens /if it is impossible, that we could have here null

so it is clear that we use something for tagging.
And it is tested.

sechub-wrapper-prepare/src/main/java/com/mercedesbenz/sechub/wrapper/prepare/modules/skopeo/SkopeoWrapper.java
```diff
@@ -114,10 +114,13 @@ private ProcessBuilder buildProcessDownload(SkopeoContext context) {
         List<String> commands = new ArrayList<>();
 
         String location = locationConverter.convertLocationForDownload(context.getLocation());
+        String additionalTag = locationConverter.removeProtocolPrefix(context.getLocation());
         File downloadDirectory = context.getToolDownloadDirectory().toFile();
 
         commands.add("skopeo");
         commands.add("copy");
+        commands.add("--additional-tag");
+        commands.add(additionalTag);
```
de-jcup, line 123: see my comments at location converter - we must ensure, that additional tag can never be null/empty or we handle the situation and do a conditional add here

sechub-wrapper-prepare/src/test/java/com/mercedesbenz/sechub/wrapper/prepare/modules/skopeo/SkopeoWrapperTest.java
```diff
@@ -143,11 +143,10 @@ void when_process_throws_exception_then_download_throws_exception() throws IOExc
 
     }
 
-    private void assertProcessBuilderFactoryCalledWithParametersForLogin(SkopeoContext context, String location, String user,
-            ProcessBuilderFactory processBuilderFactory) {
+    private void assertProcessBuilderFactoryCalledWithParametersForLogin(String location, String user, ProcessBuilderFactory processBuilderFactory) {
 
         // skopeo, login, ubuntu:22.04, --username, username1, --password, password1,
-        // --authfile, authentication.json]
+        // --authfile, authentication.json
```
de-jcup, line 149: Please add the --additional-tag part to the example output (or just remove the two example lines)

sechub-wrapper-prepare/src/main/java/com/mercedesbenz/sechub/wrapper/prepare/modules/git/GitPrepareWrapperModule.java
```diff
@@ -171,4 +173,11 @@ private void prepareRemoteConfiguration(GitContext gitContext, SecHubRemoteDataC
         gitWrapper.downloadRemoteData(gitContext);
     }
 
+    private String getRepositoryNameFromLocation(String location) {
```
de-jcup, line 176: Hm...we have no unit tests for this!

I would prefer following here:

1. Inject `GitLocationConverter` via `@Autowire` to the git prepare module here.

2. Enhance existing `GitLocationConverter` class and introduce new method `convertToRepositoryName(String location)`

3. Use it here
4. Add some unit tests in `GitLocationConverterTest` for `convertToRepositoryName(String location)`

5. You could additionally add also a unit test for the wrapper, that the 
`GitLocationConverter.convertToRepositoryName(String location)` is the name for the output folder or at least that convert is called

de-jcup (changes requested):
Good job, but I have found some parts to enhance.

sechub-wrapper-prepare/src/test/java/com/mercedesbenz/sechub/wrapper/prepare/modules/skopeo/SkopeoWrapperTest.java
lorriborri: skopeo login does not need additional tag and does not support additional tag

sechub-wrapper-prepare/src/main/java/com/mercedesbenz/sechub/wrapper/prepare/modules/git/GitLocationConverter.java
```diff
@@ -30,6 +30,13 @@ public class GitLocationConverter {
     @Autowired
     PDSLogSanitizer pdsLogSanitizer;
 
+    public String convertLocationForRepositoryName(String location) {
```
de-jcup, line 33:
```suggestion
    public String convertLocationToRepositoryName(String location) {
```

sechub-wrapper-prepare/src/main/java/com/mercedesbenz/sechub/wrapper/prepare/modules/skopeo/SkopeoLocationConverter.java
```diff
@@ -1,13 +1,17 @@
...
 import org.springframework.stereotype.Component;
 
 @Component
 public class SkopeoLocationConverter {
 
     private static final String PROTOCOL_SPLIT = "://";
 
+    private static final String DEFAULT_TAG_PREFIX = "default-tag";
+
     public String convertLocationForDownload(String location) {
```
de-jcup, line 15:
```suggestion
    public String convertLocationToDockerDownloadUrl(String location) {
```

sechub-wrapper-prepare/src/main/java/com/mercedesbenz/sechub/wrapper/prepare/modules/skopeo/SkopeoLocationConverter.java
```diff
@@ -16,7 +20,23 @@ public String convertLocationForLogin(String location) {
...
+     * default tag is created.
+     *
+     * @param location the location to convert
+     * @return the location without protocol prefix or default tag
+     */
+    public String convertLocationForAdditionalTag(String location) {
+        String withoutProtocol = removeProtocolPrefix(location);
+        if (withoutProtocol.isBlank() || withoutProtocol.isEmpty()) {
+            // we add a random number to prevent conflicts during scans
+            withoutProtocol = DEFAULT_TAG_PREFIX + ":" + new Random().nextInt(1000);
```
de-jcup, line 34: Hm... Random... 1000...
not really random.

1. I would prefer here something which could not lead to situation that two calls would produce same appendix. For example a UUID, a datetimestamp or just system nano seconds
2. I do not really get the tag purpose here. We use this logic when we download via skopeo from a docker registry. Okay. But why do we need a default flag here, which has a random part at the end? And what exactly do we suppose to download here in this case? Why is the default tag not just "latest" ?!?

sechub-solution/README.adoc
```diff
@@ -14,6 +14,18 @@ To start a single container run the script:
 ./01-start-single-docker-compose.sh
 ----
 
+=== Single Container with Prepare (remote Data)
+
+. Adjust the `docker-compose_sechub-debiian-prepare.yml` file and replace <user> with your linux home directory. +
+Make sure the ./sechub/shared_volume folder exists and is writable for the docker. If not create the directory and execute `sudo chmod -R 777 ./sechub/sharedvolume`. +
+This mounted NFS must be the same directory as the NFS directory in the PDS prepare (/pds-solutions/prepare/docker-compose-prepare_external-network.yml.
```
de-jcup, line 21: Sorry, but I do not really like that somebody has to replace pseudo parameter <user> if somebody wants to run sechub solution . I know we discussed that we could use user home location - but I was not aware that a user has to do something here manually.

SecHub shall always work "out of the box".

An alternative could be to use
https://docs.docker.com/reference/dockerfile/#arg and provide this by startup script to the docker startup phase.

But I think the easier way would be to 
- ensure /tmp/sechub-shared-volume in shell scripts (before startup)
- use this path in both parts as shared volume
- done

When this is done, the block inside the documentation should changed.

sechub-pds-solutions/prepare/README.adoc
```diff
@@ -37,7 +37,11 @@ The folder contains a start script which does the manual steps for you:
 
 The container will be started and attached to the `sechub` network.
 
-WARNING: Make sure the SecHub container is running and executor for Prepare is set up.
+WARNING: Make sure the SecHub container is running with compose prepare and executor for Prepare is set up.
+
+. Adjust the docker-compose_pds_prepare_external-network.yml file and set the volume for the shared_volume directory to your local .sechub folder +
```
de-jcup, line 42: This should not be necessary - please look at my comments below (I already mentioned this).

In a nutshell:
- user may not need to change something
- tmp/sechub-shared-volume could be an option for this
- shall always work out-of-the-box

de-jcup (changes requested):
I found still some parts.

de-jcup (approved):
Looks good to me.

PR #3248, opened 2024-06-21
Fix PDS startup problems #3246

### LINKED ISSUE #3246: PDS startup fails when not integration test
## Situation
After using implementation of 
- #3226 
the PDS does only start in integration test profile.

Reason: 
PDSSecurityConfiguration does reset given credential information to null before the registration of the sensitive setup is done. Means: registry contains null values which will always be different to the correclty defined environment variable entries....

## Wanted
May not happen

de-jcup:
This PR
- closes #3246


lorriborri (approved):
Looks good to me :+1: :star: Good work!
Also tested local with docker compose - worked with pds-prepare

PR #3254, opened 2024-06-27
SecHub data encryption #3250

### LINKED ISSUE #3250: Encrypt sechub configuration in database
## Situation
Currently the user configuration is not encrypted.

With #837 we some security helper classes which provide a secure way to handle this.

## Wanted 
We want user configuration data encrypted inside database. So user credentials inside configurations will be stored encrypted inside database

## Solution
- We introduce a new column inside Job table `JOB_CONFIG_ENCRYPTION` - per default it is `NONE` so we know which (old) job is NOT encrypted. 
- migrationscript uses `NONE` as default for database
- We introduce a new configuration value for sechub server (spring value) which contains a string with a secret key
- Before writing job data the job configuration will be encrypted by secret key, also the `JOB_CONFIG_ENCRYPTION` will be set
- Before returning job configuration object for admin rest calls we must decrypt, so admin can directly see the origin configuration by admin REST api

### LINKED ISSUE #3263: Handle SecHub configuration encryption at SecHub server
## Situation
Currently the user configuration is not encrypted.

With #837 we have now some security helper classes which provide a secure way to handle this.

:information_source: _This is a sub issue of #3250_ 

## Wanted 
We want user configuration data encrypted inside database. So user credentials inside configurations will be stored encrypted inside database

## Solution

- [x]  Introduce new table to hold encryption data for cluster (use concept created in epic #3250 )
- [x] Write migration sql flyway script - NONE encryption after migration
- [x] Startup phase blocks when a server does not support encryption / sanity check fails
- [x] Schedule job will only fetch jobs which have supported encryption
      (means trigger service repository call needs changes)
- [x] Introduce UC for Automatically update existing jobs which are done via a trigger service (spring scheduling)
- [x] Introduce REST endpoint + UC for admins to trigger cipher rotation (means add new cipher pool entry which is used as latest one + check encryption works)
- [x] Introduce REST endpoint + UC for admins to check cipher pool situation (which are used)
- [x] Autocleanup process does also 
      internal check is still done that there are no longer jobs which uses this pool id entry, also the latest will never be deleted...
- [x] New jobs have automatically encrypted configuration of latest supported entry from cipher pool.
- [x] Introduce UC for cipher pool refresh inside `ScheduleEncryptionService`
   (for all server instances in cluster)
- [ ] Update OpenAPI file definition for new REST endpoints.

### LINKED ISSUE #3264: Encrypt SecHub configuration in PDS
## Situation
We store the parts of the SecHub configuration inside PDS database

:information_source: _This is a sub issue of #3250_ 

## Wanted
This must be encrypted - similar to #3263

### LINKED ISSUE #3267: Remove  SecHub configuration from all REST end points of SecHub except one
## Situation
With #3266 there will be a dedicated REST endpoint to fetch masked SecHub configuration (as user of the project or as administrator).

It shall be the only possible way to fetch these information. But at some existing REST end points, we provide the SecHub configuration as well (not masked).

:information_source: _This is a sub issue of #3250_ 

## Wanted
No REST endpoint at SecHub except #3266 shall have the SecHub configuration inside!

## Solution
Identify all REST end points where configuration is provided accidentally and remove the information there

### LINKED ISSUE #3268: Remove SecHub configuration from all REST end points of PDS
## Situation
With #3266 there will be a dedicated REST endpoint to fetch masked SecHub configuration (as user of the project or as administrator).

It shall be the only possible way to fetch these information. But it could be possible the we provide the configuration at some existing PDS REST end points

:information_source: _This is a sub issue of #3250_ 

## Wanted
No REST endpoint at PDS shall provide the SecHub configuration

## Solution
Check all REST end points if configuration is provided accidentally and remove the information there

### LINKED ISSUE #3273: Introduce gradle submodule `sechub-encryption` and move  encryption implementations to this library
## Situation
Currently the encryption classes are settled inside `sechub-commons-core`.

## Wanted
The core encryption classes shall be inside its own gradle sub module `sechub-encryption`

## Solution
- create new gradle submodule
- use it where necessary

### LINKED ISSUE #3274: Refactorings inside sechub encryption library
## Situation
Some parts inside encryption library can be improved.

### Drop unnecessary created objects

- https://github.com/mercedes-benz/sechub/blob/14862ef7455d7f69c34edcd2135441776e54169d/sechub-commons-core/src/main/java/com/mercedesbenz/sechub/commons/core/security/persistence/AesGcmSiv.java#L97
- https://github.com/mercedes-benz/sechub/blob/14862ef7455d7f69c34edcd2135441776e54169d/sechub-commons-core/src/main/java/com/mercedesbenz/sechub/commons/core/security/persistence/AesGcmSiv.java#L126

These given secret is already a secret spec

### Reduce possibilities of cipher creation
- Currently there are two ways to create a PersistenceCipher: Via static method directly on the cipher class and also on via a factory. 
- Static create method on AesGCMSiv does also automatically do some automatic mechanism to differ between 256 bit and 128bit cipher, but the factory provides different types via enum and uses same method later on static call
- Uses should only be able to use the factory methods from "outside"

### Consistent naming
Some naming shall be more consistent - e.g. AesGCMSiv is a PersistenceCipher implementation and NoCipher also. But only NoCipher has the naming of the interface as a part inside.

### Package move
Because we move the encryption to a new library in #3273, we want the package to be renamed as well.

### Create a usage test in `sechub-test` and maybe simplify usage from outside
The existing encryption classes are tested well, but separately and never outside the package.
It is not easy to get on the first glance how to use them.

If the tests show up that there is no need to make the factories public we should make them package private.

## Refactoring
After create usages tests for the current implementation the `BinaryText` approach was very confusing and also led to many situation where it was wrong used.

The usage of static factory methods makes it impossible to inject it by spring later.

After a while decision was made to simplify usage. When base64 or hex wants to be supported by string conversion the caller side knows if the content encoding is base64 or hex and should handle it accordingly. There shall be no different handling inside `sechub-encryption` for base64, hex or plain strings. 
We also want to have the possibility to handle any kind of data to encrypt. So we the PersistentCipher interface will provide byte array as raw input source.

Via a ByteArrayTransformer implementations and a EncryptionSupport we will bring up convenience for some object types - at the beginning only for String.

## Rename library
We rename the library now to `sechub-commons-encryption`. Doing this it becomes clear that we do not want to have any dependency to Spring framework or to SecHub (having no dependency to SecHub makes it also easier to use it with PDS later). The only dependencies shall be bouncy-castle and slf4j

### LINKED ISSUE #3275: Upgrade bouncy castle version to 1.78.1 inside `sechub-encryption`
## Situation
There is a newer bouncy castle version available

## Solution
Upgrade to 1.78.1

### LINKED ISSUE #3321: Change log level for domain message handler registration to trace level
## Situation
Currently the log contains on startup always debug information about which message handler is registered etc. This information is most times not wanted, even for debugging.

## Wanted
Shall be only shown in logs if tracing for domain message service is enabled

### LINKED ISSUE #3322: Integration tests have growing test scenario failure when server restart necessary (with postgres DB)
## Situation
When I start `JobScenario2IntTest` and have following scenario:
- using a local postgres DB
- a running SecHub server in integration test mode (with postgres as DB)

the first run will work.

Now we just terminate and relaunch the SecHub server (same setup) and execute the test again (postgres database is still running and has old test data inside).
This leads to following problem:

```
java.lang.AssertionError: s02_0001_owner1 found!
	at com.mercedesbenz.sechub.integrationtest.api.AbstractAssert.expectHttpClientError(AbstractAssert.java:34)
	at com.mercedesbenz.sechub.integrationtest.api.AssertUser.doesNotExist(AssertUser.java:52)
	at com.mercedesbenz.sechub.integrationtest.api.AssertUser.doesNotExist(AssertUser.java:35)
	at com.mercedesbenz.sechub.integrationtest.internal.ScenarioInitializer.createUser(ScenarioInitializer.java:103)
	at com.mercedesbenz.sechub.integrationtest.scenario2.Scenario2.initializeTestData(Scenario2.java:81)
	at com.mercedesbenz.sechub.integrationtest.internal.AbstractSecHubServerTestScenario.prepareImpl(AbstractSecHubServerTestScenario.java:261)
	at com.mercedesbenz.sechub.integrationtest.internal.AbstractSecHubServerTestScenario.prepare(AbstractSecHubServerTestScenario.java:200)
	at com.mercedesbenz.sechub.integrationtest.api.IntegrationTestSetup$IntegrationTestStatement.evaluate(IntegrationTestSetup.java:195)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.lang.Thread.run(Thread.java:840)

```
When I restart the test multiple times, the user ids etc. are increasing and after some time the test will work again, because a user id will be used which is not existing.

Workaround works... but is very ugly and inconvenient

## Wanted
The growing scenario setup status must be persisted and reused on startup

### Analyze
- `AbstractGrowingSecHubServerTestScenario` uses already `PersistentScenarioTestDataProvider`.
- added trace log parts
- funny: worked some time as it should
- now problem appeared again:
  `0:57:16.407 [main] TRACE c.m.s.i.i.PersistentScenarioTestDataProvider -- File NOT exists: /$myRepoLocation/sechub-integrationtest/build/sechub/integrationtest/scenario-testdata_scenario2.properties`
- means the file has been deleted!

### Sum up
The file is inside the gradle build folder. Means: if a gradle clean task is executed, this build folder will be removed and the file is deleted.
This scenario must be handled by developers on their local machine by themselves.

Inside the code of `PersistentScenarioTestDataProvider` there is also a delete mechanism implemented: when the properties load failed for any reason (any exception) the file was also deleted. 

The property load mechanism shall be more resilient and should do a retry before finally deleting and failing. This will be changed.

### LINKED ISSUE #3332: Introduce SecHubMultiSpringBootTest to reduce build time
## Situation
We have currently some spring boot tests which need all classes loaded.
Those tests are extreme time consuming (> 7 seconds to startup).

- TargetResolverServiceSpringBootTest
- SchedulerCreateJobServiceSpringBootTest
- SchedulerGetJobStatusServiceSpringBootTest
- SchedulerSmokeSpringBootTest

Means, we have here > 28 seconds for those tests. The tested setup is always the same.

## Wanted
Keep tests but reduce time to run

## Solution
We introduce new spring boot test: `SecHubMultiSpringBootTest` and combine the 4 test 
classes and its methods inside one test.

We have also "normal" junit tests for the tested service, which are very fast and the naming
here is correct (XYZ is tested with XYZTest).

Not adhering to the naming pattern in these tests is acceptable due to the speed advantages and the scope of the tests (only holistic tests in the complete container).

de-jcup:
This PR

handles encryption library
- [x] closes #3273
- [x] closes #3274
- [x] closes #3275

Handles encryption of SecHub configuration
- [x] closes #3263
- [x] closes #3264
- [x] closes #3267
- [x] closes #3268
- [x] closes #3332

after all
- [x] add REST API changes + new REST API parts to manual written OpenApi file!
  - [x] Removed config from job info results (admin)
  - [x] New API or Admins and Encryption

Additionally 
- [x] closes #3321
- [x] closes #3322


sechub-shared-kernel/src/test/java/com/mercedesbenz/sechub/sharedkernel/encryption/SecHubEncryptionStatusTest.java
```diff
@@ -0,0 +1,50 @@
...
+        schedulerStatus.getData().add(scheduleEntry0);
+        schedulerStatus.getData().add(scheduleEntry1);
+
+        status.getDomains().add(schedulerStatus);
+
+        /* execute */
+        String json = status.toFormattedJSON();
+
+        /* test */
+        System.out.println(json);
```
winzj, line 46: Is this sysout wanted to see the output in a well formatted way?

sechub-test/src/test/java/com/mercedesbenz/sechub/test/encryption/EncryptionUsageTest.java
```diff
@@ -0,0 +1,165 @@
...
+        /* @formatter:on */
+        String unencryptedTestData = encryptSupport.decryptString(newEncrypted, cipherNewPassword, newInitVector);
+
+        assertThat(testData).isEqualTo(unencryptedTestData);
+
+        // additional test - we check that the decryption with old init vector does not
+        // work
+        assertThatThrownBy(() -> encryptSupport.decryptString(newEncrypted, cipherNewPassword, firstEncryptionResult.getInitialVector()))
+                .isInstanceOf(IllegalStateException.class).hasRootCauseInstanceOf(AEADBadTagException.class);
+    }
```
winzj, line 164: Just an idea for the future, we could also extend this tests later:
For example, we could add a test where the new cipher also uses a different `PersistentCipherType`, like `PersistentCipherType.AES_GCM_SIV_128` or something completely different as soon as we want to support something else.

sechub-shared-kernel/src/test/java/com/mercedesbenz/sechub/sharedkernel/encryption/SecHubSecretKeyProviderFactoryTest.java
```diff
@@ -0,0 +1,154 @@
...
+        factoryToTest = new SecHubSecretKeyProviderFactory();
+        encryptionEnvironmentEntryProvider = mock(EncryptionEnvironmentEntryProvider.class);
+
+        factoryToTest.encryptionEnvironmentEntryProvider = encryptionEnvironmentEntryProvider;
+
+    }
+
+    @ParameterizedTest
+    @EnumSource(SecHubCipherPasswordSourceType.class)
+    void none_cipher_no_encryptionEnvironmentEntryProvider_used_no_matter_which_pwd_source_type_defined(SecHubCipherPasswordSourceType type) {
```
winzj, line 40: I think ParameterizedTest is not necessary here.

winzj (changes requested):
Looking very good to me 🚀 

I found some minor things to change. Please have a look and let me know if I misunderstood something.

sechub-server/src/main/resources/db/migration/common/U29__encryption.sql
```diff
@@ -0,0 +1,17 @@
+-- SPDX-License-Identifier: MIT
+
+-- encryption parts
+-- FIXME de-jcup: discussion necessary, maybe server 2.0.0 (major update necessary)
```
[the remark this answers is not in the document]
de-jcup: I thought about it and the downgrade SQL scripts will be implemented/available, but with a remark that it is not recommended to use them because of potential data loss.

sechub-shared-kernel/src/main/java/com/mercedesbenz/sechub/sharedkernel/messaging/SecHubEncryptionMessageDataProvider.java
```diff
@@ -0,0 +1,34 @@
...
+
+    @Override
+    public SecHubEncryptionData get(String encryptionDataAsJson) {
+        if (encryptionDataAsJson == null) {
+            return null;
+        }
+        try {
+            return SecHubEncryptionData.fromString(encryptionDataAsJson);
+        } catch (IllegalArgumentException e) {
+            LOG.error("No UUID transformable because {} is not a valid UUID!", encryptionDataAsJson);
```
winzj, line 21: Can you update the error log.
```suggestion
            LOG.error("No sechub encryption data transformable because {} is not a valid sechub encryption data json!", encryptionDataAsJson);
```

winzj (approved):
Good job - looking good to me 👍 

Maybe you can have a look at the one error log I marked.

PR #3259, opened 2024-07-01
Feature 3255 solve PDS prepare s3 storage problem

### LINKED ISSUE #3255: Prepare wrapper has problems with S3 storage
## Situation
The PDSPrepareIntegrationScenario22IntTest does always fail when we start the test and use not a shared volume but a S3 storage.

## Wanted
Solve the S3 storage problem

### LINKED ISSUE #3256: Provide possibility to add PDS script output to normal logs via PDS parameter
## Situation
Every PDS job starts a launcher script. The output of this script and its sub processes is stored in files and fetched and persisted when the job has failed or succeeded.

An administrator is able to fetch the persisted output for each dedicated job.

### Problem
When it comes to problems - e.g. on a wrapper application which is used by the script - it is not possible to just look into the PDS logfile to get the information but you always need to retrieve the information from database / via REST.

## Wanted
It shall be possible to define at runtime, if the logs from launcher script and its subprocesses is added to the normal PDS log file.

## Solution
We introduce a new PDS parameter key for this. When not defined in executor configuration the initial default will be false.

When enabled, the PDS will read system-out.log and system-error.log before the workspace cleanup is done (no matter if failed or succeeded). Each line will be logged by logger.

### LINKED ISSUE #3261: Provide possibility for PDS scripts with wrapper applications to enable remote debugging
## Situation
When we have a java wrapper application which is started by PDS it can be very useful to enable remote debugging. But currently we have no PDS parameter for this

## Wanted
It shall be possible to define via PDS parameter inside a executor configuration that we want remote debugging for wrapper application.

## Solution
- We introduce new PDS parameter `pds.wrapper.remote.debugging.enabled`
- For the prepare wrapper PDS solution we use the new parameter for launching inside script 
  (as a prototype, will be changed for other wrappers in own issues)

de-jcup:
This PR
- closes #3255
- closes #3256
- closes #3261


lorriborri (changes requested):
Good Job! :star: :+1: 
I just found one small Issue.

lorriborri (approved):

PR #3277, opened 2024-07-08
Remote debugging capability for wrapper

### LINKED ISSUE #3262: Add missing remote debugging capability for other wrappers
## Situation
With #3261 we added remote debugging parameter and implemented it in PDS solution for `prepare`

## Wanted
The script call from prepare shall be adopted to

- [ ] zap pds solution
- [ ] xray pds solution
- [ ] checkmarx pds solution

## Solution
It should be enough to handle this as done for prepare.sh: 

```bash
if [[ "$PDS_WRAPPER_REMOTE_DEBUGGING_ENABLED" = "true" ]]; then
    options="$options -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=8000"
fi

java -jar $options "$pathToJar"
```

lorriborri:
- closes #3262


de-jcup (approved):
Well done :-)

PR #3279, opened 2024-07-08
implement openapi.yaml in new sechub-openapi-java module #3238

### LINKED ISSUE #3238: Implement the openapi spec files manually
### Situation

Automatically generating the OpenAPI specification file can be a good practice in some cases, but there are several reasons why you might want to avoid it:

1. **Accuracy**: Automatically generated specifications can sometimes miss out on details or not fully capture the nuances of your API. This can lead to inaccuracies in the generated documentation or client SDKs.

2. **Customization**: When you generate the specification automatically, you might not have as much control over the final output. This can limit your ability to customize the documentation or the structure of the specification.

3. **Maintenance**: If your API changes frequently, maintaining an automatically generated specification can become challenging. You would need to regenerate the specification every time there's a change in the API.

4. **Performance**: The process of generating the specification can add overhead to your build or deployment process, which might not be desirable in some cases.

5. **Security**: If the generation process is not properly secured, it could potentially expose sensitive information about your API.

### Wanted

We want to maintain our openapi spec files manually so that we have full control over the specification. Furthermore we don't want to rely on any 3rd party generators & gradle build steps.

### Solution

- (Re)Write the openapi spec files manually (check what can be re-used).
- Put the new openapi spec file into a new module `sechub-openapi-java`

hamidonos:
This PR
- closes #3238

Remark: use https://editor.swagger.io to view the openapi.yaml file


sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/mapping/UpdateMappingService.java
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT
 package com.mercedesbenz.sechub.domain.administration.mapping;
 
-import static com.mercedesbenz.sechub.sharedkernel.validation.AssertValidResult.*;
+import static com.mercedesbenz.sechub.sharedkernel.validation.AssertValidResult.assertValid;
```
de-jcup, line 4: Normally our code setup for imports says: static imports always * - here okay.

Maybe we should implement a spotless rule for this.

PR #3280, opened 2024-07-08
Feature 3141 implement first version of secret validator application

### LINKED ISSUE #3141: Implement first version of a spring application that helps to categorize and verify secretscan results
### Situation
Currently we rely on the secret scan results of the tools we are using. If the secret scan does not categorize findings with the appropriate severity we currently take the default of the tool, or in case of SARIF the default of the reporting format.  We want to be able to categorize the findings more accurately, e.g. by using defaults on our own experience or trying to verify secrets or tokens by performing API calls. 

#### Example
To verify a Github personal access token we can try to access the Github API. This can be done for other APIs as well, which makes it easier to change the severity of a finding. Verified findings like valid Github personal access tokens will always be `CRITICAL` inside the SecHub report.

### Solution
- an application that can run inside any secret scan pds-solution
- the application takes the given secret scan report (we start with the SARIF-2.1.0 format) and applies a more accurate categorization of the secrets
- the application also tries to verify secrets by calling dedicated APIs
- each finding of the SARIF report will then be provided with a generic SARIF properties bag, with defined key value patterns like:
```json
{
"sechubSeverity": "CRITICAL",
"sechubVerifiedBy": "https://api.example.com"
}
```
- the spring application must be provided with a configuration that knows about the secret scanning tools rule structure. Here an example on how this could look like for pds-gitleaks:
```json
{
  "config" : [ {
    "ruleId" : "github-pat",
    "urls" : [ "https://api.github.com" ],
    "defaultSeverity" : "high",
    "verificationInformation" : {
          "verifiedHTTPStatus" : 200,
          "responseContains" : [ "" ],
    }
  }]
}
```

### Remark
The verification of the secrets will be an iterative process, because for each secret some research is necessary to collect all necessary data on the respective webservice that is needed to verify potential secrets.

### LINKED ISSUE #3276: Sereco SarifV1Importer shall be able to import SARIF PropertyBag from Secret Scans
### Situation
With #3141 we introduce an application which provides more information on secret scan severity and if the secret was validated, there will be the URL available the secret is valid for.

### Wanted
Inside Sereco we need to be able to import these additional information.

### Solution
- extend the SarifV1Importer
- extend the GitleaksSarifImportWorkaround

winzj:
- closes #3141 
- closes #3276


sechub-wrapper-secret-validator/src/test/java/com/mercedesbenz/sechub/wrapper/secret/validator/cli/SecretValidatorSpringBootTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Test
+    void execution_service_with_correct_configuration_without_validation_categprizes_findings_with_default_configured() {
+        /* execute */
+        SarifSchema210 report = executionService.execute();
+
+        /* test */
+        Run run = report.getRuns().get(0);
+        for (Result finding : run.getResults()) {
+            Location location = finding.getLocations().get(0);
```
de-jcup, line 45: Hmm... at the first glance I was a little bit confused - it took some seconds to understand, that really each finding shall have the same severity in this test cases and that really each finding must have no validation

What about introducing some assert methods here ?

e.g.
```
for (Result finding : run.getResults()) {
    assertFindingHasSecHubSeverity("medium");
    assertFindingIsNotValidated();
}
```
Also you could think about having 2 tests here and separate the assertions...

sechub-wrapper-secret-validator/src/test/java/com/mercedesbenz/sechub/wrapper/secret/validator/execution/SecretValidatorExecutionContextFactoryTest.java
```diff
@@ -0,0 +1,158 @@
...
+        IllegalStateException exception = assertThrows(IllegalStateException.class, () -> factoryToTest.create());
+        assertEquals("Creating SARIF report model from: " + invalidFile + " failed!", exception.getMessage());
+    }
+
+    @Test
+    void not_existing_secret_validator_config_file_throws_exception() {
+        /* prepare */
+        // we need a valid Sarif report because the pds job result is read first
+        File validFile = new File("src/test/resources/config-test-files/valid-files/test-result.txt");
+        SecretValidatorPDSJobResult secretValidatorPDSJobResult = new SecretValidatorPDSJobResult();
```
de-jcup, line 77: Just a hint: You could mock the SecretValidatorPDSJobResult - here you test another implementation as well...

Also the SecretValidatorPDSJobResult could be set inside a field in a `@BeforeEach` method  would reduce boilerplate code

sechub-wrapper-secret-validator/src/test/java/com/mercedesbenz/sechub/wrapper/secret/validator/model/SecretValidatorConfigurationModelListTest.java
```diff
@@ -0,0 +1,55 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import org.junit.jupiter.api.Test;
+
+import com.mercedesbenz.sechub.commons.model.JSONConverter;
+
+class SecretValidatorConfigurationModelListTest {
+
+    @Test
+    void test() {
```
de-jcup, line 13:
```suggestion
    void json_converter_can_handle_model_in_expected_way() {
```

de-jcup (changes requested):
Good work, but I found some things to discuss.

Additionally: Even when we have no pds solution currently for the validation, we still could write an an integration test by enhancing existing secrect scan integration test (which just returns a sarif result) and call here the wrapper application from build folder (when a variant is set). So we could check this works together with SecHub at all.

sechub-wrapper-secret-validator/src/main/java/com/mercedesbenz/sechub/wrapper/secret/validator/model/SecretValidatorConfigurationModelList.java
```diff
@@ -0,0 +1,25 @@
...
+import java.util.List;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class SecretValidatorConfigurationModelList {
+
+    private List<SecretValidatorConfigurationModel> validatorConfigList = new ArrayList<>();
+
+    public List<SecretValidatorConfigurationModel> getValidatorConfigList() {
```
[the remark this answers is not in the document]
winzj: As far as I know jackson uses the setter if available. Only if no setter is implemented, the getter is used.

de-jcup (changes requested):
Looks good to me - found only some minor parts which should be changed.

de-jcup (approved):
Well done - looks good to me :+1:

PR #3308, opened 2024-07-15
cleanup empty files when remote-data-section #3307

### LINKED ISSUE #3307: Client creates empty .tar and .zip files when remote data is defined
## Situation 
When executing a remote scan, the client creates an empty binaries-projectname.tar file or an empty sources-projectname.zip file although none of these files are needed.
e.g. sourcecode-test-pds-prepare.zip or binaries-test-pds-prepare.tar 

## Wanted
No file should be created when scanning remote sources or binaries

## Solution
Remove file creation from client

sven-dmlr:
- closes #3307
- tests added


lorriborri (approved):
Works fine! Good Job! :star: :tiger:

PR #3315, opened 2024-07-22
Update gitleaks version and fix workaround for custom rules #3314

### LINKED ISSUE #3314: Update pds-gitleaks to gitleaks version 8.18.4 to fix custom rule workaround
### Situation
We currently use Gitleaks version `8.18.2`. using the custom rule file as documented on the official reporsitory results in `undefined` for the finding type. The reason is, only the rules in the custom file are part of the SARIF report even if the default file is extended correctly.

With the latest gitleaks version `8.18.4` the isssue described above was fixed.

### Wanted
Use the gitleaks custom rule feature as intended.

### Solution
- Update the  pds-gitleaks solution to gitleaks version `8.18.4`
- make changes to the pds-gitleaks solution to use the custom gitleaks rule file as intended and documented by gitleaks.

winzj:
#### Info:
The file `custom-gitleaks.toml`  got much smaller, with this PR, since it just extends the gitleaks defaults as intended, but now with the correct rule ids inside the report. For the review it might be easier to view the file from the branch because most of the changes are from deleting duplicates that are not necessary any more with the newest gitleaks version.

- closes #3314


sven-dmlr (approved):
Looks good to me :+1:

PR #3320, opened 2024-07-23
feat: implement gha caching for client download

### LINKED ISSUE #3199: Implement caching for GitHub Actions client download
## Situation
Currently the client (20mb) is downloaded every time the build pipeline runs because no caching is implemented.

## Wanted
- The client is cached by its version.
   For example: 1.6.1 is only downloaded once for every run

- Also the download should be logged properly so that the client knows that something is happening.
   Log the client version that is downloaded and progress etc.

tvcsantos:
Add support for caching client download.

Also did a cleanup on linter issues and some missing comments.

Closes #3199

### DIFF
--- a/.github/workflows/github-action-scan.yml
+++ b/.github/workflows/github-action-scan.yml
@@ -12,30 +12,30 @@ jobs:
   build-scan:
     runs-on: ubuntu-latest
     # Let's set the scan action folder as the working directory for all "run" steps:
-    
+
     defaults:
       run:
         working-directory: github-actions/scan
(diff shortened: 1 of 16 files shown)


hamidonos:
@tvcsantos I'm not sure if you gh actions run for pr inside forks
could you try renaming the branch to `gha_feature-implement-caching-gha` ? 
Hopefully the pipeline will be triggered then

tvcsantos:
@hamidonos renamed branch and PR available on #3323

PR #3343, opened 2024-08-07
Changed default time for accepted outdated encryption pool entry #3342

### LINKED ISSUE #3342: Change default value for allowance of outdated encryption pools on running servers
## Situation

With 
- #3250

a default time of 10 seconds is defined as allowed period of time of outdated pools on running servers  (startup is always terminated/never allow such a situation)

Problem: A K8s deployment can take a longer time - if the server just exits K8s could restart same (old SecHub) image again . 10 seconds are never a good default time for this.

## Wanted
We want to ensure that per default a K8s deployment will always work 

## Solution
We setup the default to 30 Minutes.

_Remark: it is not problematic that we take a long time here. The shutdown of the outdated servers and the corresponding time setup is only necessary to provide an ensured point in time where all cluster members MUST be in sync to have the possibility to do an auto cleanup and remove unused encryption pool entries without any race conditions. It is okay if old servers create jobs with older encryption pool entries in the meanwhile._

de-jcup:
This PR
- closes #3342


sven-dmlr (approved):
Looks good :+1:

PR #3404, opened 2024-09-03
Enable pull request hook in GitHub action workflow #3357

### LINKED ISSUE #3357: Enable "pull_request" hook in GitHub action workflow
## Situation

Currently workflow runs are not triggered if someone opens a PR on a fork.
This is because the hooks are only configured with the `push` event. 

image

## Wanted 

Runs should also be triggered when someone opens a PR on SecHub (e.g. PR from a Fork)

## Solution

git hub actions yaml
image

gradle yaml
image

From the official docs:

_Runs your workflow anytime the pull_request event occurs. More than one activity type triggers this event. For information about the REST API, see https:// developer. github. com/ v3/ pulls. Note: Workflows do not run on private base repositories when you open a pull request from a forked repository. When you create a pull request from a forked repository to the base repository, GitHub sends the pull_request event to the base repository and no pull request events occur on the forked repository. Workflows don't run on forked repositories by default. You must enable GitHub Actions in the Actions tab of the forked repository. The permissions for the GITHUB_TOKEN in forked repositories is read-only. For more information about the GITHUB_TOKEN, see https:// help. github. com/ en/ articles/ virtual-environments-for-github-actions_

hamidonos:
With this GitHub Actions workflows will run for:
- direct pushes to `main` or `develop`
- pull requests that are opened and target `main` or `develop`

Please note that it's not possible to exclude source branches inside the `pull_request` hook. That's why I chose to implement the skip of certain branches inside the job definition.

- closes #3357

### DIFF
--- a/.github/workflows/github-action-scan.yml
+++ b/.github/workflows/github-action-scan.yml
@@ -10,16 +10,12 @@ on:
     branches:
       - 'main'
       - 'develop'
-    # This specifies the source branches to exclude from triggering the workflow
-    # We don't want to trigger this workflow for feature increments
-    branches-ignore:
-      - 'feature_*'
-      - 'feature-*'
   # enable manual triggering of workflow
   workflow_dispatch:
 
 jobs:
   build-scan:
+    if: "startsWith(github.head_ref, 'gha_') || startsWith(github.head_ref, 'gha-')"
     runs-on: ubuntu-latest
     # Let's set the scan action folder as the working directory for all "run" steps:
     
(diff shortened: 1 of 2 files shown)


.github/workflows/github-action-scan.yml
```diff
@@ -4,12 +4,20 @@ name: Build SecHub GHA (scan)
 on:
```
lorriborri, line 4: Please rename your branch according to https://mercedes-benz.github.io/sechub/latest/sechub-techdoc.html#branching-model (I opened an issue to transfer the documentation to the developer guide or contributing.md)
We can also discuss about a better naming convention, but it should be consistent.

lorriborri (approved):
Looks Good to me! Well done! :star: 🐡

PR #3428, opened 2024-09-17
helm chart extended by env variables #3427

### LINKED ISSUE #3427: Make it possible to set environment variables for the webui via helmchart
Wanted:
It should be possible to set environment variables for the SecHub web ui via the helmchart.

sven-dmlr:
- closes #3427


haerter-tss (approved):
Looks good!

PR #3440, opened 2024-09-19
Trigger build workflows based on changed files #3439

### LINKED ISSUE #3439: Trigger build workflows based on changed files
## Situation
Currently the workflows
- "Java & Go CI" (build Java + Go and run integration tests)
- "Build SecHub GHA (scan)" (build scan Action and run its unit tests)
are not working properly (commits to `master`, `develop`, ... never trigger a run but they should)

Additionally both workflows are started even if they aren't suppoed to run (skipped by `if: ` statement.
This looks a bit confusing.

## Wanted
The build workflows are triggered based on the change's (Pr or commit) content.

## Solution
Use `paths:` and `paths-ignore`
instead of `branches` plus `if:`.

See https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

Also the docs can be updated: There is no more need for `gha_feature-xyz` branch names.

sven-dmlr:
- closes #3439

### DIFF
--- a/.github/workflows/github-action-scan.yml
+++ b/.github/workflows/github-action-scan.yml
@@ -3,27 +3,16 @@ name: Build SecHub GHA (scan)
 
 on:
   push:
-    branches:
-      - main
-      - master
-      - develop
-      - hotfix
+    paths:
+      - 'github-actions/**'
   pull_request:
-    # This workflow will run for all pull requests that target following branches
-    branches:
-      - main
-      - master
-      - develop
-      - hotfix
+    paths:
+      - 'github-actions/**'
   # enable manual triggering of workflow
   workflow_dispatch:
 
 jobs:
   build-scan:
-    # This job is only executed if the branch name starts with 'gha_feature-' or 'dependabot/'
-    # or equals 'main', 'master', 'develop' 'hotfix'
-    if: ${{ startsWith(github.head_ref, 'gha_feature-') || startsWith(github.head_ref, 'dependabot/') ||
-      github.head_ref == 'main' || github.head_ref == 'master' || github.head_ref == 'develop' || github.head_ref == 'hotfix' }}
     runs-on: ubuntu-latest
     # Let's set the scan action folder as the working directory for all "run" steps:
     
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -3,21 +3,24 @@ name: Java & Go CI
 
 on:
   push:
-    branches:
-      - main
-      - master
-      - develop
-      - hotfix
+    paths-ignore:
+      - 'docs/**'
+      - 'github-actions/**'
+      - 'sechub-doc/**'
+      - 'sechub-website/**'
+      - 'sechub-webui/**'
+      - 'sechub-webui-solution/**'
     # We ignore everything where tag starts with v* - this is done by release build!
     tags-ignore:
       - v*
   pull_request:
-    # This workflow will run for all pull requests that target following branches
-    branches:
-      - main
-      - master
-      - develop
-      - hotfix
+    paths-ignore:
+      - 'docs/**'
+      - 'github-actions/**'
+      - 'sechub-doc/**'
+      - 'sechub-website/**'
+      - 'sechub-webui/**'
+      - 'sechub-webui-solution/**'
     tags-ignore:
       - v*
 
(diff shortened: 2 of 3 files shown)


.github/workflows/gradle.yml
```diff
@@ -3,21 +3,24 @@ name: Java & Go CI
...
-      - main
-      - master
-      - develop
-      - hotfix
+    paths-ignore:
+      - 'docs/**'
+      - 'github-actions/**'
+      - 'sechub-doc/**'
+      - 'sechub-website/**'
+      - 'sechub-webui/**'
```
hamidonos, line 22: Should the `sechub-webui` really be excluded? It's a deployable Spring Boot app like the others

.github/workflows/gradle.yml
```diff
@@ -3,21 +3,24 @@ name: Java & Go CI
...
```
hamidonos, line 11: Should the sechub-webui really be excluded? It's a deployable Spring Boot app like the others

.github/workflows/github-action-scan.yml
```diff
@@ -3,27 +3,16 @@ name: Build SecHub GHA (scan)
 
 on:
   push:
-    branches:
-      - main
-      - master
-      - develop
-      - hotfix
+    paths:
+      - 'github-actions/**'
```
hamidonos, line 7: If we ever add any other action to `github-actions` directory this workflow will run for that action as well
probably won't happen but maybe it's better to set it to `github-actions/scan/**` (just to be sure in the future)

.github/workflows/github-action-scan.yml
```diff
@@ -3,27 +3,16 @@ name: Build SecHub GHA (scan)
...
   pull_request:
-    # This workflow will run for all pull requests that target following branches
-    branches:
-      - main
-      - master
-      - develop
-      - hotfix
+    paths:
+      - 'github-actions/**'
```
hamidonos, line 10: same here (if comment above is applicable)

hamidonos:
Something is working properly 
Your PR did not execute any workflow
It should have executed the `gradle.yml` workflow considering your branch name starts with `feature-`

.github/workflows/gradle.yml
sven-dmlr: removed sechub-webui from excludes

.github/workflows/github-action-scan.yml
sven-dmlr: Hmm. I thought that if we have more Actions in the future, we would have one workflow for them all.
(Maybe rename this one...)

.github/workflows/gradle.yml
sven-dmlr: Good point :+1:  - suggestion added.

sven-dmlr:
> When using the `paths` parameter instead of `branches` you should always include the workflow files in the paths so that changes to the actual workflows trigger the run(s)

Good point 👍 - suggestion added.

.github/workflows/gradle.yml
sven-dmlr: sechub-webui removed from the exclude list

.github/workflows/github-action-scan.yml
hamidonos: I don't think that makes sense.

If you take a look at the at the workflow file you'll see that it's actually tailored to the source code of the `github-actions/scan` directory.

I'd rather stick to my suggestion from above

.github/workflows/github-action-scan.yml
```diff
@@ -3,27 +3,17 @@ name: Build SecHub GHA (scan)
...
 on:
   push:
-    branches:
-      - main
-      - master
-      - develop
-      - hotfix
+    paths:
+      - 'github-actions/**'
+      - '.github/workflows/github-action*'
```
hamidonos, line 8: why should this workflow run if changes to other workflow files are made?

I think you should just remove the wildcard and set the full name instead

.github/workflows/gradle.yml
```diff
@@ -3,21 +3,20 @@ name: Java & Go CI
...
   pull_request:
-    # This workflow will run for all pull requests that target following branches
-    branches:
-      - main
-      - master
-      - develop
-      - hotfix
+    paths-ignore:
+      - 'docs/**'
+      - 'github-actions/**'
```
hamidonos, line 17: this is problematic
since you made changes to `gradle.yml` this PR should also execute the Java Build Pipeline, but it doesn't
This is because `github-actions/**` ignores changes to all workflow files

hamidonos (commented):
Something

.github/workflows/github-action-scan.yml
sven-dmlr: full path and name added

.github/workflows/github-action-scan.yml
sven-dmlr: full path and name added

.github/workflows/github-action-scan.yml
sven-dmlr: done

.github/workflows/gradle.yml
sven-dmlr: The workflow files are under `.github`.
Unfortunately this path (or everything starting with a `.` seems to be ignored by default. :disappointed: 
Changed the filter so the gradle.yml is included...

de-jcup (approved):
Looks good to me - ready to merge :rocket:

PR #3449, opened 2024-09-24
Improved GHA documentation

haerter-tss:
Added section that explains variables and config files better.

### DIFF
--- a/github-actions/scan/README.adoc
+++ b/github-actions/scan/README.adoc
@@ -81,6 +81,16 @@ To be able to use this action you need a SecHub project. Check the https://merce
     fail-job-with-findings: true
 ----
 
+*Note:* The following variables take priority over the configuration file:
+
+- `USERID`
+- `SERVER`
+- `PROJECT`
+- `DEBUG`
+- `TRUSTALL`
+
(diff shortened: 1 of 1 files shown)


github-actions/scan/README.adoc
```diff
@@ -81,6 +81,16 @@ To be able to use this action you need a SecHub project. Check the https://merce
     fail-job-with-findings: true
 ----
 
+*Note:* The following variables take priority over the configuration file:
```
sven-dmlr, line 84: I suggest to use an IMPORTANT section here:
```suggestion
[IMPORTANT]
====
The following variables take priority over the configuration file:
...
====
```

github-actions/scan/README.adoc
```diff
@@ -81,6 +81,16 @@ To be able to use this action you need a SecHub project. Check the https://merce
...
+
+- `USERID`
```
sven-dmlr, line 86: The user cannot set this variable directly - so this note will not help.
You should mention above parameters. In this case:
```suggestion
- `user:`
```

sven-dmlr (changes requested):
Please check my comments

sven-dmlr (approved):
Looks good to me :+1: :eyeglasses:

PR #3452, opened 2024-09-24
fixed windows client exclude handling #3447

### LINKED ISSUE #3447: SecHub Windows Client Includes Zip File in Compression, Causing Endless Loop
### Situation
When doing a scan for secrets ("secretScan") the SecHub client compresses source code into a zip file. When including "." (the folder you are currently in), the SecHub Linux client successfully excludes the zip file from the compression and works as expected. However, when performing the same action with the Windows SecHub client, the zip file itself is included in the compression, resulting in an endless loop.

Example sechub.json:
```json
{
  "apiVersion": "1.0",
  "secretScan": { "use": [ "myscan" ] },
  "data": {
    "sources": [
      {
        "name": "myscan",
        "fileSystem": { "folders": [ "." ] }
      }
    ]
  }
}
```

### Wanted
It is desired that this bug is investigated and fixed. 

### Solution
A potential solution could involve checking if the exclude pattern does not work on Windows for some reason. Alternatively, it might be that the temporary zip file on Windows has a different name or extension.

sven-dmlr:
- closes #3447

### DIFF
--- a/sechub-cli/src/mercedes-benz.com/sechub/cli/config.go
+++ b/sechub-cli/src/mercedes-benz.com/sechub/cli/config.go
@@ -447,7 +447,7 @@ func validateTempDir(config *Config) bool {
 func validateOutputLocation(config *Config) bool {
 	if config.outputLocation == "" || config.outputLocation == "." {
 		config.outputFolder = "."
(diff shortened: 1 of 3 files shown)


haerter-tss (approved):
Looks good! 🐳🎈

PR #3456, opened 2024-09-25
client shall not follow symlinks #3367

### LINKED ISSUE #3367: Client shall not follow symlinks
## Situation
Currently the client follows symbolic links.

This causes potential problems:
- If the symlink points to an existing file then it will be scanned twice.
- The client exits with error when the symlink does not point to an existing file
- The symlink could point to anywhere in the filesystem. This can be a security issue

## ToDo
- Check the file type before zipping and ignore symbolic links.
- Check the file type before tar'ing and ignore symbolic links.
- In both cases print a notice to the user

sven-dmlr:
Also replaced some occurences of deprecated ioutil calls.

- closes #3367

### DIFF
--- a/sechub-cli/src/mercedes-benz.com/sechub/testutil/fileTestHelpers.go
+++ b/sechub-cli/src/mercedes-benz.com/sechub/testutil/fileTestHelpers.go
@@ -3,8 +3,8 @@
 package util
 
 import (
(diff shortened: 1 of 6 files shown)


haerter-tss (approved):
Looks good 🎀

PR #3457, opened 2024-09-25
added support for Pro*C file extensions #3271

### LINKED ISSUE #3271: Add Support for Pro*C (.pc) File Extensions in SecHub Client
### Situation
The SecHub client is currently unable to process Pro*C files with the .pc extension. This limitation will become an issue since the future update of Checkmarx will introduce support for .pc files.

### Wanted
The SecHub client should be updated to allow users to upload .pc files.

### Solution
To achieve the desired functionality, the following changes are proposed:
Add `.pc` to "sechub-cli/script/supported-source-extensions.txt" so the client will support the upload.

sven-dmlr:
- closes #3271


haerter-tss (approved):
Looks good 🍪

PR #3473, opened 2024-09-30
Feature implement o auth for sec hub web UI #3406

### LINKED ISSUE #3406: Implement OAuth for SecHub WebUI
## Situation

Currently we only use Basic Auth to authenticate in SecHub.

## Wanted

We want to provide users with a more sophisticated authentication solution.

## Solution

Implement OAuth inside SecHub WebUI.

### LINKED ISSUE #3505: Implement OAuth2 for WebUI Application
## Situation

Currently we only use Basic Auth to authenticate in SecHub.

## Wanted

We want to provide users with a more sophisticated authentication solution.

## Solution

Implement OAuth inside SecHub WebUI.

hamidonos:
- closes #3505


sechub-webui/src/main/java/com/mercedesbenz/sechub/webui/security/SecurityConfiguration.java
```diff
@@ -1,49 +1,117 @@
...
+                .tokenEndpoint(token -> token.accessTokenResponseClient(base64EncodedClientIdAndSecretOAuth2AccessTokenClient))
+                .successHandler(authenticationSuccessHandler));
+        }
+
+        /*
+            Enable Form Login
+            Note: This must be the last configuration in order to set the default 'loginPage' to oAuth2
+            because spring uses the 'loginPage' from the first authentication method configured
+        */
+        httpSecurity
```
de-jcup, line 102: Should this not be inside a 
```java
if (!isOAuth2Enabled()) {
   ///httpSecurity... for login page with CLASSIC...
}
```
?

sechub-webui/src/main/resources/application-basic-auth-mocked.yml
```diff
@@ -6,3 +6,9 @@ webui:
     ## Mocked sechub user (necessary for credentialinjection, we have no defaults...)
     userid: "mocked-user"
     apitoken: "mocked-apitoken"
+
+spring:
+  security:
+    user:
+      name: sechub
```
de-jcup, line 13: I would prefer name and password where it is clear on first sight what the purpose of it is. Does also minimimze problems with secret scanners etc.

e.g. "mock-user" or "test-user" etc.

sechub-webui/src/test/java/com/mercedesbenz/sechub/webui/YamlPropertyLoaderFactory.java
```diff
@@ -0,0 +1,34 @@
...
+ * properties from YAML files, which is often more convenient and readable for
+ * complex configurations.
+ *
+ * @author hamidonos
+ */
+public class YamlPropertyLoaderFactory extends DefaultPropertySourceFactory {
+
+    @Override
+    public PropertySource<?> createPropertySource(@Nullable String name, EncodedResource resource) throws IOException {
+        Resource actualResource = resource.getResource();
```
de-jcup, line 26: Hm... if the actual resource exists... shouldn't we check if the resource name ends with ".yml" or ".yaml" as well - otherwise it could happen IMHO that this loader factory would try to load also other kind of resources - and fail (but maybe I am wrong).

Ah, have seen it now in usage: its only for tests and defined directly inside tests... so normally not problematic.

sechub-webui/src/test/java/com/mercedesbenz/sechub/webui/YamlPropertyLoaderFactory.java
```diff
@@ -0,0 +1,34 @@
...
+        if (actualResource.exists()) {
+            YamlPropertySourceLoader loader = new YamlPropertySourceLoader();
+            List<PropertySource<?>> propertySources = loader.load(name != null ? name : actualResource.getFilename(), actualResource);
```
de-jcup, line 29: Hm... is the actualResource.getFilename() always the same as name, or it is only an acceptable fallback? Depending if the name is set or not, the load is called with different names - is this intended?

sechub-webui/src/test/java/com/mercedesbenz/sechub/webui/page/LoginClassicControllerTest.java
```diff
@@ -0,0 +1,37 @@
...
+    @Autowired
+    LoginClassicControllerTest(MockMvc mockMvc) {
+        this.mockMvc = mockMvc;
+    }
+
+    @Test
+    void login_classic_page_is_accessible_anonymously() throws Exception {
+        /* @formatter:off */
+        mockMvc
+                .perform(get("/login/classic"))
```
de-jcup, line 33: Login classic seems to be always possible - but shouldn't . 
See my comments before about `@Profile("!"+...) ` or giving own profile for classic.

sechub-webui/src/test/java/com/mercedesbenz/sechub/webui/page/LoginOAuth2ControllerTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+    private final MockMvc mockMvc;
+
+    @Autowired
+    LoginOAuth2ControllerTest(MockMvc mockMvc) {
+        this.mockMvc = mockMvc;
+    }
+
+    @Test
+    void login_o_auth_2_page_is_accessible_anonymously() throws Exception {
```
de-jcup, line 32: Should we test also (in another test method) that login in classic mode is not supported in this case?

sechub-webui/src/test/java/com/mercedesbenz/sechub/webui/security/Base64EncodedClientIdAndSecretOAuth2AccessTokenClientTest.java
```diff
@@ -0,0 +1,187 @@
...
+                .andExpect(header(HttpHeaders.CONTENT_TYPE, Constants.APPLICATION_FORM_URLENCODED_VALUE))
+                .andExpect(header(HttpHeaders.AUTHORIZATION, authorizationHeaderValue))
+                .andExpect(content().formData(getMultiValueMap(authorizationCodeGrantRequest)))
+                .andRespond(withSuccess(jwtResponseJson, MediaType.APPLICATION_JSON));
+        // @formatter:on
+
+        // execute
+        client.getTokenResponse(authorizationCodeGrantRequest);
+
+        // verify
```
de-jcup, line 103: Please replace the "// verify" parts with "// test" . We have everywhere (Sechub server, PDS, go client, plugins, wrapper applications, github-action) the combinations
```java
// prepare
// check preconditions
// execute
// test
```
I would prefer not to have a new version for "// test"

```suggestion
        // test
```

de-jcup (changes requested):
Good job - but I still found some parts I want to discuss/to be changed.

sechub-webui/src/main/java/com/mercedesbenz/sechub/webui/credentials/NewApiTokenController.java
```diff
@@ -1,26 +1,26 @@
...
 import org.springframework.web.bind.annotation.GetMapping;
 
 import com.mercedesbenz.sechub.webui.RequestConstants;
-import com.mercedesbenz.sechub.webui.page.user.UserInfoService;
 import com.mercedesbenz.sechub.webui.sechubaccess.SecHubAccessService;
+import com.mercedesbenz.sechub.webui.user.UserInfoService;
 
 @Controller
-public class NewApiTokenController {
+class NewApiTokenController {
```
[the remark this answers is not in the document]
hamidonos: Spring can instantiate package-private classes using reflection without any problems.

If a package-private bean is managed by the IoC container and used by classes in the same package it's no problem. Problems only arise when you try to wire that bean across packages. Which is quite obvious of course.

I like to use the minimum needed visibility on my classes
Making the class package private enforces test classes to be in the same package which I think is beneficial

Why should a controller be public?

de-jcup (approved):
Good job - I found 2 minor things, but I think it is up to you if you change it or not.

PR #3480, opened 2024-10-02
Add totp to sechub webscan login config #3471

### LINKED ISSUE #3471: Extend the sechub webscan configuration with TOTP data
### Situation
Many web applications use multi factor authentication with [TOTP](https://datatracker.ietf.org/doc/html/rfc6238) because it is easy to use via authenticator apps on a smartphone. The SecHub DAST module shall support TOTP as an second authentication factor as well, because  is a widely known and a common approach to add an additional authentication factor.

### Wanted
Integrate a TOTP configuration in the SecHub webscan configuration model:

```json
{
"apiVersion": "1.0",
"webScan": {
   "url": "https://example.com",
   "login": {
      "url": "https://example.com/login",
      "totp": {
         "seed": "{{ .TOTP_SEED }}",
         "validityInSeconds": "{{ .TOTP_VALIDITY }}",
         "tokenLength": "{{ .TOTP_LENGTH }}",
         "hashAlgorithm": "{{ .TOTP_ALGORITHM }}"
    }
  }
}
}
```

### Solution
- Integrate the configuration model and add the model to the openapi.yaml file (necessary until we use the generated models)
- Add documentation via RestDocTest
- It must be possible to only provide the `seed` as a mandatory field. If only the `seed` is provided the most common default configuration will be used, which is:
  - "validityInSeconds": 30
  - "tokenLength": 6
  - "hashAlgorithm": HMAC_SHA1

winzj:
- closes #3471

### DIFF
--- a/sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/model/login/WebLoginTOTPConfiguration.java
+++ b/sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/model/login/WebLoginTOTPConfiguration.java
@@ -1,0 +1,64 @@
+// SPDX-License-Identifier: MIT
+package com.mercedesbenz.sechub.commons.model.login;
+
+import javax.crypto.SealedObject;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.mercedesbenz.sechub.commons.core.security.CryptoAccess;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class WebLoginTOTPConfiguration {
+
+    public static final String PROPERTY_SEED = "seed";
+    public static final String PROPERTY_VALIDITY_IN_SECONDS = "validityInSeconds";
+    public static final String PROPERTY_TOKEN_LENGTH = "tokenLength";
+    public static final String PROPERTY_HASH_ALGORITHM = "hashAlgorithm";
+
+    private static final int DEFAULT_VALIDITY_IN_SECONDS = 30;
+    private static final int DEFAULT_TOKEN_LENGTH = 6;
+    private static final TOTPHashAlgorithm DEFAULT_HASH_ALGORITHM = TOTPHashAlgorithm.HMAC_SHA1;
+
+    private SealedObject seed;
+    private int validityInSeconds;
+    private int tokenLength;
+    private TOTPHashAlgorithm hashAlgorithm;
+
+    public WebLoginTOTPConfiguration() {
+        this.validityInSeconds = DEFAULT_VALIDITY_IN_SECONDS;
+        this.tokenLength = DEFAULT_TOKEN_LENGTH;
+        this.hashAlgorithm = DEFAULT_HASH_ALGORITHM;
+    }
+
+    public String getSeed() {
+        return CryptoAccess.CRYPTO_STRING.unseal(seed);
+    }
+
+    public void setSeed(String seed) {
+        this.seed = CryptoAccess.CRYPTO_STRING.seal(seed);
+    }
+
+    public int getValidityInSeconds() {
+        return validityInSeconds;
+    }
+
+    public void setValidityInSeconds(int validityInSeconds) {
+        this.validityInSeconds = validityInSeconds;
+    }
+
+    public int getTokenLength() {
+        return tokenLength;
+    }
+
+    public void setTokenLength(int tokenLength) {
+        this.tokenLength = tokenLength;
+    }
+
+    public TOTPHashAlgorithm getHashAlgorithm() {
+        return hashAlgorithm;
+    }
+
+    public void setHashAlgorithm(TOTPHashAlgorithm hashAlgorithm) {
+        this.hashAlgorithm = hashAlgorithm;
+    }
+
+}
--- a/sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/model/login/WebLoginTOTPConfiguration.java
+++ b/sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/model/login/WebLoginTOTPConfiguration.java
@@ -14,9 +14,9 @@ public class WebLoginTOTPConfiguration {
     public static final String PROPERTY_TOKEN_LENGTH = "tokenLength";
     public static final String PROPERTY_HASH_ALGORITHM = "hashAlgorithm";
 
-    private static final int DEFAULT_VALIDITY_IN_SECONDS = 30;
-    private static final int DEFAULT_TOKEN_LENGTH = 6;
-    private static final TOTPHashAlgorithm DEFAULT_HASH_ALGORITHM = TOTPHashAlgorithm.HMAC_SHA1;
+    public static final int DEFAULT_VALIDITY_IN_SECONDS = 30;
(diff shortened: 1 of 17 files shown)


sechub-doc/src/docs/asciidoc/documents/shared/configuration/sechub_config_example21_webscan_login_form_with_totp.json
```diff
@@ -0,0 +1,16 @@
+{
+  "apiVersion" : "1.0",
+  "webScan" : {
+    "url" : "https://productfailure.demo.example.org",
+    "login" : {
+      "url" : "https://productfailure.demo.example.org/login",
+      "form" : { ... },
```
de-jcup, line 7: Please either
- remove the form element or 
- create empty element with "//..." part inside (should be accepted by output formatter)
- add a simple example there.

Reason: I did not check it, but the output formatter in asciidoc output should have probllems here, because it is not valid json

sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/model/login/WebLoginTOTPConfiguration.java
```diff
@@ -0,0 +1,64 @@
...
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class WebLoginTOTPConfiguration {
+
+    public static final String PROPERTY_SEED = "seed";
+    public static final String PROPERTY_VALIDITY_IN_SECONDS = "validityInSeconds";
+    public static final String PROPERTY_TOKEN_LENGTH = "tokenLength";
+    public static final String PROPERTY_HASH_ALGORITHM = "hashAlgorithm";
+
+    private static final int DEFAULT_VALIDITY_IN_SECONDS = 30;
```
de-jcup, line 17: We could make these default values public and use it inside the documentation (e.g. RESTDOC tests). If we would change the defaults at this point the documentation would always be valid.

sechub-commons-model/src/test/java/com/mercedesbenz/sechub/commons/model/login/WebLoginTOTPConfigurationTest.java
```diff
@@ -0,0 +1,48 @@
...
+        /* execute */
+        String json = JSONConverter.get().toJSON(expectedConfig);
+        WebLoginTOTPConfiguration config = JSONConverter.get().fromJSON(WebLoginTOTPConfiguration.class, json);
+
+        /* test */
+        assertEquals(config.getSeed(), expectedConfig.getSeed());
+        assertEquals(config.getValidityInSeconds(), expectedConfig.getValidityInSeconds());
+        assertEquals(config.getTokenLength(), expectedConfig.getTokenLength());
+        assertEquals(config.getHashAlgorithm(), expectedConfig.getHashAlgorithm());
+    }
```
de-jcup, line 26: Hm... what if "default" values would be null here... would not exist
please make the default constants public (as mentioned before) and check like this:

```java 
assertEquals(config.getSeed(), null); // there is no default
assertEquals(config.getTokenLength(), WebLoginTOTPConfiguration.DEFAULT_TOKEN_LENGTH); 
...
```

sechub-doc/src/docs/asciidoc/documents/shared/configuration/sechub_config_example21_webscan_login_form_with_totp.json
de-jcup: Additionally: we have currently no unit test which reads a sechub configuration with totp setup inside. It would be nice, if this document file would be read by a test and checked that it contains the expected totp information and that it is valid json.

if it is not... the documentation would contain a invalid example file

So please add a unit test which reads the file and does some simple checks.

sechub-doc/src/docs/asciidoc/documents/shared/configuration/sechub_config.adoc
```diff
@@ -300,6 +300,35 @@ WARNING: `username` and `password` are like input but {sechub} tries
          to hide these information in logs or at UI where possible.
          So do *NOT* use `input` for credentials!
 
+====== Example form based login with TOTP
+The TOTP configuration is generally never used without any other login mechanism. It is meant to be used as an additional authentication factor.
+In this example TOTP is used in combination with form based login, which is left out in the example for better readability.
+
+TIP: Most of the time it is enough to specify a `seed` for the TOTP configuration in SecHub.
```
de-jcup, line 307: We have some variables to avoid naming failures etc.
Please use {sechub} variable here:

```suggestion
TIP: Most of the time it is enough to specify a `seed` for the TOTP configuration in {sechub}.
```

sechub-doc/src/docs/asciidoc/documents/shared/configuration/sechub_config.adoc
```diff
@@ -300,6 +300,35 @@ WARNING: `username` and `password` are like input but {sechub} tries
...
+[source,json]
+----
+include::sechub_config_example21_webscan_login_form_with_totp.json[]
+----
+<1> Start of TOTP configuration.
+<2> The `seed` is the only __mandatory__ field for the TOTP configuration.
+It represents the secret key used to generate TOTP values.
+<3> The `validityInSecods` is an __optional__ field, which represents the maximum amount of seconds a TOTP is valid.
+This depends on the authentication provider or the application, providing the `seed` to the user.
+If no specific `validityInSecods` is provided do not configure this value and SecHub will use the default.
```
de-jcup, line 320:
```suggestion
If no specific `validityInSecods` is provided do not configure this value and {sechub} will use the default.
```

sechub-doc/src/docs/asciidoc/documents/shared/configuration/sechub_config.adoc
```diff
@@ -300,6 +300,35 @@ WARNING: `username` and `password` are like input but {sechub} tries
...
+<4> The `tokenLength` is an __optional__ field, which represents the length of the TOTP the authentication provider or the application expects.
+As before, this depends on the authentication provider or the application, providing the `seed` to the user.
+If no specific `tokenLength` is provided do not configure this value and SecHub will use the default.
```
de-jcup, line 323:
```suggestion
If no specific `tokenLength` is provided do not configure this value and {sechub} will use the default.
```

sechub-doc/src/docs/asciidoc/documents/shared/configuration/sechub_config.adoc
```diff
@@ -300,6 +300,35 @@ WARNING: `username` and `password` are like input but {sechub} tries
...
+<5> The `hashAlgorithm` is an __optional__ field, which represents hash algorithm the authentication provider or the application uses during the computation of the TOTP.
```
de-jcup, line 324:
```suggestion
<5> The `hashAlgorithm` is an __optional__ field, representing the hash algorithm which is used by the authentication provider or the application during the computation of the TOTP.
```

de-jcup (changes requested):
Good work - i found some smaller parts, see my comments.

de-jcup (approved):
LGTM - ready for merge :-)

PR #3510, opened 2024-10-14
Sanitize GitHub action shell cmd to prevent cmd injection #3509

### LINKED ISSUE #3509: Sanitize GitHub Action input to prevent command injection
## Situation

Currently there is no safe guard to prevent command injection when executing the `scan` action inside the GitHub action.

## Wanted

Command injection should not be possible under any circumstances.

## Solution

- Sanitize the shell command before executing it
- Check against a whitelist of characters
- If any character is detected that is not part of the whitelist throw an error and interrupt the process
- `/^[a-zA-Z0-9_\- ]+$/`

hamidonos:
- closes #3509

### We use `execFileSync` to execute shell commands in a a separate child process.

**Separate Process:**
The child process created by execFileSync runs independently from the main Node.js process. It does not share the same memory or execution context with the parent process, so they do not interfere with each other directly.
If the child process crashes or exits, it won't directly crash the parent Node.js process.

**Controlled Input and Output:**
The child process has its own stdin, stdout, and stderr streams, and the parent process can choose to interact with them (read/write), but they are isolated unless explicitly connected.
By default, execFileSync captures the output of the child process and returns it to the parent process only after execution completes.

**Environment Variables:**
The child process runs with its own set of environment variables. You can specify a custom environment for the child process, which is separate from the environment of the parent process.

Every argument is passed separately.

### Every shell argument is sanitized independently

Before we pass shell commands and arguments to the child process we sanitize the input parameters.
Sanitization checks for illegal characters and if a argument is a executable command -> if so we throw an error and exit the program


de-jcup (approved):
Very good work - let's merge it :+1:

PR #3512, opened 2024-10-14
Feature 3507 introduce x sechub dast header

### LINKED ISSUE #3507: Introduce x-sechub-dast http header for webscans
### Situation
When someone performs a webscan using SecHub looking at the scanned applications logs, it might be difficult to tell the difference between an outside party performing potentially malicious requests or the SecHub webscan.

### Wanted
It should be easy to identify SecHub webscan requests.

### Solution
With each request a custom http header will be sent to identify the SecHub scan. This way it is easier to identify the SecHub webscan requests in application logs. For this we can introduce a header called `x-sechub-dast`, like in the following example:
-`x-sechub-dast : SecHub DAST job: $JOB_UUID`

If this header contains the jobUUID like in the example above, the scan job can be identified.

### LINKED ISSUE #3511: Improve SecHub DAST user messages for scanned URLs
### Situation
Currently we send a list of URLs as user messages, which appear at the bottom of the report. The problem is, we do not provide information if the URL was accessible. There is another way to retrieve information about URLs detected during a scan, with additional data, like HTTP status code.
This would give everyone using the SecHub webscan more information about what happend, e.g. if the requests returned something like 401 Unauthorized, the login or session handling did not work as expected.

### Solution
Add more information to the user messageds like
- the URL
- the HTTP method used
- the HTTP status code received

winzj:
- closes #3507 
- closes #3511

### DIFF
--- a/sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/zapwrapper/helper/ZapProductMessageHelper.java
+++ b/sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/zapwrapper/helper/ZapProductMessageHelper.java
@@ -3,11 +3,9 @@ package com.mercedesbenz.sechub.zapwrapper.helper;
 
 import java.io.IOException;
 import java.util.List;
-import java.util.Map;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.zaproxy.clientapi.core.ApiResponse;
 
 import com.mercedesbenz.sechub.commons.TextFileWriter;
 import com.mercedesbenz.sechub.commons.model.SecHubMessage;
@@ -60,26 +58,6 @@ public class ZapProductMessageHelper {
         }
     }
 
-    public void writeUserMessagesWithDetectedURLs(List<Map<String, ApiResponse>> results) {
-        // The map looks like this: {processed=true, statusReason=OK, method=GET,
-        // reasonNotProcessed=, messageId=6, url=http://example.com, statusCode=200}
-        for (Map<String, ApiResponse> result : results) {
-            String url = result.get("url").toString();
-            // robots.txt and sitemap.xml always appear inside the sites tree even if they
-            // are not available. Because of this it is skipped here.
-            if (url.contains("robots.txt") || url.contains("sitemap.xml")) {
-                continue;
-            }
-            String statusCode = result.get("statusCode").toString();
-            String statusReason = result.get("statusReason").toString();
-            String method = result.get("method").toString();
-
-            String message = "URL: '%s' returned status code: '%s/%s' on detection phase for request method: '%s'".formatted(url, statusCode, statusReason,
-                    method);
-            writeSingleProductMessage(new SecHubMessage(SecHubMessageType.INFO, message));
-        }
-    }
-
     private void writeProductErrorForExitCode(ZapWrapperExitCode exitCode) throws IOException {
         if (exitCode == null) {
             return;
@@ -92,7 +70,7 @@ public class ZapProductMessageHelper {
             break;
         case API_DEFINITION_CONFIG_INVALID:
             productMessageSupport.writeMessage(new SecHubMessage(SecHubMessageType.ERROR,
-                    "Please use files instead of folders for the API definition with the filesystem->files section of the SecHub configuration."));
+                    "Please check your webscan api section inside the sechub configuration file. Only use supported api types and use filesystem->files instead of filesystem->folders for the API definition files."));
             break;
         case TARGET_URL_INVALID:
             productMessageSupport.writeMessage(new SecHubMessage(SecHubMessageType.ERROR,
--- a/sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/zapwrapper/helper/ZapProductMessageHelper.java
+++ b/sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/zapwrapper/helper/ZapProductMessageHelper.java
@@ -92,7 +92,7 @@ public class ZapProductMessageHelper {
             break;
         case API_DEFINITION_CONFIG_INVALID:
             productMessageSupport.writeMessage(new SecHubMessage(SecHubMessageType.ERROR,
-                    "Please files instead of folders for the API definition inside the filesystem->files section of the SecHub configuration."));
+                    "Please use files instead of folders for the API definition with the filesystem->files section of the SecHub configuration."));
             break;
         case TARGET_URL_INVALID:
             productMessageSupport.writeMessage(new SecHubMessage(SecHubMessageType.ERROR,
--- a/sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/zapwrapper/helper/ZapProductMessageHelper.java
+++ b/sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/zapwrapper/helper/ZapProductMessageHelper.java
@@ -3,9 +3,11 @@ package com.mercedesbenz.sechub.zapwrapper.helper;
 
 import java.io.IOException;
 import java.util.List;
+import java.util.Map;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.zaproxy.clientapi.core.ApiResponse;
 
 import com.mercedesbenz.sechub.commons.TextFileWriter;
 import com.mercedesbenz.sechub.commons.model.SecHubMessage;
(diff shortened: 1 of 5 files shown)


sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/zapwrapper/helper/ZapProductMessageHelper.java
```diff
@@ -58,14 +60,23 @@ public void writeProductError(ZapWrapperRuntimeException zapWrapperRuntimeExcept
         }
     }
 
-    public void writeUserMessagesWithScannedURLs(List<String> urls) {
-        for (String url : urls) {
+    public void writeUserMessagesWithDetectedURLs(List<Map<String, ApiResponse>> results) {
+        // The map looks like this: {processed=true, statusReason=OK, method=GET,
+        // reasonNotProcessed=, messageId=6, url=http://example.com, statusCode=200}
+        for (Map<String, ApiResponse> result : results) {
+            String url = result.get("url").toString();
```
de-jcup, line 67: You should handle the situation when result.get("url") returns null - or can this never happen?

sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/zapwrapper/helper/ZapProductMessageHelper.java
```diff
@@ -58,14 +60,23 @@ public void writeProductError(ZapWrapperRuntimeException zapWrapperRuntimeExcept
...
             // robots.txt and sitemap.xml always appear inside the sites tree even if they
             // are not available. Because of this it is skipped here.
             if (url.contains("robots.txt") || url.contains("sitemap.xml")) {
                 continue;
             }
-            writeSingleProductMessage(new SecHubMessage(SecHubMessageType.INFO, "Detect url to scan: " + url));
+            String statusCode = result.get("statusCode").toString();
```
de-jcup, line 73: again: is this ensured that it cannot be null? It's just a map...
Maybe the easiest way would be to provide a fetchSafe(map, $key, $defaultFallback) method to implement and to use...

sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/zapwrapper/helper/ZapProductMessageHelper.java
```diff
@@ -81,7 +92,7 @@ private void writeProductErrorForExitCode(ZapWrapperExitCode exitCode) throws IO
             break;
         case API_DEFINITION_CONFIG_INVALID:
             productMessageSupport.writeMessage(new SecHubMessage(SecHubMessageType.ERROR,
-                    "Only a single API file can be provided. Please use a single file for the API definition inside the filesystem->files section of the SecHub configuration."));
+                    "Please use files instead of folders for the API definition with the filesystem->files section of the SecHub configuration."));
```
de-jcup, line 95: The meaning has complete changed.
1. is it correct that you can add multiple files?
2. API_DEFINITION_CONFIG_INVALID: is the message always correct for the case?

sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/zapwrapper/scan/ZapScanner.java
```diff
@@ -247,7 +250,10 @@ void addIncludedAndExcludedUrlsToContext() throws ClientApiException {
         String followRedirects = "false";
         for (String url : scanContext.getZapURLsIncludeSet()) {
             clientApiFacade.addIncludeUrlPatternToContext(scanContext.getContextName(), url);
-            clientApiFacade.accessUrlViaZap(url, followRedirects);
+            // cannot try to access wildcarded URLs
+            if (!url.contains(".*")) {
+                clientApiFacade.accessUrlViaZap(url, followRedirects);
+            }
```
de-jcup, line 256: I think we should at least log the situation when the state "url.contains(".*") returns true

sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/zapwrapper/scan/ZapScanner.java
```diff
@@ -417,6 +423,10 @@ void cleanUp() {
             clientApiFacade.createNewSession("Cleaned after scan", "true");
             LOG.info("New and empty session inside Zap created.");
 
+            // Remove x-sechub-dast header
+            LOG.info("Remove '{}' replacer rule.", X_SECHUB_DAST_HEADER_NAME);
+            clientApiFacade.removeReplacerRule(X_SECHUB_DAST_HEADER_NAME);
+
             // Replacer rules are persistent even after restarting ZAP
             // This means we need to cleanUp after every scan.
             LOG.info("Start cleaning up replacer rules.");
```
de-jcup, line 432: Should line 426-429 not be below log statement of 432?

sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/zapwrapper/scan/ZapScanner.java
```diff
@@ -833,6 +843,31 @@ private String urlEncodeUTF8(String stringToEncode) {
...
+
+        // setting initiators to null means all initiators (ZAP components),
+        // this means spider, active scan, etc will send this rule for their requests.
+        String initiators = null;
+        // default URL is null which means the header would be send on any request to
+        // any URL
+        String url = null;
+
+        LOG.info("Add '{}' replacer rule.", X_SECHUB_DAST_HEADER_NAME);
+        clientApiFacade.addReplacerRule(description, enabled, matchtype, matchregex, matchstring, replacement, initiators, url);
```
de-jcup, line 868: The API is so... using local variables here is a good idea.

sechub-wrapper-owasp-zap/src/test/java/com/mercedesbenz/sechub/zapwrapper/scan/ZapScannerTest.java
```diff
@@ -343,7 +343,7 @@ void set_includes_and_excludes_api_facade_is_called_once_for_each_include_and_on
 
         /* test */
         verify(clientApiFacade, times(includes.size())).addIncludeUrlPatternToContext(any(), any());
-        verify(clientApiFacade, times(includes.size())).accessUrlViaZap(any(), any());
+        verify(clientApiFacade, times(2)).accessUrlViaZap(any(), any());
```
de-jcup, line 346: why is this a 2 and not as formerly the included size. Please at least document why this magic number is used (or maybe I just do not get it...)

sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/zapwrapper/internal/scan/ClientApiFacade.java
```diff
@@ -370,18 +372,34 @@ public ApiResponse stopSpiderScan(String scanId) throws ClientApiException {
...
      */
-    public List<String> getAllSpiderUrls() throws ClientApiException {
-        List<ApiResponse> results = ((ApiResponseList) clientApi.spider.allUrls()).getItems();
-        List<String> urls = new ArrayList<>();
-        for (ApiResponse response : results) {
-            urls.add(response.toString());
+    public List<Map<String, ApiResponse>> getFullSpiderResults(String scanId) throws ClientApiException {
+        List<Map<String, ApiResponse>> fullResults = new LinkedList<>();
+
+        ApiResponseList results = (ApiResponseList) clientApi.spider.fullResults(scanId);
```
de-jcup, line 387: ZAP has two different spiders:
1. "normal"
2. ajax spider

do we get here all results?

sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/zapwrapper/helper/ZapProductMessageHelper.java
```diff
@@ -58,14 +60,23 @@ public void writeProductError(ZapWrapperRuntimeException zapWrapperRuntimeExcept
...
                 continue;
             }
-            writeSingleProductMessage(new SecHubMessage(SecHubMessageType.INFO, "Detect url to scan: " + url));
+            String statusCode = result.get("statusCode").toString();
+            String statusReason = result.get("statusReason").toString();
+            String method = result.get("method").toString();
+
+            String message = "URL: '%s' returned status code: '%s/%s' on detection phase for request method: '%s'".formatted(url, statusCode, statusReason,
+                    method);
+            writeSingleProductMessage(new SecHubMessage(SecHubMessageType.INFO, message));
```
de-jcup, line 79: IMO this can handle to an extreme amount of messages!
Let's assume ZAP has scanned 2000 URLs -> we have 2000 entries here.
Even when the scans were for the same URL but with different content!

Should we filter/reduce here to avoid duplicates?

de-jcup (changes requested):
Good job, but I found some parts we could change/should discuss.

sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/zapwrapper/helper/ZapProductMessageHelper.java
winzj: Yes in the first version we supported only a single file, but currently we can handle multiple files, but folders are not supported inside the data section. I forgot to update the message the last time and saw the old user message while I was working on the current issue. It is documented with the examples for any user, that they can use multiple files. I will update the message to a more generic version so the user knows which section to check.

de-jcup (approved):
Good work :+1:

PR #3527, opened 2024-10-17
implemented signal handling for sechub server container #3470

### LINKED ISSUE #3470: Adapt SIGTERM handling in K8s solution for SecHub deployments with scheduler on
## Situation
With 
- #390 
 
the SecHub server is now able suspend and resume jobs when it comes to SIGTERM from OS.

But in our `sechub-solution` the SIGTERM signal is currently not pushed through. So a K8s deployment does not lead to supsending.

## Wanted
K8s deployment `SIGTERM` signal shall be sent to SecHub server application a K8s environment

## Solution
Like done for PDS solutions, the `SIGTERM` signal shall be trapped inside the `start.sh` script and given to the java process

sven-dmlr:
- also updated Alpine version
- and fixed a typo
- closes #3470

### DIFF
--- a/.github/workflows/release-client-server-pds.yml
+++ b/.github/workflows/release-client-server-pds.yml
@@ -33,7 +33,7 @@ permissions:
   pull-requests: write
 
 env:
-  ACTIONS_BASE_IMAGE_ALPINE: alpine:3.17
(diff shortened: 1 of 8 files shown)


haerter-tss (approved):
Looks good! 🥇

PR #3528, opened 2024-10-18
Implement TOTP generation #3526

### LINKED ISSUE #3526: Implement possibility to compute TOTP as an additional authentication factor during webscans
### Situation
In the future it shall be possible for webscan to support more complex login mechanism, even multi factor authentication webscan logins.
One of the most common mechanisms is using TOTP. There are many authenicator apps for smartphones and a lot of internet services support this solution as a second factor for the authentication.

### Wanted
The ZAP wrapper application shall be able to compute TOTP needed as an additional authentication factor during webscans.

### Solution
Use a library or implement TOTP for the ZAP wrapper application according to [RFC-6238](https://datatracker.ietf.org/doc/html/rfc6238).

winzj:
- closes #3526

### DIFF
--- a/gradle/libraries.gradle
+++ b/gradle/libraries.gradle
@@ -55,6 +55,8 @@ ext {
       apache_commons_fileupload2_jakarta:      "2.0.0-M1",
       apache_commons_compress:                 "1.25.0",
       apache_commons_lang3:                    "3.14.0",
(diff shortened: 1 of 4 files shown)


de-jcup (approved):
Good work :+1:

PR #3539, opened 2024-10-21
Changed log output and level + renamed method name #3538

### LINKED ISSUE #3538: Job marker log output is misleading and must be changed
## Situation
`ScheduleJobMarkerService` contains code part with LOG message :

```java
 LOG.error("Did not found executablejob for next job UUID:{}. This is very problematic and should never happen!", nextJobId);
```
This was done for 
- #390 

The intention was to log situations where no real job processing was done (there was a implementation failure at this time, where suspended jobs where fetched in first step, but not accepted as executed which led to an infinite loop).

But the situation can happen a lot without being a real error: Depending when the different SecHub instances try to fetch the next job, race conditions may happen.

Because this is an error message it can lead to confusion at deployment / runtime when looking into the logs: I looks like a SecHub job is beeing processed by different instances the same time (which is not happening).

## Wanted
NO misunderstanding about log entries.

## Solution
1. Change log level to trace
2. Change log output message, so it is clear, this is an info and not 100% a problem.

de-jcup:
This PR
- closes #3538


sven-dmlr (approved):
Lokks good to me. :+1: 
Successfully tested with 5 server instances.

PR #3546, opened 2024-10-22
Put encrypted access token in web ui cookies #3479

### LINKED ISSUE #3479: [WebUI] Encrypt the access token for every response
## Situation

Currently the access token is not encrypted when it's sent back to the user (or his browser). 

## Wanted

We want to protect the access token whenever it leaves the WebUI Server.

## Solution

- Encrypt the access token with AES256 before sending it to the client
- The token must be only sent in the `http only session cookie` (that way its not accessible through JavaScript)
- When the server receives the token it decrypts it before validation
- Set AES encryption key as environment variable 
- Ensure that the token is really not accessible in Browser and is encrypted (probably manual testing)

image

hamidonos:
- closes #3479


sechub-web-server/src/main/java/com/mercedesbenz/sechub/webserver/encryption/AES256Encryption.java
```diff
@@ -0,0 +1,64 @@
...
+
+    AES256Encryption(AES256EncryptionProperties properties) throws GeneralSecurityException {
+        SecretKey secretKey = new SecretKeySpec(properties.getSecretKeyBytes(), TRANSFORMATION);
+        this.encrypt = Cipher.getInstance(TRANSFORMATION);
+        encrypt.init(Cipher.ENCRYPT_MODE, secretKey);
+        this.decrypt = Cipher.getInstance(TRANSFORMATION);
+        decrypt.init(Cipher.DECRYPT_MODE, secretKey);
+    }
+
+    public String encrypt(String plainText) {
```
de-jcup, line 37: The encrypted text must be here always base64 encoded. If we do not separate the encoding and the encryption, the parameter should be renamed to `base64plainText` IMO.

Also the method should be rneamed to `encryptBase64`.

Remark: But I think separate encoding and encryption, would make things clearer, easier etc.

de-jcup (changes requested):
Most looks very good. But I found some parts to discuss and currently the build is red (archunit test fails - I think this because the generic runtime exception I mentioned inside the review comments).

de-jcup (approved):
Looks good to me! Well done :+1: 
I really like the WithMockJwtUser annotation :-)

PR #3586, opened 2024-11-05
Feature 456 user list projects

### LINKED ISSUE #456: Usecase: lists projects with data
## Situation
There exists no API call for a user, owner or admin to get a list of his assigned projects.

## Wanted 
The user should be able to list all projects one is assigned to.

For example, user Alice has access to three projects:

- inventory_app
- store_managment
- crm_app

Alice should be able to list all projects she has access to, as well as the name of the owner of the projects.

If Alice is the owner of a project, she should be able to see the userID of all assigned users.

If Alice is an Admin, she gets all information about assigned projects.

## Solutions
- implement new Rest Controller
- implement Service to get List of projects
- response should be JSON Array

lorriborri:
- closes #456

### DIFF
--- a/sechub-testframework/src/main/java/com/mercedesbenz/sechub/test/SecHubTestURLBuilder.java
+++ b/sechub-testframework/src/main/java/com/mercedesbenz/sechub/test/SecHubTestURLBuilder.java
@@ -23,7 +23,7 @@ public class SecHubTestURLBuilder extends AbstractTestURLBuilder {
     private static final String API_ADMIN_CONFIG = API_ADMIN + "/config";
     private static final String API_ADMIN_CONFIG_MAPPING = API_ADMIN_CONFIG + "/mapping";
     private static final String API_PROJECT = "/api/project";
-    private static final String API_USER = "/api/user";
+    public static final String API_PROJECTS = "/api/projects";
 
     public static SecHubTestURLBuilder https(int port) {
         return new SecHubTestURLBuilder("https", port);
@@ -462,7 +462,7 @@ public class SecHubTestURLBuilder extends AbstractTestURLBuilder {
     /* +............................ user self service .....+ */
     /* +-----------------------------------------------------------------------+ */
     public String buildUserListProjectsUrl() {
-        return buildUrl(API_USER, "projects");
+        return buildUrl(API_PROJECTS);
     }
 
     /* +-----------------------------------------------------------------------+ */
--- a/sechub-testframework/src/main/java/com/mercedesbenz/sechub/test/SecHubTestURLBuilder.java
+++ b/sechub-testframework/src/main/java/com/mercedesbenz/sechub/test/SecHubTestURLBuilder.java
@@ -23,6 +23,7 @@ public class SecHubTestURLBuilder extends AbstractTestURLBuilder {
     private static final String API_ADMIN_CONFIG = API_ADMIN + "/config";
     private static final String API_ADMIN_CONFIG_MAPPING = API_ADMIN_CONFIG + "/mapping";
     private static final String API_PROJECT = "/api/project";
+    private static final String API_USER = "/api/user";
 
     public static SecHubTestURLBuilder https(int port) {
         return new SecHubTestURLBuilder("https", port);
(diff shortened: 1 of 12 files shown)


sechub-administration/src/test/java/com/mercedesbenz/sechub/domain/administration/project/ProjectServiceTest.java
```diff
@@ -0,0 +1,164 @@
...
+
+    private ProjectService serviceToTest;
+    private UserRepository userRepository;
+
+    @BeforeEach
+    public void before() {
+        userRepository = mock(UserRepository.class);
+        UserInputAssertion userInputAssertion = mock(UserInputAssertion.class);
+
+        serviceToTest = new ProjectService(userRepository, userInputAssertion);
```
hamidonos, line 28: You don't need to initialize the mocks and service class every time again.

You might as well initialize the mocks at class level once and the in the `@BeforeEach` method use `Mockito.reset(...)` (when needed)

Your service can then be initialized once as well.

sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/project/ProjectData.java
```diff
@@ -0,0 +1,52 @@
...
+        this.isOwned = isOwned;
+    }
+
+    public void setAssignedUsers(String[] assignedUsers) {
+        this.assignedUsers = assignedUsers;
+    }
+
+    @JsonProperty(PROPERTY_IS_OWNED)
+    public boolean isOwned() {
+        return isOwned;
```
hamidonos, line 38: You will run into a NullPointerException here when `this.isOwned` is null

sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/project/ProjectService.java
```diff
@@ -0,0 +1,62 @@
...
+        this.userInputAssertion = userInputAssertion;
+    }
+
+    public ProjectData[] getProjectDataList(String userId) {
+        userInputAssertion.assertIsValidUserId(userId);
+
+        User user = userRepository.findOrFailUser(userId);
+        List<ProjectData> projects = new ArrayList<>();
+
+        collectProjectDataForUser(user, projects);
```
hamidonos, line 29: Since the `projects` collection is empty I'd rather have a statement like this

`List<ProjectData> projects = collectProjectDataForUser(user)`

The method creates and returns the collection, and the caller will store it in the projects variable.

This is generally a more declarative style, making the code easier to understand. It clearly indicates that `collectProjectDataForUser` is responsible for returning a new collection based on the user, without modifying any external state.

This is a better choice if you don't want the method to depend on external mutable state (like modifying an existing list) and if you prefer a method to return fresh data instead of updating passed parameters.

hamidonos (changes requested):
Good work but there are some findings that need to be looked at

1. You are missing the license header in many classes
2. Think about using AssertJ "assertThat(...).is(...)" for testing. It has a fluent API and is generally better for testing in my opinion
3. Simple POJO constructs like entities don't require mocking in my opinion. This will make the test setup much more complex with no real benefit. Just initialize the object right away with some default values.
4. Your test cases could be more readable. I'd suggest to take a look at the interface `ArgumentsProvider` where you initialize the desired user objects and projects directly with the right values. The method `setUpTestCase` does to much in my opinion.

sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/project/ProjectData.java
```diff
@@ -0,0 +1,54 @@
...
+public class ProjectData {
+    public static final String PROPERTY_PROJECT_ID = "projectId";
+    public static final String PROPERTY_OWNER = "owner";
+    public static final String PROPERTY_IS_OWNED = "isOwned";
+    public static final String PROPERTY_ASSIGNED_USERS = "assignedUsers";
+
+    private String projectId;
+    private String owner;
+    private Boolean isOwned;
+    @JsonInclude(JsonInclude.Include.NON_NULL)
```
hamidonos, line 16: here you put the annotation on the field while you put the other annotations on the getters
while both approaches work I would not mix them 

consider putting all annotations on one level (either field or getter)

sechub-testframework/src/main/java/com/mercedesbenz/sechub/test/SecHubTestURLBuilder.java
```diff
@@ -23,6 +23,7 @@ public class SecHubTestURLBuilder extends AbstractTestURLBuilder {
     private static final String API_ADMIN_CONFIG = API_ADMIN + "/config";
     private static final String API_ADMIN_CONFIG_MAPPING = API_ADMIN_CONFIG + "/mapping";
     private static final String API_PROJECT = "/api/project";
+    public static final String API_PROJECTS = "/api/projects";
```
hamidonos, line 26: did you mean to make this public?

all other constants are private

hamidonos (approved):

PR #3589, opened 2024-11-06
Feature 3581 release workflow for sechub web UI frontend

### LINKED ISSUE #3581: Create a release workflow for sechub-web-ui frontend
## ToDo
Similar to the other release workflows, a release has to be built which includes the generated files as artifacts,

sven-dmlr:
- closes #3581


haerter-tss (approved):
Looks good! 🎯🎃

PR #3603, opened 2024-11-11
implement OAuth2 authentication & integration for SecHub server #3559

### LINKED ISSUE #3559: SecHub Server: Enable OAuth2 Authentication
## Situation

Currently the SecHub Server does not support OAuth2 Authentication but only Basic Auth through Http Headers.

## Wanted

We want to be able to use OAuth2 Authentication as well, when talking to the SecHub Server.

## Solution

- Implement OAuth2 Authentication for SecHub Server
- Integrate OAuth2 Provider with SecHub Server (must be configurable)
- Use SecHub Server as `oAuth2ResourceServer`
- SecHub Server checks if Basic Auth header is present (current authentication method)
- If no Basic Auth header is present check for the `Bearer` Token Header
- Put common spring security code in a library `sechub-commons-spring`

hamidonos:
- closes #3559


sechub-shared-kernel/src/main/java/com/mercedesbenz/sechub/sharedkernel/security/JwtDecoderConfiguration.java
```diff
@@ -0,0 +1,20 @@
...
+import org.springframework.security.oauth2.jwt.NimbusJwtDecoder;
+
+import com.mercedesbenz.sechub.sharedkernel.Profiles;
+
+@Configuration
+@Profile(Profiles.OAUTH2)
+class JwtDecoderConfiguration {
+
+    @Bean
+    JwtDecoder jwtDecoder(OAuth2Properties oAuth2Properties) {
```
de-jcup, line 17: Could it be possible that OAuth2Properties is null?
Is this autowired - without the annotation? Should we add `@Autowired` to make it clear that the properties are injected?

sechub-shared-kernel/src/main/java/com/mercedesbenz/sechub/sharedkernel/security/OAuth2Properties.java
```diff
@@ -0,0 +1,23 @@
...
+@ConfigurationProperties(OAuth2Properties.PREFIX)
+public class OAuth2Properties {
+
+    static final String PREFIX = "sechub.security.oauth2";
+    private static final String ERR_MSG_FORMAT = "The property '%s.%s' must not be null";
+
+    private final String jwkSetUri;
+
+    public OAuth2Properties(String jwkSetUri) {
+        this.jwkSetUri = requireNonNull(jwkSetUri, ERR_MSG_FORMAT.formatted(PREFIX, "jwk-set-uri"));
```
de-jcup, line 17: As long as we do not have multiple properties which require non null
I think this would be simpler, easier to understand:
```suggestion
        this.jwkSetUri = requireNonNull(jwkSetUri, "The property 'sechub.security.oauth2.jwk-set-uri' must not be null"));
```

sechub-shared-kernel/src/test/java/com/mercedesbenz/sechub/sharedkernel/security/OAuth2IntegrationTest.java
```diff
@@ -0,0 +1,138 @@
...
+import com.mercedesbenz.sechub.testframework.spring.OAuth2SecurityTestConfiguration;
+import com.mercedesbenz.sechub.testframework.spring.YamlPropertyLoaderFactory;
+
+/**
+ * This test class verifies the integration of Spring Security OAuth2
+ * components.
+ *
+ * <p>
+ * Unlike {@link SecHubApiSecurityConfigurationTest}, which primarily tests if a
+ * endpoint is secured on a abstract level, this class exercises the full OAuth2
```
de-jcup, line 32:
```suggestion
 * endpoint is secured on an abstract level, this class exercises the full OAuth2
```

sechub-shared-kernel/src/test/java/com/mercedesbenz/sechub/sharedkernel/security/OAuth2IntegrationTest.java
```diff
@@ -0,0 +1,138 @@
...
+ * flow with real OAuth2 mechanisms. We do that by relying the
```
de-jcup, line 33:
```suggestion
 * flow with real OAuth2 mechanisms. We do that by relying to the
```

sechub-testframework-spring/src/main/java/com/mercedesbenz/sechub/testframework/spring/OAuth2SecurityTestConfiguration.java
```diff
@@ -0,0 +1,119 @@
...
+ * This configuration class provides the necessary beans to test Springs OAuth2
+ * integration with SecHub components.
+ *
+ * @author hamidonos
+ */
+@Configuration
+@Import({ AuthUserDetailsService.class })
+public class OAuth2SecurityTestConfiguration {
+
+    public static final String BEARER = OAuth2AccessToken.TokenType.BEARER.getValue() + " ";
```
de-jcup, line 35:
```suggestion
    public static final String BEARER_PREFIX = OAuth2AccessToken.TokenType.BEARER.getValue() + " ";
```

de-jcup (approved):
Very good work! I especially like the way to mock the OAuhth2 process by just adding `sechub-testframework-spring` as a test dependency... 

I found only some typos or little things that could be improved IMO. 
Feel free to change it or not.

sechub-shared-kernel/src/main/java/com/mercedesbenz/sechub/sharedkernel/security/JwtDecoderConfiguration.java
hamidonos: No
The Configuration is profiled with oauth2
OAuth2Properties must not be null when using oauth2 profile
in this case spring will throw a missing bean exception

You generally don't need @Autowire here
Knowing that you're inside the Bean definition it should be clear (similar to constructor injection)

PR #3612, opened 2024-11-14
implement actuator for web ui server #3515

### LINKED ISSUE #3515: Enable actuator for Web UI Server (Spring Boot)
## Situation

Currently we have no healthcheck endpoint for the SecHub Server WebUI.

## Solution

- Enable `/actuator/health/liveness` and `/actuator/health/readiness` (https://spring.io/blog/2020/03/25/liveness-and-readiness-probes-with-spring-boot)
- These endpoints should accessible by the K8S cluster but not publicly (we should not expose any internal states to the outside world)
- Host the actuator under an internal port (this port is never exposed by K8S)

hamidonos:
- closes #3515


lorriborri (approved):
Looks good to me! Good work! 🐢

PR #3613, opened 2024-11-14
Feature 3609 web UI forwarding to web server

### LINKED ISSUE #3609: web-ui: implement forwarding to web-server
## Situation
The whole communication shall run through a single URL.
No split between web-ui static and web-server Spring app.
This gives a consistent experience to the end user.

## ToDo
Forward the web-server pathes to the same and serve the static content directly (Nginx configuration).

Locations to forward:
* /api
* /error
* /login
* /oauth2

Additionally add the SecHub logo image and favicon.ico to Nginx's web root directory.

### LINKED ISSUE #3614: web-ui: Fix builds and release workflow
## Situation
After #3606 these parts are broken because the generated artifacts are in a different folder than before (`dist/` vs. `.output/public`)
* Container build type "copy"
* Container build type "download"
* The web-ui release workflow (GHA)

## ToDo
Fix above

sven-dmlr:
- closes #3609
- closes #3614

### DIFF
--- a/sechub-web-ui-solution/docker/nginx/nginx.conf
+++ b/sechub-web-ui-solution/docker/nginx/nginx.conf
@@ -48,6 +48,8 @@ http {
     location /health {
       # no access logging of health checks
       access_log off;
+      return 200 'OK';
+      add_header Content-Type text/plain;
     }
 
     ssl_certificate /etc/nginx/certificates/sechub-web-ui.cert;
--- a/sechub-web-ui-solution/docker/Web-UI-Debian.dockerfile
+++ b/sechub-web-ui-solution/docker/Web-UI-Debian.dockerfile
@@ -105,9 +105,7 @@ ENV HTDOCS_FOLDER="${HTDOCS_FOLDER}"
 # using fixed group and user ids + prepare alive check file
 RUN usermod -u "$UID" "$USER" && \
     groupmod -g "$GID" "$USER" && \
-    NGINX_ALIVE_DIR="$HTDOCS_FOLDER/health" && \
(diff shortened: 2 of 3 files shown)


sechub-web-ui-solution/docker/nginx/nginx.conf
```diff
@@ -0,0 +1,72 @@
+worker_processes 1;
```
haerter-tss, line 1: Do you limit the nginx to one core on purpose? Otherwise `auto`  could be an alternative.

sechub-web-ui-solution/docker/nginx/nginx.conf
```diff
@@ -0,0 +1,72 @@
...
+
+    location /oauth2 {
+      proxy_pass https://web-server-internal:4443/oauth2;
+      include sechub_proxy_settings.conf;
+    }
+
+    location /health {
+      # no access logging of health checks
+      access_log off;
+    }
```
haerter-tss, line 51: I guess the healthcheck is already working, but it could be an improvement if the return code is specified.
```suggestion
location /health {
  # no access logging of health checks
  access_log off;
  return 200 'OK';
  add_header Content-Type text/plain;
}
```

sechub-web-ui-solution/docker/nginx/nginx.conf
sven-dmlr: @lorriborri has set this value.
I left it because I know "one process per container" as best-practise in Kubernetes.

sechub-web-ui-solution/docker/nginx/nginx.conf
sven-dmlr: Good suggestion!
Is now implamented.

PR #3651, opened 2024-11-26
Improve documentation output for REST doc + fix scope calculation

### LINKED ISSUE #3647: REST documentation does not contain query parameters
## Situation
The generated REST documentation does contain query parameters only inside curl examples but not in description.

### Example
when looking (currently) at latest REST documentation of UC_71
https://mercedes-benz.github.io/sechub/latest/sechub-restapi.html#section-ucrestdoc-uc-UC_071

The query parameters are only visible inside the curl example:

image

## Wanted
Query parameters shall be contained as well

## Solution
The query parameters are generated but currently not included in documentation.
Means the solution is to include the generated files `query-parameters.adoc` after ``path-parameters.adoc

### LINKED ISSUE #3652: Fix wrong auto scope calculation for documented parts
## Situation

The documentation for system properties contains some strange scope definitions like `s` or `p` inside documentation.

Here an example:
image

## Analysis
The problem is inside asciidoc generator util: 
when `DocGeneratorUtil.toCamelOne(Class<?> clazz)` is called with a class which has multiple uppercased characters, all other parts are not recognized. Because at generation time the scope in output is also lowercased only one lowercased parameter is printed out.

For example:
- `SMTPServer` will become currently `s`.
- `PDSResilienceConsultant` will become currently `p`.

## Wanted
- Scope for `SMTPServer` shall be: `smtpserver`
- Scope for `PDSResilienceConsultant` shall be `pdsresilience`

de-jcup:
This PR 
- closes #3647 
- closes #3652

### DIFF
--- a/sechub-doc/src/main/java/com/mercedesbenz/sechub/docgen/util/DocGeneratorUtil.java
+++ b/sechub-doc/src/main/java/com/mercedesbenz/sechub/docgen/util/DocGeneratorUtil.java
@@ -84,21 +84,28 @@ public class DocGeneratorUtil {
 
     /**
      * Reduces string to second upper cased char - e.g. "NetsparkerInstallSetupImpl"
-     * would be replaced to "Netsparker"
+     * would be replaced to "Netsparker". Having multiple upper case letters at the
+     * beginning will keep all upper cased parts. E.g. "SMTPServerConfiguration"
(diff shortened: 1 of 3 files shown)


sven-dmlr (approved):
Looks good :+1:

PR #3656, opened 2024-11-26
add VIN pattern rule to pds-gitleaks #3655

### LINKED ISSUE #3655: pds-gitleaks shall be able to detect VINs
### Wanted
The pds-gitleaks should be extended to detect VINs, because sometimes VINs are treated as PII.

### Solution
Implement a regex rule for Gitleaks and perform some tests.

winzj:
This pull request:
- closes #3655 
- updates the gitleaks version to the latest one `8.21.2`, since at least 8.21 is required for the `[[rules.allowlists]]` feature used in this pull request
- removes the customized `generic-api-key` rule since the default changed the keywords to what was customized before


winzj:
This is too specific and won't be implemented

PR #3702, opened 2024-12-03
readiness checks added #3514

### LINKED ISSUE #3514: Implement Readiness Check for K8S
## Situation 

Currently we do not use readiness check to check if traffic should be routed to the pods. 
This means that all traffic will hit the pod while it's alive.

## Wanted

To know if we can route traffic to the pod we should also include a readiness check

## Solution

- Implement the readiness check
- Rethink the startup probe. Do we really need a start up probe? Most of our apps are "fast-starting" apps. Could be over engineered.
- Readiness checks are typically configured in shorter intervals than liveness checks

## Notes

**1. Startup Probe**
**Purpose**: Checks whether the application within the container has started successfully.
**Use Case**: It is especially useful for slow-starting applications (e.g., those that need a long initialization time).
**Behavior**: During the startup period, only the startup probe is active. Kubernetes will not check liveness or readiness probes until the startup probe succeeds. If the startup probe fails, the container will be restarted.
**Typical Configuration**: You configure it when your application needs extra time to initialize before it is considered "alive."
**Key Point**: Prevents premature liveness failures and restarts during slow application startup.

**2. Readiness Probe**
**Purpose**: Determines whether a container is ready to accept traffic.
**Use Case**: It ensures that only containers that are ready can receive traffic. If the readiness probe fails, Kubernetes will stop sending traffic to the container (removes it from the service endpoints), but the container remains running.
**Behavior**: A failure in the readiness check doesn't restart the container; it simply marks the container as not ready to serve traffic until it succeeds again.
Typical Configuration: You configure it to check that your application can handle requests, such as checking whether it can connect to a database.
**Key Point**: Controls traffic flow by determining if the application is ready to serve.

**3. Liveness Probe**
**Purpose**: Checks whether the container is still running and healthy.
**Use Case**: If the liveness probe fails, Kubernetes will restart the container. This is useful for detecting deadlocks, memory leaks, or other conditions where the container is still running but is unhealthy.
**Behavior**: Liveness probes are checked periodically. If the probe fails, Kubernetes assumes the application is in a bad state and restarts the container.
**Typical Configuration**: You configure it to detect critical application health issues, such as checking whether a health check endpoint responds within a specified time.

sven-dmlr:
- readiness check added
- use Spring's readiness probe
- probe timings improved
- closes #3514

### DIFF
--- a/sechub-pds-solutions/checkmarx/helm/pds-checkmarx/Chart.yaml
+++ b/sechub-pds-solutions/checkmarx/helm/pds-checkmarx/Chart.yaml
@@ -9,4 +9,4 @@ type: application
 # This is the chart version.
 # This version number should be incremented each time you make changes to the chart and its templates.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.4.0
+version: 1.5.0
--- a/sechub-pds-solutions/checkmarx/helm/pds-checkmarx/templates/deployment.yaml
+++ b/sechub-pds-solutions/checkmarx/helm/pds-checkmarx/templates/deployment.yaml
@@ -124,15 +124,15 @@ spec:
               value: "{{ .Values.storage.sharedVolume.upload.dir }}"
 {{- end}}
           ports:
-            - name: pds-actuator-port
+            - name: pds-health-port
               containerPort: 10251
             - name: pds-https-port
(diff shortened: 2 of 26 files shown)


haerter-tss (approved):
Looks good! 🛷🎡

PR #3715, opened 2024-12-09
Use template data for script authentication #3710

### LINKED ISSUE #3710: ZAP wrapper shall use template data
### Situation
With #3545 we are now able to use template data inside pds-solutions.

### Wanted
Make use of the template data to enable script authentication in pds-owaspzap solution.

### Solution
- the script file must be correctly loaded from the extracted directory
- the template webscans always use the template variables `username` and `password` for pds-owaspzap solution
- make sure the path to the script is exported as environment variable in the `owasp-zap.sh` script if it exists
- document the following requirements on the pds-solution side:
  - the script file is always called `script.groovy`
  - the variables `username` and `password` must be present for template webscans. The value of `username` can be any user identification like an ID, an email or a name.

winzj:
- closes #3710

Additionally:
- unnecessary mock configurations were deleted, since the implicit default is sufficient
- all Collections for the ZAP configuration are now handled the same way

As discussed, improved exception handling as well.


de-jcup (changes requested):
Good job. But I found some parts which must be changed.

de-jcup (approved):
Good work - let's merge :rocket:

PR #3728, opened 2024-12-12
added Oauth2 section to server Helm Chart #3727

### LINKED ISSUE #3727: Server Helm Chart: Add Oauth2 section
## Situation
Currently all Oauth2 setting must be defined as environment parameters when deploying.
Documentation: https://mercedes-benz.github.io/sechub/latest/sechub-operations.html#section-general-configuration

## Wanted
A section `oauth2` in the values.yaml which eases Oauth2 configuration by listing the attributes

## Solution
Implement above.

sven-dmlr:
- closes #3727

### DIFF
--- a/sechub-solution/helm/sechub-server/templates/deployment.yaml
+++ b/sechub-solution/helm/sechub-server/templates/deployment.yaml
@@ -473,6 +473,22 @@ spec:
         - name: SERVER_TOMCAT_ACCESSLOG_PATTERN
           value: "{{ .Values.server.spring.embeddedTomcat.logging.accessLogFormat }}"
 {{- end }}
+{{- if .Values.oauth2.enabled }}
+# ----------------------------------------------------------------------------------------------#
+# - Scope: Oauth2
+# --------------------------------------------------------------------------------------------- #
+        # See https://mercedes-benz.github.io/sechub/latest/sechub-operations.html#section-general-configuration
+        - name: SECHUB_SECURITY_OAUTH2_MODE
+          value: "{{ .Values.oauth2.mode }}"
+        - name: SECHUB_SECURITY_OAUTH2_JWK_SET_URI
+          value: "{{ .Values.oauth2.jwt.uri }}"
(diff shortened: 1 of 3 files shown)


sechub-solution/helm/sechub-server/templates/deployment.yaml
```diff
@@ -473,6 +473,22 @@ spec:
...
+        - name: SECHUB_SECURITY_OAUTH2_MODE
+          value: "{{ .Values.oauth2.mode }}"
+        - name: SECHUB_SECURITY_OAUTH2_JWK_SET_URI
+          value: "{{ .Values.oauth2.jwt.uri }}"
+        - name: SECHUB_SECURITY_OAUTH2_OPAQUE_TOKEN_CLIENT_ID
+          value: "{{ .Values.oauth2.opaque_token.client_id }}"
+        - name: SECHUB_SECURITY_OAUTH2_OPAQUE_TOKEN_CLIENT_SECRET
+          value: "{{ .Values.oauth2.opaque_token.client_secret }}"
+        - name: SECHUB_SECURITY_OAUTH2_OPAQUE_TOKEN_INTROSPECTION_URI
+          value: "{{ .Values.oauth2.opaque_token.introspection_uri }}"
```
haerter-tss, line 490: In no other value variables are underscores. This might even be a forbidden character so I would suggest to replace them with dots:

```suggestion
          value: "{{ .Values.oauth2.opaque_token.client_id }}"
        - name: SECHUB_SECURITY_OAUTH2_OPAQUE_TOKEN_CLIENT_SECRET
          value: "{{ .Values.oauth2.opaque.token.client.secret }}"
        - name: SECHUB_SECURITY_OAUTH2_OPAQUE_TOKEN_INTROSPECTION_URI
          value: "{{ .Values.oauth2.opaque.token.introspection.uri }}"
```

haerter-tss (changes requested):
Please have a look at my suggestion. I think we should use dots instead of underscores in value variables.

haerter-tss (approved):
Looks good 🏔️

PR #3737, opened 2024-12-13
Fix rule deactivation error handling #3736

### LINKED ISSUE #3736: Ensure ZAP wrapper only throws exception during rule deactivation if ZAP is not reachable
### Situtation
If anything goes wrong commucation with ZAP we get an error of type `ClientApiException.class`, not matter what error happened exactly.
To find out the real problem behind the exception, it must be analyzed further. Currently due to a false assumption of the content of the error message, the exception is re-thrown sometimes even when the ZAP is reachable.

### Solution
Only throw an exception when the ZAP is not reachable. Otherwise the rule cannot be deactivated because it does not exist, has a unrecognized format or somethign else was wrong with the parameter.

winzj:
- closes #3736


sven-dmlr (approved):
Looks good to me :+1:

PR #3749, opened 2024-12-16
Overview of project jobs

### LINKED ISSUE #3451: [WebUI] Use Case: Overview of project Jobs
## Situation
No Overview if projects jobs

## Wanted
When clicking on a project I want to view the jobs that have been started for the project. There should be an state, to see if the job is already finished or still running or was canceled. Also I wish to see the color of the traffic light.

## Solution
- Jobs listed with pagination
- Refresh Button for view
- Job status
- Job result
- Job stage

Projekt Auswahl - Scan Status

lorriborri:
- closes #3451


sechub-web-ui/src/components/ProjectsList.vue
```diff
@@ -40,18 +41,24 @@
...
       const projects = ref([])
       const loading = ref(true)
       const error = ref(null)
+      const router = useRouter()
+      const store = useProjectStore()
 
       onMounted(async () => {
         try {
           projects.value = await defaultClient.withProjectApi.getAssignedProjectDataList()
+          store.storeProjects(projects)
```
de-jcup, line 61: currently you only store the projects - but it seems you do not check if already loaded and use the store as a cache.

Maybe this will done later - at the moment I did not find anything that uses `getProjectById` (at least directly) to obtain data from store.

de-jcup (approved):
Good job :rocket: - I had only one thing: IMO the store is currently only written but not read anywhere. But maybe this was a necessary setup for coming features.

sechub-web-ui/src/components/ProjectsList.vue
lorriborri: You observed this correctly - the store is intended for the next issue, where it will be used to show the project details. At the moment, the functionality to check if a project is already loaded and use the store as a cache is not implemented - we could address this Issue in upcoming tasks.

PR #3755, opened 2024-12-17
Fix DAUI fat jar problem #3754

### LINKED ISSUE #3754: Fix start problems with DAUI fatjar
## Situation
When building the DAUI (developer admin UI) as a fat jar via 

```
./gradlew buildDeveloperAdminUI
```

the build works, but when calling 

```
java -jar sechub-developertools/build/libs/sechub-developer-admin-ui-0.0.0.jar
```

The application does not start because main class is not found

## Analyze
The fat jar is not a Spring Boot jar but instead a self made fat jar by a custom gradle task which allows duplicated entries etc. (quick and dirty). IMO this is the root reason for the problems.

## Wanted
It shall be possible to start the DAUI application without any problems

## Solution
Because `sechub-developertools` is not referenced from any other gradle submodule we 
can change it to be a Spring Application.

For backward compatibility, the original Application class `DeveloperAdministrationUI` will 
be kept as is.
But we provide a new `DAUIApplication` class which is a Spring Boot Application.
Via injected component `DAUICLIRunner` which is a command line runner, the original  `DeveloperAdministrationUI` will be called.

Means all existing IDE run configurations will work as before (without Spring) but at command line we have full working fat jar.

Steps to do
- introduce DAUI as Spring Boot Application
- remove old custom gradle tasks to build the DAUI
- update documentation
- remove every usage of `buildDeveloperAdminUI`

de-jcup:
This PR
- closes #3754


sven-dmlr (approved):
Looks good :+1: 

Although because the jar has been renamed from `sechub-developer-admin-ui-x.y.z.jar` to `sechub-developertools-x.y.z.jar` I had to change the jar name in
- `.github/workflows/release-client-server-pds.yml` and
- `sechub-developertools/scripts/launch-developer-admin-ui`

PR #3767, opened 2024-12-23
Start a scan via web UI

### LINKED ISSUE #1900: [WebUI] Use Case: Code Scan/Secret Scan/Licence Scan via WebUI
## Situation
The user can not perform a scan via the web-ui.

## Wanted
The user should be able to start a codeScan or a secretScan via the web-ui. For now the configuration should be minimal and created automatically from user input. The user should be able to select the desired scan. Scans types, where he is not registered for are not allowed and shown as inactive/ disabled checkboxes.

## Solution
Implement the scan according to the Go client (in three steps):

1. The client creates a new job
- create minimal sechub.json from selected scan, example:
 ```json
{
  "apiVersion" : "1.0",
  "data" : {
    "sources" : [ {
      "name" : "source",
      "fileSystem" : {
        "folders" : [ "." ]
      }
    } ]
  },
  "codeScan" : {
    "use" : ["source"]
  }
}
```
- API POST on project/[project ID]/job
2. The client uploads the sources/ binaries (tbd)
- Go clients zips the sources from filesystem and creates checksum -> user should not upload to web-ui (tbd)
- API project/[project ID]/job/[job UUID]/sourcecode
- or API project/[project ID]/job/[job UUID]/binaries
- Be Aware: we need to filter the files (like Go Client), user must ZIP - or maybe just hand over path to web client?
- Binary Upload: might need the size of the file for the upload?
3. The client approves the job (starts the scan)
- API PUT project/[project ID]/job/[job UUID]/approve
4. When job is approved the client switches to jobs view (where the new job is shown)

Part of #1813

New Scan - DAST II

### PR DESCRIPTION
**Pull request by lorriborri**
- closes #1900

### CODE REVIEW

File: sechub-web-ui/src/components/ProjectJobList.vue [hunk 1]
```diff
@@ -114,14 +125,29 @@
...
+            id: projectId.value,
+          },
+        })
+      }
+
+      function backToProjectsList () {
+        router.go(-1)
+      }
+
       onMounted(async () => {
```
hamidonos, line 147: whenever initializing a timer you should always make sure that its destroyed when the ui component is not active anymore

otherwise It will keep on running

```
onUnmounted(() => {
   clearInterval(interval);
});
```

sechub-web-ui/src/components/ProjectJobList.vue
```diff
@@ -61,8 +63,8 @@
             </v-table>
           </div>
           <Pagination
-            :current-page="jobsObject.page + 1"
-            :total-pages="jobsObject.totalPages"
+            :current-page="jobsObject.page || 0 + 1"
```
hamidonos, line 66: why not just `1` instead of `0 + 1` ?

sechub-web-ui/src/components/ProjectJobList.vue
```diff
@@ -114,14 +125,29 @@
...
+      }
+
+      function backToProjectsList () {
+        router.go(-1)
+      }
+
       onMounted(async () => {
         fetchProjectJobs(currentRequestParameters)
+        // polling every 10 sec. for job states
+        setInterval(() => fetchProjectJobs(currentRequestParameters), 10000)
```
hamidonos, line 150: I'd poll with an exponential backoff.. It's always a good practice
The server might be under heavy load or the API might return an error
In this case you want to give the system time to handle the request gracefully

This could be an improvement to your solution:

```
const data = ref([]);
const error = ref(null);
const maxAttempts = 10; // Maximum number of retries for backoff
const baseDelay = 1000; // Initial delay in milliseconds
let polling = true;

const fetchData = async (attempt = 0) => {
  try {
    console.log(`Polling attempt: ${attempt + 1}`);
    const response = await fetch('/api/data'); // Replace with your API endpoint

    if (!response.ok) {
      throw new Error(`HTTP error! status: ${response.status}`);
    }

    data.value = await response.json();
    error.value = null;

    // If successful, schedule the next poll with the base delay
    if (polling) {
      setTimeout(() => fetchData(), baseDelay);
    }
  } catch (err) {
    error.value = err.message;

    // If max attempts haven't been reached, apply exponential backoff
    if (attempt < maxAttempts - 1 && polling) {
      const delay = baseDelay * Math.pow(1.5, attempt);
      console.error(`Error: ${err.message}. Retrying in ${delay}ms...`);
      setTimeout(() => fetchData(attempt + 1), delay);
    } else {
      console.error('Max retry attempts reached. Stopping polling.');
    }
  }
};

onMounted(() => {
  polling = true;
  fetchData(); // Start polling immediately
});

onUnmounted(() => {
  polling = false; // Stop polling when the component is destroyed
});
```

With this you won't poll indefinitely
You also make sure that you don't keep on polling when the api returns an error

hamidonos (changes requested):
Looks good :) 

you should definitely make sure that the timer is removed when the component is closed

the improvement to the polling is optional imo but I think it makes sense

hamidonos (approved):

PR #3770, opened 2024-12-30
Add documentation example how to identify sechub webscans #3649

### LINKED ISSUE #3649: Describe way to identifiy SecHub DAST scans
## Situation
With #3644 we wanted a dedicated identifier field inside sechub configuration.

But it is enough to just document that inside a web scan it is possible to define custom headers.

## Wanted
It shall be clear for users how to define custom headers to determine scan is from SecHub or an attacker

## Solution
Update documentation

winzj:
- closes #3649


haerter-tss (approved):
Looks good! 🍒🏝️

PR #3778, opened 2025-01-08
Fixed ArchUnit tests

### LINKED ISSUE #3713: ArchUnit refactore code, fix tests and test imports
## Situation
The ArchUnit tests are flaky, because they wok on the build level. While e.g. the eclipse build builds the whole project (test classes and main classes), gradle and intelliJ build, builds only the dependent classes. Further some tests seem not to work properly.

## Wanted
Working ArchUnit tests wihtout side effects on different build tools.

## Solution 
1. We need to fix the resideInPackage with ..test.. (no working properly)
2. We need to fix the execution of the unit tests, test classes are only available when tests were executed before executing the archunit tests:
Solution 1: write ArchUnitExtension annotation for Junit test,
test are only executed when sechub.archunit.buildfolder is set,
GH action: ./gradlew :sechub-archunit-test:test after build,
add shell command to sdc.sh
Solution 2: we might can fix this by adjusting the gradle build execution order and dependencies
Solution 3: There might be an archunit gradle plugin?
3. We need to document if side effects still exists (e.g. with .because, when to use "sechub.archunit.ignoreFolders")

lorriborri:
- closes #3713


sechub-commons-core/src/test/java/com/mercedesbenz/sechub/commons/core/resilience/ResilientRunnableExecutorTest.java
```diff
@@ -38,27 +38,27 @@ void one_retry_defined__exception_on_first_run_but_not_on_seconds_try() throws E
...
-        FailDefinedTimes failDefinedTimes = new FailDefinedTimes(1);
+        FailDefinedTestTimes failDefinedTestTimes = new FailDefinedTestTimes(1);
 
         /* execute */
-        executorToTest.executeResilient(failDefinedTimes);
+        executorToTest.executeResilient(failDefinedTestTimes);
 
         /* test */
-        assertEquals(2, failDefinedTimes.runs);
+        assertEquals(2, failDefinedTestTimes.runs);
```
hamidonos, line 47: you might wanna switch to AssertJ for declarative assertions

It's much more fluent than the old assertEquals method

e.g. `assertThat(failDefinedTestTimes.runs).isEqualTo(2)`

sechub-commons-core/src/test/java/com/mercedesbenz/sechub/commons/core/resilience/ResilientRunnableExecutorTest.java
```diff
@@ -38,27 +38,27 @@ void one_retry_defined__exception_on_first_run_but_not_on_seconds_try() throws E
...
     void one_retry_defined__exception_on_first_and_second_run_will_fail_with_exception() throws Exception {
         /* prepare */
         executorToTest.add(new AlwaysRetry1TimeWith10MillisecondsTestConsultant());
 
-        FailDefinedTimes failDefinedTimes = new FailDefinedTimes(2);
+        FailDefinedTestTimes failDefinedTestTimes = new FailDefinedTestTimes(2);
 
         /* execute */
-        assertThrows(IOException.class, () -> executorToTest.executeResilient(failDefinedTimes));
+        assertThrows(IOException.class, () -> executorToTest.executeResilient(failDefinedTestTimes));
```
hamidonos, line 58: same here

```
    assertThatThrownBy(() -> executorToTest.executeResilient(failDefinedTestTimes))
    .isInstanceOf(IllegalArgumentException.class)
    .hasMessage("Some message to verify that it's really your exception...");
```

hamidonos (changes requested):
I like the new rules for test classes. Good job 👍 

But I'd like you to have a look at my comments:

1)

You are mixing `// ...` and `/* ... */` style of comments
Please use /* ... */` for real comments

2)

I don't see a reason to put a generic warning message behind specific ArchUnit rules (because(...)). Either check for the violations at the end or print a generic warning message at the start/end.

3)

I did not look at every violation here but please make sure that every test class ends with _Test_ and any supporting test class starts with _Test_. Nothing in between should be possible.

4) 

Moving from `assertEquals` to fluent AssertJ ist optional. But it is beneficial in the long run.

de-jcup (approved):
Looks good to me

hamidonos (approved):

PR #3799, opened 2025-01-14
Feature 3789 user can cancel job

### LINKED ISSUE #3789: Provide job cancel operation for User
## Situation
Only admins can cancel jobs (/api/admin/jobs/cancel/{jobUUID})

## Wanted
Everybody that has access to the project (user) should be able to cancel a job

## Solution
Implement new API endpoint for user to cancel job. 
Endpoint /api/jobs/cancel/{jobUUID}
Action should only be executable, if user has access to the project

lorriborri:
- closes #3789

### DIFF
--- a/sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/AdministrationAPIConstants.java
+++ b/sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/AdministrationAPIConstants.java
@@ -128,5 +128,5 @@ public class AdministrationAPIConstants {
     /* +-----------------------------------------------------------------------+ */
     /* +............................... Jobs ..................................+ */
     /* +-----------------------------------------------------------------------+ */
-    public static final String API_USER_CANCEL_JOB = API_JOBS + "cancel/{jobUUID}";
+    public static final String API_USER_CANCEL_JOB = API_MANAGEMENT + "jobs/cancel/{jobUUID}";
 }
--- a/sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/AdministrationAPIConstants.java
+++ b/sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/AdministrationAPIConstants.java
@@ -126,7 +126,7 @@ public class AdministrationAPIConstants {
     public static final String API_REQUEST_NEW_APITOKEN = API_ANONYMOUS + "refresh/apitoken/{emailAddress}";
 
     /* +-----------------------------------------------------------------------+ */
-    /* +............................ Jobs ................................+ */
+    /* +............................... Jobs ..................................+ */
     /* +-----------------------------------------------------------------------+ */
     public static final String API_USER_CANCEL_JOB = API_JOBS + "cancel/{jobUUID}";
 }
--- a/sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/AdministrationAPIConstants.java
+++ b/sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/AdministrationAPIConstants.java
@@ -128,5 +128,5 @@ public class AdministrationAPIConstants {
     /* +-----------------------------------------------------------------------+ */
     /* +............................... Jobs ..................................+ */
     /* +-----------------------------------------------------------------------+ */
-    public static final String API_USER_CANCEL_JOB = API_MANAGEMENT + "jobs/cancel/{jobUUID}";
+    public static final String API_USER_CANCEL_JOB = API_MANAGEMENT + "jobs/{jobUUID}/cancel";
 }
--- a/sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/AdministrationAPIConstants.java
+++ b/sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/AdministrationAPIConstants.java
@@ -1,8 +1,7 @@
 // SPDX-License-Identifier: MIT
 package com.mercedesbenz.sechub.domain.administration;
 
-import static com.mercedesbenz.sechub.sharedkernel.security.APIConstants.API_ADMINISTRATION;
-import static com.mercedesbenz.sechub.sharedkernel.security.APIConstants.API_ANONYMOUS;
+import static com.mercedesbenz.sechub.sharedkernel.security.APIConstants.*;
 
 public class AdministrationAPIConstants {
 
@@ -126,4 +125,8 @@ public class AdministrationAPIConstants {
     public static final String API_FETCH_NEW_API_TOKEN_BY_ONE_WAY_TOKEN = API_ANONYMOUS + "apitoken";
     public static final String API_REQUEST_NEW_APITOKEN = API_ANONYMOUS + "refresh/apitoken/{emailAddress}";
 
+    /* +-----------------------------------------------------------------------+ */
+    /* +............................ Jobs ................................+ */
+    /* +-----------------------------------------------------------------------+ */
+    public static final String API_USER_CANCEL_JOB = API_JOBS + "cancel/{jobUUID}";
 }
--- a/sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/job/JobCancelService.java
+++ b/sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/job/JobCancelService.java
@@ -58,7 +58,6 @@ public class JobCancelService {
         informCancelJobRequested(message);
     }
 
-    @Validated
     @UseCaseUserCancelsJob(@Step(number = 2, name = "Cancel job", description = "Will trigger event that job cancel requested"))
     public void userCancelJob(UUID jobUUID, String userId) {
         userInputAssertion.assertIsValidJobUUID(jobUUID);
@@ -74,7 +73,10 @@ public class JobCancelService {
     }
 
     private void assertUserAllowedCancelJob(UUID jobUUID, String userId) {
-        JobInformation jobInfo = jobInformationRepository.findById(jobUUID).orElseThrow(() -> new NotFoundException("Job not found: " + jobUUID));
+        JobInformation jobInfo = jobInformationRepository.findById(jobUUID).orElseThrow(() -> {
+            LOG.debug("Job not found: {}", jobUUID);
+            return new NotFoundException("Job not found: " + jobUUID);
+        });
 
         User user = userRepository.findOrFailUser(userId);
         for (Project project : user.getProjects()) {
@@ -82,6 +84,7 @@ public class JobCancelService {
                 return;
             }
         }
+        LOG.debug("User {} is not allowed to cancel job {}", userId, jobUUID);
         throw new NotFoundException("Job not found: " + jobUUID);
     }
 
--- a/sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/job/JobCancelService.java
+++ b/sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/job/JobCancelService.java
@@ -2,12 +2,12 @@
 package com.mercedesbenz.sechub.domain.administration.job;
 
 import java.util.Optional;
+import java.util.Set;
 import java.util.UUID;
 
 import org.slf4j.Logger;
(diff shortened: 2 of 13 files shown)


sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/job/JobCancelService.java
```diff
@@ -22,47 +23,74 @@
...
 
         /* trigger event */
         informCancelJobRequested(message);
     }
 
+    private void assertUserAllowedCancelJob(UUID jobUUID, String userId) {
+        JobInformation jobInfo = jobInformationRepository.findById(jobUUID).orElseThrow(() -> new NotFoundException("Job not found: " + jobUUID));
+
+        User user = userRepository.findOrFailUser(userId);
+        for (Project project : user.getProjects()) {
```
hamidonos, line 80: This could be an improvement to your solution

```
Set<Project> projects = user.getProjects() == null ? Set.of() : user.getProjects();
boolean noneMatch = projects.stream().noneMatch(project -> project.getId().equals(jobInfo.getProjectId()));
if (noneMatch) {
    throw new ...
}
```

(I did not test this)

sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/job/JobCancelService.java
```diff
@@ -22,47 +23,74 @@
...
+        this.jobInformationRepository = jobInformationRepository;
+        this.userRepository = userRepository;
+    }
 
-    @Autowired
-    AuditLogService auditLogService;
-
-    @Autowired
-    UserInputAssertion assertion;
+    @Validated
```
hamidonos, line 48: `@Validated` is pointless if Spring Bean validation is not used

sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/job/JobCancelService.java
```diff
@@ -22,47 +23,74 @@
...
+        auditLogService.log("User {} requested cancellation of job {}", userId, jobUUID);
+        assertUserAllowedCancelJob(jobUUID, userId);
 
         JobMessage message = buildMessage(jobUUID);
 
         /* trigger event */
         informCancelJobRequested(message);
     }
 
+    private void assertUserAllowedCancelJob(UUID jobUUID, String userId) {
```
hamidonos, line 76: You throw the same exception in all cases, which is ok

I think it'd be helpful to have debug logs for the different cases

sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/job/JobRestController.java
```diff
@@ -0,0 +1,39 @@
...
+
+    public JobRestController(UserContextService userContextService, JobCancelService jobCancelService) {
+        this.userContextService = userContextService;
+        this.jobCancelService = jobCancelService;
+    }
+
+    @UseCaseUserCancelsJob(@Step(number = 1, name = "Rest call", description = "Triggers job cancellation request, owners of project will be informed", needsRestDoc = true))
+    @RequestMapping(path = AdministrationAPIConstants.API_USER_CANCEL_JOB, method = RequestMethod.POST, produces = { MediaType.APPLICATION_JSON_VALUE })
+    @RolesAllowed({ RoleConstants.ROLE_USER, RoleConstants.ROLE_SUPERADMIN, RoleConstants.ROLE_OWNER })
+    public void userCancelJob(@PathVariable(name = "jobUUID") UUID jobUUID) {
```
hamidonos, line 34: Controller methods with `void` return type should return the HttpStatus `204 NO_CONTENT`

Spring offers a `@ResponseStatus` annotation for that

sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/job/JobRestController.java
```diff
@@ -0,0 +1,39 @@
...
+    private final UserContextService userContextService;
+    private final JobCancelService jobCancelService;
+
+    public JobRestController(UserContextService userContextService, JobCancelService jobCancelService) {
+        this.userContextService = userContextService;
+        this.jobCancelService = jobCancelService;
+    }
+
+    @UseCaseUserCancelsJob(@Step(number = 1, name = "Rest call", description = "Triggers job cancellation request, owners of project will be informed", needsRestDoc = true))
+    @RequestMapping(path = AdministrationAPIConstants.API_USER_CANCEL_JOB, method = RequestMethod.POST, produces = { MediaType.APPLICATION_JSON_VALUE })
```
hamidonos, line 32: There is a shorter version of this `@PostMapping`

sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/job/JobRestController.java
```diff
@@ -0,0 +1,39 @@
...
+import com.mercedesbenz.sechub.domain.administration.AdministrationAPIConstants;
+import com.mercedesbenz.sechub.sharedkernel.Step;
+import com.mercedesbenz.sechub.sharedkernel.security.RoleConstants;
+import com.mercedesbenz.sechub.sharedkernel.security.UserContextService;
+import com.mercedesbenz.sechub.sharedkernel.usecases.job.UseCaseUserCancelsJob;
+
+import jakarta.annotation.security.RolesAllowed;
+
+@RestController
+public class JobRestController {
```
hamidonos, line 21: nit: can be package private

sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/AdministrationAPIConstants.java
```diff
@@ -126,4 +125,8 @@ private AdministrationAPIConstants() {
     public static final String API_FETCH_NEW_API_TOKEN_BY_ONE_WAY_TOKEN = API_ANONYMOUS + "apitoken";
     public static final String API_REQUEST_NEW_APITOKEN = API_ANONYMOUS + "refresh/apitoken/{emailAddress}";
 
+    /* +-----------------------------------------------------------------------+ */
+    /* +............................ Jobs ................................+ */
```
hamidonos, line 129: not sure if you're missing a few characters here? 

the format seems wrong

sechub-administration/src/test/java/com/mercedesbenz/sechub/domain/administration/job/JobCancelServiceTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+import com.mercedesbenz.sechub.domain.administration.project.Project;
+import com.mercedesbenz.sechub.domain.administration.user.User;
+import com.mercedesbenz.sechub.domain.administration.user.UserRepository;
+import com.mercedesbenz.sechub.sharedkernel.error.NotFoundException;
+import com.mercedesbenz.sechub.sharedkernel.logging.AuditLogService;
+import com.mercedesbenz.sechub.sharedkernel.messaging.DomainMessageService;
+import com.mercedesbenz.sechub.sharedkernel.validation.UserInputAssertion;
+
+public class JobCancelServiceTest {
```
hamidonos, line 23: This test should also test that the other methods are called properly

```
userInputAssertion.assertIsValidJobUUID(jobUUID);
userInputAssertion.assertIsValidUserId(userId);

 auditLogService.log("Requested cancellation of job {}", jobUUID);
auditLogService.log("User {} requested cancellation of job {}", userId, jobUUID);

assertUserAllowedCancelJob(jobUUID, userId);    // <-- You are only testing this

JobMessage message = buildMessage(jobUUID);

/* trigger event */
informCancelJobRequested(message);
```

hamidonos (changes requested):

sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/job/JobCancelService.java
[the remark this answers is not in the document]
lorriborri: I'm not sure, since it is nullable = false? 

@Column(name = COLUMN_USER_PROJECTS, nullable = false)
@ManyToMany(cascade = CascadeType.REFRESH, mappedBy = Project.PROPERTY_USERS, fetch = FetchType.EAGER)
Set<Project> projects;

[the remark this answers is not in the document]
hamidonos: I think it's fine, but bear in mind that these annotations are **database level**. Your access to `user.getProjects()` is **application level**. 
You don't have any guarantee that the JPA annotation is actually enforced, since JPA and Flyway schemas are two separate things.

I also assume that Hibernate does not check the data when it's fetched. But I might be wrong here.
In that case you will have a problem if that annotation hasn't been there always ;) 

As long as the property doesn't have a default value like `Set<Project> projects = new HashSet<>()` you can never be sure.

You can leave it as it is but I'd rather include a simple null check. **Defensive programming** doesn't hurt here.

de-jcup: I see here a complete other problem:

Only administrators have access to the administration domain!
But the implementation for normal users is done inside administration domain.

A "normal" user has only access to the scheduler domain!

Inside the Scheduler domain we already a service to assert user project access:
See `ScheduleAssertService#assertUserHasAccessToProject`

The `ScheduleAssertService` has also `assertProjectAllowsWriteAccess` which could be relevant as well (we provide different access levels for projects: write, read (only), none). 

The cancel REST controller implementation for the user must be done inside the scheduler domain. This implementation does the assertion. If all is okay, there must be a cancel request event triggered to domain event bus. This should be (hopefully) enough.

Unit tests should just check that the event is triggered when the user cancel operation is wanted by - e.g. `UserJobCancelService`. We should test here also that the assert service is used. We could also simulate an assertion error here and that no event is triggered.

Integration test could be similar to existing integration tests for administrator calls.

de-jcup: Remark: If it is not clear inside architecture documentation that the scheduler domain is the only domain a "normal" user has access we should add this to technical and architecture documentation.

lorriborri: Team discussion today:
- administration domain is not a domain for admins, but to manage sechub user, projects ...
- we will rename the module in future tasks
- since we plan a self service also users should be able to create projects etc. 

=> we introduce /api/management/

sechub-openapi-java/src/main/resources/openapi.yaml
```diff
@@ -4171,4 +4173,29 @@ paths:
         "404":
           description: "Not found"
       tags:
-        - Configuration
\ No newline at end of file
+        - Configuration
+
+  /api/management/jobs/cancel/{jobUUID}:
```
hamidonos, line 4178: It should be like this:

`/api/management/jobs/{jobUUID}/cancel`

The resource identifier always comes after the resource path

sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/AdministrationAPIConstants.java
```diff
@@ -126,4 +125,8 @@ private AdministrationAPIConstants() {
     public static final String API_FETCH_NEW_API_TOKEN_BY_ONE_WAY_TOKEN = API_ANONYMOUS + "apitoken";
     public static final String API_REQUEST_NEW_APITOKEN = API_ANONYMOUS + "refresh/apitoken/{emailAddress}";
 
+    /* +-----------------------------------------------------------------------+ */
+    /* +............................... Jobs ..................................+ */
+    /* +-----------------------------------------------------------------------+ */
+    public static final String API_USER_CANCEL_JOB = API_MANAGEMENT + "jobs/cancel/{jobUUID}";
```
hamidonos, line 131: `jobs/{jobUUID}/cancel`

hamidonos (changes requested):

hamidonos (approved):

PR #3801, opened 2025-01-15
Explicitly set geckodriver system property #3800

### LINKED ISSUE #3800: Ensure selenium cannot find a geckodriver on pds-owaspzap
### Situation
For authentication with multiple steps we implemented scripts that use selenium inside the pds-owaspzap solution.
It seems like selenium tries to download the latest version of the firefox-geckodriver if it cannot be found locally. Sometimes this is not possible or not wanted.

### Solution
Ensure the firefox-geckodriver can be found by selenium to avoid implicit installation of the latest version.

#### First option
Install the geckodriver inside the docker image and set the corresponding system property.

#### Second option
ZAP already comes with a geckodriver because the necessary plugins are installed. The geckodriver already available because of ZAP can be reused by setting the corresponding system property.

winzj:
- closes #3800

### DIFF
--- a/sechub-pds-solutions/owaspzap/docker/scripts/owasp-zap.sh
+++ b/sechub-pds-solutions/owaspzap/docker/scripts/owasp-zap.sh
@@ -134,7 +134,7 @@ echo ""
 echo "Start scanning"
 echo ""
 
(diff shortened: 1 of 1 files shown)


haerter-tss (approved):
Looks good! 🥐🪄

PR #3809, opened 2025-01-17
implement user self service api for user detail information #3723

### LINKED ISSUE #3723: Add New API Call for User Profile
## Situation
Currently, our application lacks an endpoint to retrieve user profile information (username, email, Role). Users need a way to access their profile details through a dedicated API call. 

## Wanted
We need a new API endpoint 
e.g. `api/user/profile` that allows authenticated users to retrieve their profile information. 

The endpoint should:
- Be accessible via a GET request.
- Return user profile details such as username, email and role.

## Solution

The new Endpoint should be in a dedicated domain for the webui
(maybe move the getAssignedProjectDataList in this domain as well - the usecase must be moved anyway)

- add GET request to openapi.yaml
- create new UseCase 
- create new REST controller
- create new Service

hamidonos:
- closes #3723


sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/user/UserDetailInformation.java
```diff
@@ -20,15 +20,15 @@ public class UserDetailInformation {
...
-    private boolean superAdmin;
+    private final boolean superAdmin;
 
-    private List<String> projects = new ArrayList<>();
+    private final List<String> projects = new ArrayList<>();
 
-    private List<String> ownedProjects = new ArrayList<>();
+    private final List<String> ownedProjects = new ArrayList<>();
 
     public UserDetailInformation(User user) {
```
lorriborri, line 33: (Optional) I know it's not your responsibility, but you might want to revise the constructor.
Again, we have the case here where we assume that the getter methods don't return null. 
But as we've already established, we can't trust that.

sechub-administration/src/test/java/com/mercedesbenz/sechub/domain/administration/user/UserDetailInformationServiceTest.java
```diff
@@ -0,0 +1,129 @@
...
+        /* test */
+        assertThat(userDetailInformation).isNotNull();
+        assertThat(userDetailInformation.getUserId()).isEqualTo(USER_ID);
+        assertThat(userDetailInformation.getEmail()).isEqualTo(EMAIL_ADDRESS);
+        assertThat(userDetailInformation.getProjects()).isNotEmpty();
+        assertThat(userDetailInformation.getProjects()).isEqualTo(getProjectIds(PROJECTS));
+        assertThat(userDetailInformation.getOwnedProjects()).isNotEmpty();
+        assertThat(userDetailInformation.getOwnedProjects()).isEqualTo(getProjectIds(OWNED_PROJECTS));
+        assertThat(userDetailInformation.isSuperAdmin()).isTrue();
+        verify(userContextService).getUserId();
```
lorriborri, line 58: (Optional) You could add times(1) to the verify tests to make clear that the method only executed once.
But since every method is only executed once and its not very complex you can leave it.
 e.g.:
```suggestion
        verify(userContextService, times(1)).getUserId();
```

Or maybe add the verifyNoMoreInteractions method at the end ?(https://pub.dev/documentation/mocktail/latest/mocktail/verifyNoMoreInteractions.html )

lorriborri (changes requested):
Looks good to me! Good work! :star: 🦉

I found only one minor issue and some optional issues you might want to look at:

Minor Issue:

- usage of primitives in data transfer objects

Optional Issues:

- NullPointer proove
- Mockito verify tests

lorriborri:
And if you want to fix the build, you need to fix the UserRestControllerRestDocTest: use instead of the AuthenticationHelper, the TestAuthenticationHelper (It was renamed in the ArchUnit fix)  🙂

lorriborri (approved):
Looks Good! Well done! :star: 🐘

PR #3821, opened 2025-01-22
User self service: change own mail address

### LINKED ISSUE #3583: User shall be able to change his/her mail adress by REST API
## Situation
Currently only an administrator is able to change the user mail address

## Wanted
A user shall be able to change his/her email address via REST.
The process shall ensure that the user has access to the mail address.

## Solution
Should also work with go client
The process shall be similar to the registration process

REST: 
1. User request mail change over web-ui or go client with initial authenticated request:
POST /api/mamangement/user/email/
          body={newEmail}
2. SecHub Server checks request and sends verification link to new mail address (link valid for 24h)
- generates JWT token
- encrypts new  email address
- sends link: POST /user/change-email?accessToken=<token-encrypted>&email=<new-email-encrypted>
- URL must be open to unauthenticated
3. When user verifies the new mail, the mail is changed by the server 
4.  The user receives 204
5. A notification is send to the old mail address

### LINKED ISSUE #3947: Redirect after user verifies email change
## Situation
Currently we receive only a 204 after the mail verification was successful

## Wanted
Redirect to web-ui hidden page

## Solution
Create hidden page /user/email_verify
Redirect to frontend after successful mail change

lorriborri:
- closes #3583 
- closes #3947


sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/user/UserEmailAddressUpdateService.java
```diff
@@ -88,4 +168,20 @@ private void informUserEmailAddressUpdated(UserMessage message) {
...
 
+    @IsSendingAsyncMessage(MessageID.USER_EMAIL_ADDRESS_CHANGE_REQUEST)
+    private void informUserWantsToChangeEmailAddress(UserMessage message) {
+        DomainMessage infoRequest = new DomainMessage(MessageID.USER_EMAIL_ADDRESS_CHANGE_REQUEST);
+        infoRequest.set(MessageDataKeys.USER_EMAIL_ADDRESS_CHANGE_DATA, message);
+
+        eventBusService.sendAsynchron(infoRequest);
+    }
+
+    private static UserMessage getUserMessage(User user, String formerEmailAddress) {
```
hamidonos, line 179: this is a create method
`createUserMessage`

sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/user/UserRestController.java
```diff
@@ -28,4 +31,12 @@ public UserDetailInformation fetchUserDetailInformation() {
         /* @formatter:on */
         return userDetailInformationService.fetchDetails();
     }
+
+    /* @formatter:off */
+    @UseCaseUserUpdatesEmailAddress(@Step(number = 1, name = "Rest call", description = "User request to update his email address", needsRestDoc = true))
+    @PostMapping (AdministrationAPIConstants.API_USER_EMAIL+ "/{emailAddress}")
```
hamidonos, line 37: Return HttpStatus `NO_CONTENT` here

hamidonos (changes requested):

hamidonos (changes requested):

sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/user/AnonymousUserRestController.java
```diff
@@ -0,0 +1,85 @@
...
+
+        String redirectUri = loginProperties.getRedirectUri();
+        if (redirectUri == null) {
+            return DEFAULT_REDIRECT_URI;
+        }
+
+        try {
+            String baseUrl = getBaseUri(redirectUri);
+            return baseUrl + FRONTEND_REDIRECT_URI_PATH;
+        } catch (URISyntaxException e) {
```
hamidonos, line 63: This is a rather big error case (misconfiguration of the application)

I think we should not swallow this exception and handle with a default value

Use a error log for such cases and throw the exception

We should know about it

hamidonos: Btw: If you choose to define the redirect uri in the properties you should probably move the URI validation there as well

sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/user/UserEmailChangeTokenService.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.Base64;
+
+import org.springframework.stereotype.Service;
+
+import com.mercedesbenz.sechub.sharedkernel.error.BadRequestException;
+import com.mercedesbenz.sechub.spring.security.AES256Encryption;
+
+@Service
+public class UserEmailChangeTokenService {
+    private static final long EXPIRATION_TIME = 1000 * 60 * 60 * 24; // 24 hours
```
hamidonos, line 14: You can use the `Duration` class to safely define your durations/periods.

E.g.: `EXPIRATION_DURATION = Duration.ofHours(24)`

hamidonos: It's always helpful to suffix your time variables with the type if you manually calculate it like this

here it would be 

`private static final long EXPIRATION_TIME_MILLIS = ....`

hamidonos (changes requested):

sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/user/AnonymousUserRestController.java
```diff
@@ -0,0 +1,85 @@
...
+@RestController
+public class AnonymousUserRestController {
+
+    private final UserEmailAddressUpdateService emailAddressUpdateService;
+    private final SecHubSecurityProperties secHubSecurityProperties;
+    private final String REDIRECT_URI;
+    // todo: we should define a default redirect uri which is always available
+    // (/login is not)
+    private final String DEFAULT_REDIRECT_URI = "/api/anonymous/check/alive";
+    private final String FRONTEND_REDIRECT_URI_PATH = "/user/email_verify";
```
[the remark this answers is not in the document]
lorriborri: Decided to go with the login redirect URL (less configurations for the user),
if the URL is not set, response is 204

de-jcup (changes requested):
Good job,but I found some parts - see my comments.
I will review further when re-request the review.

sechub-administration/src/test/java/com/mercedesbenz/sechub/domain/administration/user/AnonymousUserRestControllerTest.java
```diff
@@ -0,0 +1,57 @@
...
+import com.mercedesbenz.sechub.sharedkernel.Profiles;
+import com.mercedesbenz.sechub.spring.security.SecHubSecurityProperties;
+import com.mercedesbenz.sechub.test.TestPortProvider;
+
+@RunWith(SpringRunner.class)
+@WebMvcTest
+@ContextConfiguration(classes = { AnonymousUserRestController.class })
+@ActiveProfiles({ Profiles.TEST })
+@Import(TestAdministrationSecurityConfiguration.class)
+public class AnonymousUserRestControllerTest {
```
de-jcup, line 32: We have a naming convention for tests which do need spring boot container running (to identify the "potentially slow ones")

$xyzSpringBootTest

Please rename to
```suggestion
public class AnonymousUserRestControllerSpringBootTest {
```

sechub-administration/src/test/java/com/mercedesbenz/sechub/domain/administration/user/AnonymousUserRestControllerTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+    private static final int PORT_USED = TestPortProvider.DEFAULT_INSTANCE.getWebMVCTestHTTPSPort();
```
de-jcup, line 34: Hm... normally we have integration tests here which do test this.
But I see this is faster and we test here that no content is returned when
...

and that is unclear: 
userEmailAddressUpdateService is injected as mock, okay, but what kind of behavior does lead to the situation that no content is delivered back here? The kind of behavior is not mentioned here - I would expect something like


when(userEmailAddressUpdateService.xyz).thenReturn(...); // we simulate service behaves....

Also a verify(userEmailAddressUpdateService).xyz would be nice.

sechub-administration/src/test/java/com/mercedesbenz/sechub/domain/administration/user/AnonymousUserRestControllerTest.java
```diff
@@ -0,0 +1,57 @@
...
+    @Autowired
+    private MockMvc mockMvc;
+    @MockBean
+    private UserEmailAddressUpdateService userEmailAddressUpdateService;
+    @MockBean
+    private SecHubSecurityProperties secHubSecurityProperties;
+
+    @Test
+    @WithAnonymousUser
+    public void anonymous_user_verifies_new_mail_address_with_one_time_token() throws Exception {
```
de-jcup, line 45: There should be a test case where the user verifies the token and the service says "ok" and what kind of status is expected in this case.

or - other option : just test that the result from the service is returned always (a parameterized test would be a good option here)

sechub-administration/src/test/java/com/mercedesbenz/sechub/domain/administration/user/UserRestControllerTest.java
```diff
@@ -1,11 +1,12 @@
...
 import static com.mercedesbenz.sechub.test.SecHubTestURLBuilder.https;
 import static org.hamcrest.CoreMatchers.equalTo;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
-import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
+import static org.mockito.Mockito.*;
+import static org.springframework.security.test.web.servlet.request.SecurityMockMvcRequestPostProcessors.csrf;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.*;
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.jsonPath;
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
```
de-jcup, line 11: Not your fault, but please rename the class to
`UserRestControllerSpringBootTest`

sechub-integrationtest/src/test/java/com/mercedesbenz/sechub/integrationtest/scenario2/UserChangeMailScenario2IntTest.java
```diff
@@ -0,0 +1,55 @@
...
+        /* test */
+        String link = getLinktToVerifyEmailAddressAfterChangeRequest(newEmail, USER_1);
+        assertThat(link).isNotNull();
+
+        // workaround to avoid double encoding of the link
+        URI uri = null;
+        try {
+            uri = new URI(link);
+        } catch (Exception e) {
+            e.printStackTrace();
```
de-jcup, line 44: Please don't do such things! (catch + printstacktrace + go further)

Why do we here a e.printStacktrace? It is also not clear why this is a workaround for what... 

Please add `@throws Exception` to the test method and just do Uri uri = new Uri(link). If there is an exception there is something odd and the test must fail.

Remark: It's possible to use junit5 extension `IntegrationTestExtension` instead of the old junit4 rules. Consider to use the new way.

sechub-integrationtest/src/test/java/com/mercedesbenz/sechub/integrationtest/scenario2/UserChangeMailScenario2IntTest.java
```diff
@@ -0,0 +1,55 @@
...
+import java.net.URI;
+
+import org.junit.Rule;
+import org.junit.Test;
+
+import com.mercedesbenz.sechub.integrationtest.api.IntegrationTestSetup;
+import com.mercedesbenz.sechub.integrationtest.api.TestUserDetailInformation;
+import com.mercedesbenz.sechub.integrationtest.api.TextSearchMode;
+
+public class UserChangeMailScenario2IntTest {
```
de-jcup, line 20: Thank you. Integration test is really necessary for the changes and having an own dedicated one is a good idea here.

de-jcup (changes requested):
Please see my comments.

sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/user/UserEmailChangeTokenService.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.Base64;
+
+import org.springframework.stereotype.Service;
+
+import com.mercedesbenz.sechub.sharedkernel.error.BadRequestException;
+import com.mercedesbenz.sechub.spring.security.AES256Encryption;
+
+@Service
+public class UserEmailChangeTokenService {
+    private static final long EXPIRATION_TIME_MILLIS = Duration.ofHours(24).toMillis();
```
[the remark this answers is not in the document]
lorriborri: new Issue: https://github.com/mercedes-benz/sechub/issues/3948

sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/user/UserRestController.java
```diff
@@ -28,4 +32,18 @@ public UserDetailInformation fetchUserDetailInformation() {
         /* @formatter:on */
         return userDetailInformationService.fetchDetails();
     }
+
+    /* @formatter:off */
+    @UseCaseUserUpdatesEmailAddress(
+            @Step(number = 1,
+                    name = "Rest call",
+                    description = "User request to update his email address",
```
de-jcup, line 40:
```suggestion
                    description = "User request to update their email address",
```

sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/user/UserTransactionService.java
```diff
@@ -0,0 +1,29 @@
...
+
+    public UserTransactionService(UserRepository userRepository) {
+        this.userRepository = userRepository;
+    }
+
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
+    public void saveInOwnTransaction(User user) {
+        requireNonNull(user);
+
+        userRepository.save(user);
```
de-jcup, line 25: Please change the method signature to return the user object!

Reason:
When you look at the javadoc of the called save method you read:

> Saves a given entity. **Use the returned instance** for further operations as the save operation might have changed the entity instance completely.

An example for correct implementation is `ProjectTransactionService`.

On the caller side you have to use the returned object for further interactions.
Reading your code it seems you do no further interactions so it does not matter.
But because we provide here a service which can be used in future by others we need to return the object.

There are implementations like `JobRunStatisticTransactionService` which have also only void as return value in their methods, but only because they have no parameters which are changed by the method: They internally create the new object and do not give it back. Here we have a user object which is given as parameter and maybe no longer be valid/used by JPA implementation (e.g. hibernate) under the hood - see remark from javadoc of save method.

sechub-administration/src/test/java/com/mercedesbenz/sechub/domain/administration/UserRepositoryDBTest.java
```diff
@@ -60,46 +54,34 @@ public void before() {
...
         Project project = TestProjectCreationFactory.createProject("project_repo_test1", user1);
-
         entityManager.persist(project);
 
-        /* execute */
-        userRepository.deleteUserWithAssociations(user1.getName());
-
+        /* execute & test */
+        assertThatThrownBy(() -> userRepository.deleteUserWithAssociations(user1.getName())).isInstanceOf(NotAcceptableException.class)
+                .hasMessage("User db_test_testuser1 is 1 times still owner of a project! Move ownership before delete!");
```
de-jcup, line 80: Just a remark:
`.hasMessage(...)` versus `.hasMessageContaining(...)`: 

It is okay to use has message ... but most times I would prefer `.hasMessageContaining(...)` and to check only a relevant part.

Why? Because of maintenance. If you change just a bit of the sentence - e.g. "Move owner ship before delete!" would be changed to "Move all ownerships before you deltee the user!" the test will fail. But the test shall only test that there is a not acceptable exception thrown which complains that still owner of project.

So an alternative would have been:
```suggestion
                .hasMessageContaining("db_test_testuser1 is 1 times still owner");
```
It is also possible to combine with relevant parts like:
```suggestion
                .hasMessageContaining("db_test_testuser1").hasMessageContaining("1 times still owner");
```

If you would write tests this way, they would become more easy to maintain and concentrate not on the exact test but on the text essence.

sechub-administration/src/test/java/com/mercedesbenz/sechub/domain/administration/UserRepositoryDBTest.java
```diff
@@ -132,7 +114,28 @@ public void user_NOT_being_owner_of_project_but_listed_as_user_can_be_deleted__a
...
+    }
+
+    @Test
+    void existsByEmailIgnoreCase_returns_true_when_email_exists_in_database() {
+        /* prepare */
+        String mail1 = user1.getEmailAddress();
+        String mail2 = user2.getEmailAddress();
+        String mail3 = mail1.toUpperCase();
+        String mail4 = mail2.toUpperCase();
+        String mail5 = mail1.replace("example", "exaMplE");
```
de-jcup, line 131: I think it is a good way to change the existing mail address to a case randomized variant. But please add following here

```java
/* check preconditions */
assertThat(mail5.contains("exaMplE")).isTrue(); // ensure example mail address is as expected...

```
why ? You rely on the situation that the test helper class created a mail address containing always "example" inside but... somebody could change the Test helper class... 
So it is necessary to check the precondition.

de-jcup (approved):
Looks good to me :rocket: Let's merge it

PR #3822, opened 2025-01-24
Template healthcheck + integration test improvements and fixes #3616

### LINKED ISSUE #3616: Introduce Templates healthcheck
## Situation
With the "template an assets" feature SecHub is able to handle different resources at runtime.
It is possible to remove them, assign different assets to projects which have same product setup - so we are able to customize behavior when necessary at runtime.

While the PDS template prarameter is constructed, the SecHub server does automatically check if the template asset does contains a file `${PdsProductIdentifier}.zip` inside database.
If it is not available, the Server will throw an exception and the job will fail (without calling a PDS instance at all).

The situation happens only at runtime and it is not easy to understand as an administrator at the first glance why there is a problem etc.

## Wanted
There shall be a possibility to check if the current situation of our templates setup could lead to problems and how to solve them

## Solution
We provide a REST endpoint for administrators:
`/api/admin/templates/healthcheck`

The call will inspect all asociations between

project->Template -> asset
project->Profile->Executor (product identifier) + enabling states

When `assets/${assetId}/${productIdentifier}.zip` does not exist for a template, the validation will mark this as serious problem and will suggest to upload an empty zipfile as a workaround.

The result of the rest call shall be JSON:

```json
{
  "status" : "ERROR", /* enum with: "OK|ERROR" */
  "assetFilesMissing" : [
      "template" : "template-x",
      "asset" : "asset-1",
      "file" : "$filename",

      "executionProfile" : "$profileId",
      "executorConfiguration" : "$executorId",

      "projects" : ["projectId1","projectId2", ".."],

      "suggestion" : "Create at least an empty zip file '$filename' and upload to asset $assetId"
      
  ] 
}
```

### LINKED ISSUE #3828: Provide JUnit5 extension for integration tests as a replacement for old junit4 rule
## Situation
Currently we have for integration tests only a Junit4 way:

```java

import org.junit.*;

public class XyzScenario1Test {

    @Rule
    public IntegrationTestSetup setup = IntegrationTestSetup.forScenario(Scenario12.class);

    @Test
    public void integrationTestXyz(){
        // ... the integration test
    }
}
```

## Wanted
We want to provide Junit5 way

## Solution
We will provide a Junit5 extension called `IntegrationTestExtension`.
To provide test scenario parameters we introduce: `@WithTestScenario`

Here an example how it will look like when using Junit5: 

```java
import org.junit.jupiter.api.Test;

@ExtendWith(IntegrationTestExtension.class)
@WithTestScenario(Scenario1.class)
class XyzScenario1IntTest {

     @Test
     void integrationTestXyz(){
        // ... as usual
     }
}
```

### LINKED ISSUE #3829: IDE gradle builds cannot start single integration tests or manual tests
## Situation
We have currently the situation that it is not possible to start integration tests from IDE.
The tests are not found - not existing (for the gradle build).

## Analyze
The first problem is that
`*IntTest` and `*ManualTest` are currently filtered for all test tasks by `build-java.gradle`

Means the build system will not find them in normal testing.

Remark: I think the reason for the filtering was, that skipping the integration tests in normal builds and doing an integration test afterwards lead sometimes to the situation that gradle handled the integration test files as "unchanged" and did not execute the tests.

But it it IMO better to do a `:sechub-integrationtest:cleanTest` instead of the former workaround.

## Wanted
Integrationtests shall be found again by IDE gradle toolings (IntelliJ, BuildShip or EGradle)

## Solution 
Drop the filtering

### LINKED ISSUE #3830: Starting single integration test by gradle inside IDE does not use developer settings
## Situation
With 
- #3829

it became clear that the filtering of integration tests is a problem for local IDE gradle based exeuction of tests.

While fixing it, another problem appeared:

`~/.sechub/sechub-developer.properties`
containing:
`sechub.integrationtest.running=true`

were still skipped after the fix.

## Analyze
The reason is, that the files are not fetched in case of gradle builds, means also local gradle builds.
This comes very inconvenient and is also extreme confusing.

## Wanted
A gradle build shall search for the `~/.sechub/sechub-developer.properties` as well.
If its not existing, nothing shall happen.

## Solution
Remove the special gradle handling

### LINKED ISSUE #3833: Drop  the "long running integration tests" feature
## Situation
`sechub.integrationtest.longrunning` was introduced to handle long running integration tests different.

But...
- we do not longer have any integration test which does the use the `markLongRunning()`
- it makes it more complicated to developers to understand what to enable and what not
- it is not really used/useful
- we should always execute all integration tests (which could be the reason why no more longrunning markers used...

## Wanted
Keep it simple and drop unused parts

## Solution
We will remove the complete long running stuff.

This means:

- [ ] remove from gradle build
- [ ] remove from integration test handling
- [ ] remove from documentation

### LINKED ISSUE #3840: Delete template action missing in DAUI
## Situation
There is no "delete template" action inside DAUI

## Solution
Add action

### LINKED ISSUE #3841: Create overview action for templates inside DAUI
## Situation
Inside DAUI tThere is no convenient possibility to have an overview which templates are defined andwhich templates are assigned to which projects.

## Solution
Add an action to create/show an overview about the current template situation.

Just show a list of all templates and their assined project (ids)

de-jcup:
This PR
- closes #3616 
- closes #3828
- closes #3829
- closes #3830
- closes #3833
- closes #3840
- closes #3841


winzj (approved):
Looking all good to me,

thanks for the comments in the code :rocket:

PR #3824, opened 2025-01-27
Add defineFalsePositives step to GH action before scan execution #2078

### LINKED ISSUE #2078: Add "falsePositives" github action
## Situation
With #1443 we have now a GitHub action which is able to scan wit SecHub (thanx to @masterhanddev ) 
We decided to provide here only a `scan` GitHub action at the beginning.

## Wanted
We want another github action `falsePositives` . 

It shall be possible to define false positives by  a commited json file (default `.sechub-false-positives.json` in root folder)

```
- uses mercedes-benz/sechub/github-actions/false
```

## Solution
IMPORTANT: We need 
- #2077
and use the new action to handle this correctly.

winzj:
- closes #2078

### DIFF
--- a/github-actions/scan/action.yml
+++ b/github-actions/scan/action.yml
@@ -57,6 +57,9 @@ inputs:
     description: 'Job will be marked as failed if SecHub finds something'
     required: false
     default: true
+  define-false-positives:
+    description: 'The file that defines false positives. This step is executed before the scan. The SecHub cli defineFalsePositives overwrites the false positives on SecHub side with the ones defined in this file. Use the SecHub cli getFalsePositives to get the current false positives and extend the returned false positive list.'
+    required: false
 
 outputs:
   scan-trafficlight:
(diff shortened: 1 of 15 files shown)


github-actions/scan/action.yml
```diff
@@ -57,6 +57,9 @@ inputs:
     description: 'Job will be marked as failed if SecHub finds something'
     required: false
     default: true
+  define-false-positives:
+    description: 'The file that defines false positives. This step is executed before the scan. The SecHub cli defineFalsePositives overwrites the false positives on SecHub side with the ones defined in this file. Use the SecHub cli getFalsePositives to get the current false positives and extend the returned false positive list.'
```
de-jcup, line 61:
```suggestion
    description: 'The file that defines false positives. This step is executed before the scan. The SecHub cli `defineFalsePositives` overwrites the false positives on SecHub side with the ones defined in this file. Uses the SecHub cli `getFalsePositives` to get the current false positives and extend the returned false positive list.'
```

de-jcup (approved):
Looks good to me :rocket: 

I only had some suggestions for some text parts - please see my comments.

PR #3838, opened 2025-01-30
Fixing authentication mode #3837

### LINKED ISSUE #3837: Web-Ui: fix ENV for using local Basic Auth and OAuth
## Situation
Currently the ENV VITE_API_LOCAL_DEV should manage if the loal basic auth is set or not. The env is a string, but we is read as boolean.

## Wanted
Transform ENV into boolean

## Solution
 LOCAL_DEV: Boolean(import.meta.env.VITE_API_LOCAL_DEV === 'true') || false

lorriborri:
- closes #3837

### DIFF
--- a/sechub-web-ui/src/config.ts
+++ b/sechub-web-ui/src/config.ts
@@ -7,7 +7,7 @@ const config = ref({
   HOST: String(import.meta.env.VITE_API_HOST) || 'http://localhost:3000',
   USERNAME: String(import.meta.env.VITE_API_USER) || '',
   PASSWORD: String(import.meta.env.VITE_API_PASSWORD) || '',
-  LOCAL_DEV: String(import.meta.env.VITE_API_LOCAL_DEV) || 'false',
+  LOCAL_DEV: Boolean(import.meta.env.VITE_API_LOCAL_DEV === 'true') || false,
 })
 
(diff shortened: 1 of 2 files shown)


sechub-web-ui/src/config.ts
```diff
@@ -7,7 +7,7 @@ const config = ref({
   HOST: String(import.meta.env.VITE_API_HOST) || 'http://localhost:3000',
   USERNAME: String(import.meta.env.VITE_API_USER) || '',
   PASSWORD: String(import.meta.env.VITE_API_PASSWORD) || '',
-  LOCAL_DEV: String(import.meta.env.VITE_API_LOCAL_DEV) || 'false',
+  LOCAL_DEV: Boolean(import.meta.env.VITE_API_LOCAL_DEV === 'true') || false,
```
hamidonos, line 10: I don't think this should work

if `import.meta.env.VITE_API_LOCAL_DEV` is of type `boolean` then `import.meta.env.VITE_API_LOCAL_DEV === 'true'` should fail since `===` explicitly checks if the value is type string 

Why not simply write

`LOCAL_DEV: import.meta.env.VITE_API_LOCAL_DEV || false`

?

hamidonos (approved):
👍

PR #3843, opened 2025-02-03
Validate templates on job creation + use sealed objects in domain communication

### LINKED ISSUE #3610: Problems with template definition shall be returned to user on job creation
## Situation
Currently the user has no direct feedback when something is wrong with his template definition, instead only when the SecHub job starts/a PDS job starts, those jobs will fail and an error happens.

But the user has no direct feedback and will only see that there was a technical/internal problem with the job.

## Wanted
When a user defines template data we can have following situations:

- no template defined by administrator -> not possible to handle
- mandatory template variables are missing
- additional/wrong template variable names are used
- the validation setup of the template definition does not accept user input

## Solution
Inside the scheduler domain we check if at least one dataTemplate definition exists inside configuration (we use the existing resolver class for this).
If there is a template data setup found, the scheduler will send an event REQUEST_SECHUB_CONFIGURATION_VALIDATION . The scan domain will receive and handle the event.
as a result there will be a list of error messages (or none).

The event will be handled synchronous.
If the error messages are not empty an exception with user info will happen. Also a dedicated HTTP error code (TBD) will be returned. 

The go client shall be able to handle this accordingly (but this will be done in another issue)

### LINKED ISSUE #3842: Domain messaging shall use sealed objects to store infomation
## Situation
The internal used domain event bus system is currently implemented by a simple observer pattern.
The data for event communication is provided by child implementations of `DomainMessagePart`.

Unfortunately the information is stored as string inside those objects but not in a secure way.
This means on a JVM crash the JVM dump could contain the string information as is.

## Wanted
A JVM crash shall not contain information of domain message parts

## Solution
Use a sealed object to store the string information

de-jcup:
This PR
- [x] closes #3842
- [x] closes #3610

### DIFF
--- a/sechub-commons-core/src/main/java/com/mercedesbenz/sechub/commons/core/CachingPatternProvider.java
+++ b/sechub-commons-core/src/main/java/com/mercedesbenz/sechub/commons/core/CachingPatternProvider.java
@@ -1,0 +1,112 @@
+package com.mercedesbenz.sechub.commons.core;
+
+import static java.util.Objects.*;
+
+import java.time.LocalDateTime;
+import java.util.Map;
+import java.util.SortedSet;
+import java.util.TreeMap;
+import java.util.TreeSet;
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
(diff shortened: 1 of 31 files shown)


winzj (approved):
Looking good to me :+1: 

I have seen that the newly added java files do not contain the license header, maybe you can add them. I think they would be added during the release workflow otherwise anyway.

PR #3851, opened 2025-02-04
Feature implement api token authentication in web UI #3598

### LINKED ISSUE #3598: [WebUI] Implement API Token Authentication in WebUI
## Situation 

Currently the API Token is not used for authentication inside the SecHub WebUI.

## Wanted

We want to enable API Token authentication so that users can either log in with OAuth2 or API Token.

## Solution

- Add Basic Auth login api to SecHub WebUI
- Upon login attempt the WebUI Backend calls the SecHub Server to validate the Basic Auth Credentials
- Make sure the API Token is set in the cookies (encrypted) after successful authentication
- That way we ensure that the user is not logged out in the next request
- The cookie should be alive for the 24h (configured by the backend)

hamidonos:
- closes #3598


de-jcup (changes requested):
First of all: Very good job!
- I added some additional documentation - as discussed.
- I added some additional logging in AbstractSecurityConfiguration
- The implementation works
- I found no security problems

What I found/what is to discuss
- Some namings
- Some unclear parts
- One test is missing (IMO)

Please read my comments

## Additionally:
When I simulate an API token change 
  (which can be requested by user) the web ui did no longer work, (one failure appeared) and afterwards I always got a basic auth header from browser.
I am not sure if this is the right issue to handle/discus this - maybe it completely out of scope here, but I wanted to mention it.
Maybe a logout mechanism to cleanup the classic token + a call from web ui to login again could solve the problem here?

### How I simulated the api token change:
(I logged into a running integration test server via login (classic) and afterwards I restarted the server (but with different api token)

sechub-commons-security-spring/src/main/java/com/mercedesbenz/sechub/spring/security/AbstractSecurityConfiguration.java
```diff
@@ -145,7 +148,7 @@ SecurityFilterChain securityFilterChainLogin(HttpSecurity httpSecurity,
 												 SecHubSecurityProperties secHubSecurityProperties,
 												 RestTemplate restTemplate,
 												 AES256Encryption aes256Encryption,
-												 OAuth2AuthorizedClientService oAuth2AuthorizedClientService) throws Exception {
+												 @Autowired(required = false) OAuth2AuthorizedClientService oAuth2AuthorizedClientService) throws Exception {
```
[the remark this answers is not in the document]
hamidonos: login is also possible in `classic` mode with **no** `oauth2` mode at all

sechub-commons-security-spring/src/test/java/com/mercedesbenz/sechub/spring/security/LoginModeOAuth2ActiveConditionTest.java
```diff
@@ -0,0 +1,119 @@
...
+import com.mercedesbenz.sechub.testframework.spring.YamlPropertyLoaderFactory;
+
+@SpringBootTest
+@TestPropertySource(locations = "classpath:application-login-mode-oauth2-active-condition-test.yaml", factory = YamlPropertyLoaderFactory.class)
+class LoginModeOAuth2ActiveConditionTest {
+
+    private static final ConditionContext conditionContext = mock();
+    private static final Environment environment = mock();
+    private static final AnnotatedTypeMetadata metadata = mock();
+    private static final LoginModeOAuth2ActiveCondition conditionToTest = new LoginModeOAuth2ActiveCondition();
```
de-jcup, line 31: Normally we differ between "normal" unit tests and spring boot tests.
Here it is mixed.

Why is this not so good (IMO)?
Reason: harder to understand whats going on

unit test part: uses the field `conditionToTest` which will use the mocked environment etc.

spring boot test part: this uses the `classpath:application-login-mode-oauth2-active-condition-test.yaml`, uses the `TestConfig` creates a random bean, uses autowired etc. but DOES NOT use the mocked environment etc. (of course,but not obviously on first sight) 

So it was not clear which of the tests did use the yaml file and which one does use the mock etc.

Also we have a naming convention for spring boot tests:
see https://mercedes-benz.github.io/sechub/latest/sechub-techdoc.html#naming-of-unit-tests

So please do me a favor and split the test class into two separated test classes:

`LoginModeOAuth2ActiveConditionTest` (no spring magic) and 
`LoginModeOAuth2ActiveConditionSpringBootTest` (spring parts)

sechub-commons-security-spring/src/test/java/com/mercedesbenz/sechub/spring/security/LoginModeOAuth2ActiveConditionTest.java
```diff
@@ -0,0 +1,119 @@
...
+        boolean result = conditionToTest.matches(conditionContext, metadata);
+
+        /* test */
+        assertThat(result).isTrue();
+    }
+
+    @Test
+    void matches_when_no_mode_enabled_returns_false() {
+        /* prepare */
+        when(environment.getProperty("sechub.security.login.modes[0]")).thenReturn(null);
```
de-jcup, line 84: "when_no_mode_enabled" -->
when(environment.getProperty("sechub.security.login.modes[0]")).thenReturn(null);

hm... Reading the test method name I would expect more a 
`when(environment.getProperty("sechub.security.login.modes")).thenReturn(new String[]);`
or a
`when(environment.getProperty("sechub.security.login.modes")).thenReturn(null);`

de-jcup (approved):
Looks good to me - only a test should be separated into two different test classes. I approved anyway. But pllease read my comments.

PR #3852, opened 2025-02-04
Implemented template delete handling in admin domain #3848

### LINKED ISSUE #3848: Template delete action does not delete project relations at admin domain
## Situation
When an administrator deletes a template, the project template list still contains the template
The template list does not contain the template anymore.

Starting a scan for the project will be a problem because the template is not found.

## Wanted
The template relation must be cleaned up

## Solution
Fix it

de-jcup:
This PR
- closes #3848


winzj (approved):
Looking good :rocket:

PR #3895, opened 2025-02-18
Introduce communcation data converter #3712

### LINKED ISSUE #3712: Introduce communication data converter
## Situation
When we want to communicate with external endpoints (for example: calling a REST service of an AI endpoint) we want to have this configurable.

## Wanted
An easy way to define the way to communicate with an external service in a customizable way.

## Solution
We differ here between `send` and `receive` and provide a communication data converter which can be configured.

### Out of scope
Here we only provide simple key mapping without substructure creation.
Means we do currently not handle complex keys with parent child structures

As an example:
```json
{
    "result" : {
        "status" : "OK",
        "message" : "some message inside child structure"
    }
}
```
which could be provided by a key like "result.status" is currently not supported.

If it becomes necessary we will address this in another issue.

### Converter config syntax
```json
{
    "send" : {
        "targetType" : "$type",
        "mapping" : {
         "$sourceKey" : "$destinationKey"
        }
    },

    "receive" : {
        "sourceType" : "$type",
        "mapping" : {
         "$sourceKey" : "$destinationKey"
        }
    } 
   
}
```

### Example

```json
{
    "send" : {
        "targetType" : "JSON",
        "mapping" : {
          "cweId" : "cwe_id",
          "language" : "lang",
          "details" : "cwe_details"
        }
    },

    "receive" : {
        "sourceType" : "JSON",
        "mapping" : {
          "error" : "result_problem",
          "status" : "result_status",
          "description" : "content"
        }
    } 
}
```

#### Usage

```java
CommunicationDataConverter converter = new CommunicationDataConverter(config);

val requestData = new HashMap<String,String>();
requestData.put("cweId", "79");
requestData.put("language", "java");
requestData.put("details", "some technical details about the cwe problem");

String jsonToSend = converter.convertForSending(data);

String jsonReceived = serviceXyz.fetchLoadByJsonContent(jsonToSend);

Map<String, String> resultMappingData= converter.convertFromReceived(jsonReceived);
String content = resultMappingData.get("content");
String status = resultMappingData.get("result_status")
```

de-jcup:
This PR
- closes #3712

### DIFF
--- a/sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/communication/CommunicationDataConverter.java
+++ b/sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/communication/CommunicationDataConverter.java
@@ -1,0 +1,116 @@
+package com.mercedesbenz.sechub.commons.communication;
+
+import static java.util.Objects.*;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.SortedMap;
+import java.util.TreeMap;
+
+import com.mercedesbenz.sechub.commons.communication.CommunicationDataConverterConfig.Receive;
+import com.mercedesbenz.sechub.commons.communication.CommunicationDataConverterConfig.Send;
+import com.mercedesbenz.sechub.commons.model.JSONConverter;
+
+/**
+ * Converts data for communication between sending and receiving. The way of
+ * conversion can be configured by configuration object.
+ *
+ * A usage example:
+ *
+ * <pre>
+ * <code>
+ * CommunicationDataConverter converter = new CommunicationDataConverter(config);
+ *
+ * val requestData = new HashMap<String,String>();
+ * requestData.put("cweId", finding.getCweId());
+ * requestData.put("language", finding.getLanguage());
+ * requestData.put("snippet", finding.getSnippet());
+ *
+ * String jsonToSend = converter.convertForSending(data);
+ *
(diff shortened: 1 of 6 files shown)


sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/communication/CommunicationDataConverter.java
```diff
@@ -0,0 +1,114 @@
...
+ *
+ */
+public class CommunicationDataConverter {
+
+    private CommunicationDataConverterConfig config;
+
+    /**
+     * Creates the converter
+     *
+     * @param config the configuration to user, may not be <code>null</code>
```
winzj, line 49:
```suggestion
     * @param config the configuration to use, may not be <code>null</code>
```

sechub-commons-model/src/main/java/com/mercedesbenz/sechub/commons/communication/CommunicationDataConverter.java
```diff
@@ -0,0 +1,114 @@
...
+     *
+     * @throws IllegalArgumentException when config is <code>null</code>
```
winzj, line 51: The Javadoc `throws` does not fit the methods behaviour.
```suggestion
     * @throws IllegalArgumentException when config send or receive CommunicationDataConversionType is not supported
```

winzj (approved):
Look good,

I just have one small suggestion for the javadoc.

PR #3918, opened 2025-02-24
Handle SecHub Server 2.6.0 startup and documentation problems #3917

### LINKED ISSUE #3914: SecHub Server fails on startup with unclear message about missing OAuth2Client
## Situation
When we start SecHub server 2.6.0 with a (maybe) wrong configuration, we got the error message:

```java
Exception encountered during context initialization - cancelling refresh attempt: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'org.springframework.security.config.annotation.web.configuration.WebSecurityConfiguration': 

Unsatisfied dependency expressed through method 'setFilterChains' 
parameter 0: Error creating bean with name 'securityFilterChainLogin' defined in 
class path resource [com/mercedesbenz/sechub/sharedkernel/security/SecHubSecurityConfiguration.class]: 

Failed to instantiate [org.springframework.security.web.SecurityFilterChain]: Factory method 'securityFilterChainLogin' threw exception with message: 

No qualifying bean of type 'org.springframework.security.oauth2.client.OAuth2AuthorizedClientService' available

```

## Wanted
The server shall fail with a dedicated message which explain the situation and gives a hint about how to solve this

## Solution
- reproduce problem
- introduce validation inside `AbstractSecurityConfigruation` which throws dedicated error messages

### LINKED ISSUE #3915: SecHub documentation of general configuration oauth2 is not correct
## Situation

https://mercedes-benz.github.io/sechub/latest/sechub-techdoc.html#section-general-configuration

mentions:
```
sechub.security.oauth2.jwt.jwk-set-uri
sechub.security.oauth2.mode
sechub.security.oauth2.opaque-token.client-id
sechub.security.oauth2.opaque-token.client-secret
sechub.security.oauth2.opaque-token.introspection-uri
```

but this has changed /is outdated and is no longer correct!

Problem here: the outdated information is "handmade" and not generated (as our other setup information)

## Wanted
The outdated information shall be removed and be replaced by correct one.

## Solution
- remove handmade (wrong) parts
- we do not write any new handmade parts here but instead rely on the current example yamls + the output of #3916 later

### LINKED ISSUE #3917: InMemoryCache blocks application shutdown when application failure on start
## Situation
While #3914 appeared the SecHub server did not start, but the process was still running

### Analyse
The class InMemoryCache registers at `SpringApplicationShutdownHandler`.

But the `SpringApplicationShutdownHandler` only recognizes `ContextClosedEvent` but not `ApplicationFailedEvent`... which means the scheduled thread pool future will not stop at startup failues and the aplication process will be kept on running.

## Wanted
A failing application shall exit

## Solution
Change behavior

### LINKED ISSUE #3919: Oauth2 modes must support comma separated lists
## Situation
In our deployment of SecHub server 2.6.0 in combination with oauht2, the server failed to start - see #3914 

### Analyze
After analyzing the problem it became clear that you have cannot use comma separated lists directly as a set:

> For example, the configuration property my.service[0].other would use an environment variable named MY_SERVICE_0_OTHER.

For details look at:
https://docs.spring.io/spring-boot/reference/features/external-config.html#features.external-config.typesafe-configuration-properties.relaxed-binding.environment-variables

We assumed this would work like for `spring.active.profiles` which is a comma separated list. But look at the code, it appeared that this also not a normal `ConfigurationProperties` class bust solved different.

```yaml
sechub:
  security:
    server:
      modes:
        - oauth2
        - classic

```
would lead to the situation, that we would need to configure via environment variables

```
SECHUB_SECURITY_SERVER_MODES_0_=oauth2
SECHUB_SECURITY_SERVER_MODES_1_=classic
```
And this is not the way we want it.

## Wanted
We want to have the possibility to define lists very flexible via a comma separated list ( as we do already in other configuration parts of SecHub). The main reason is the inconvenience inside helm charts - it would not be possible to just change one value. And to provide multiple values which could be changed on deployments would have to inject dummies like:
```
SECHUB_SECURITY_SERVER_MODES_0_=oauth2
SECHUB_SECURITY_SERVER_MODES_1_=dummy0
SECHUB_SECURITY_SERVER_MODES_3_=dummy1

```

Following shall be possible:
`SECHUB_SECURITY_SERVER_MODES=oauth2,classic`
and also for
`SECHUB_SECURITY_LOGIN_MODES=oauth2,classic`

## Solution

- [x] Change the implemation of `SecHubSecurityProperties` to provide comma separated lists
- [x] Change the documented yaml examples
- [x] Write some notes about this inside coding convention and forbid usage of lists in yaml files for SecHub configuration parts.
- [x] Make tests work again

de-jcup:
This PR 

- [x] closes #3917
- [x] closes #3915
- [x] closes #3914
- [x] closes #3919

### DIFF
--- a/sechub-doc/src/docs/asciidoc/documents/shared/springboot_security.adoc
+++ b/sechub-doc/src/docs/asciidoc/documents/shared/springboot_security.adoc
@@ -28,7 +28,7 @@ Below is an example configuration with `Classic` mode and `OAuth2` in `JWT` mode
 sechub:
   security:
     server:
-      modes: oauth2, classic <1>
+      modes: oauth2,classic <1>
       oauth2:
         mode: jwt
         jwt:
@@ -85,7 +85,7 @@ Below is a full configuration example with `Login` enabled in `Classic` and `OAu
 sechub:
   security:
     server:
-      modes: oauth2, classic <1>
+      modes: oauth2,classic <1>
       oauth2:
         mode: jwt
         jwt:
@@ -97,7 +97,7 @@ sechub:
       login-page: /login
       # Here you can configure the target application where the user will be redirected after successful login
       redirect-uri: <your-target-application>
-      modes: oauth2, classic
+      modes: oauth2,classic
       oauth2:
         client-id: <client-id>
         client-secret: <client-secret>
--- a/sechub-doc/src/docs/asciidoc/documents/shared/springboot_security.adoc
+++ b/sechub-doc/src/docs/asciidoc/documents/shared/springboot_security.adoc
@@ -28,15 +28,15 @@ Below is an example configuration with `Classic` mode and `OAuth2` in `JWT` mode
 sechub:
   security:
     server:
-      modes:
-        - oauth2
-        - classic
+      modes: oauth2, classic <1>
       oauth2:
         mode: jwt
         jwt:
           jwk-set-uri: <jwk-set-uri-of-your-idp>
 ----
-
+<1> A list of comma separated resource server modes. Supported values are: `oauth2` and `classic`.
+    At least one mode must be defined.
+ 
 In `Opaque Token` mode the server forwards the token to an introspection endpoint for every request.
 
 NOTE:  `Opaque Token` mode increases the load on the IDP server hence caching is very important. Adjust the caching settings to your
@@ -49,8 +49,7 @@ Below is an example configuration with Classic mode and OAuth2 in `Opaque Token`
 sechub:
   security:
     server:
-      modes:
-        - oauth2
+      modes: oauth2
       oauth2:
         mode: opaque-token
         opaque-token:
@@ -86,9 +85,7 @@ Below is a full configuration example with `Login` enabled in `Classic` and `OAu
 sechub:
   security:
     server:
-      modes:
-        - oauth2
-        - classic
+      modes: oauth2, classic <1>
       oauth2:
         mode: jwt
         jwt:
@@ -100,9 +97,7 @@ sechub:
       login-page: /login
       # Here you can configure the target application where the user will be redirected after successful login
       redirect-uri: <your-target-application>
-      modes:
-        - oauth2
-        - classic
+      modes: oauth2, classic
       oauth2:
         client-id: <client-id>
         client-secret: <client-secret>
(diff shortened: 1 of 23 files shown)


sechub-doc/src/docs/asciidoc/documents/shared/springboot_security.adoc
```diff
@@ -28,15 +28,15 @@ Below is an example configuration with `Classic` mode and `OAuth2` in `JWT` mode
 sechub:
   security:
     server:
-      modes:
-        - oauth2
-        - classic
+      modes: oauth2, classic <1>
```
sven-dmlr, line 31: I would prefer if blanks after the ',' would be omitted
in the documentation and examples.
```suggestion
      modes: oauth2,classic <1>
```

sechub-doc/src/docs/asciidoc/documents/shared/springboot_security.adoc
```diff
@@ -86,9 +85,7 @@ Below is a full configuration example with `Login` enabled in `Classic` and `OAu
...
```
sven-dmlr, line 88:
```suggestion
      modes: oauth2,classic <1>
```

sechub-doc/src/docs/asciidoc/documents/shared/springboot_security.adoc
```diff
@@ -100,9 +97,7 @@ sechub:
       login-page: /login
       # Here you can configure the target application where the user will be redirected after successful login
       redirect-uri: <your-target-application>
-      modes:
-        - oauth2
-        - classic
+      modes: oauth2, classic
```
sven-dmlr, line 100:
```suggestion
      modes: oauth2,classic
```

sechub-doc/src/docs/asciidoc/documents/techdoc/02_security_in_development.adoc
```diff
@@ -75,6 +83,23 @@ sechub:
...
+- 3) **OAuth2 + Classic Mode**
+
+It is possible to use both modes at same time - for example:
+
+[source,yaml]
+----
+sechub:
+  security:
+    server:
+      modes: oauth2, classic
```
sven-dmlr, line 95:
```suggestion
      modes: oauth2,classic
```

sechub-server/src/test/resources/application-test.yml
```diff
@@ -20,9 +20,7 @@
 sechub:
   security:
     server:
-      modes:
-        - oauth2
-        - classic
+      modes:  oauth2, classic
```
sven-dmlr, line 23:
```suggestion
      modes:  oauth2,classic
```

sechub-shared-kernel/src/main/java/com/mercedesbenz/sechub/sharedkernel/shutdown/SpringApplicationShutdownHandler.java
```diff
@@ -33,14 +32,13 @@ public void register(ShutdownListener shutdownListener) {
...
-        log.info("Spring application context closed - informing shutdown listeners");
+    public void handleShutdown() {
+        log.info("Start handling shutdown - informing shutdown listeners");
         shutdownListeners.forEach(listener -> {
             try {
                 listener.onShutdown();
-            } catch (Exception e) {
-                log.error("Failed to notify shutdown listener {}", listener.getClass(), e);
+            } catch (RuntimeException e) {
+                log.error("Notified shutdown listener failed: {}", listener.getClass(), e);
```
sven-dmlr, line 41: unclear wording.
Maybe one of these:
"Shutdown of listener {} failed"
"Shutdown notification of listener {} failed"

sven-dmlr (approved):
Looks good to me :+1: 
Please check my comments

PR #3926, opened 2025-02-26
Improve auto generated documentation #3916

### LINKED ISSUE #3537: SecHub documentation for general config and scheduler definitions must be improved
## Situation
The documentation about SecHub config values is a little bit messy and attributes are often not easy to find.

### Wrong and duplicated scopes
In our SecHub documentation (techdoc, operations) we have
 
- general config
- scheduler definitions

where the SecHub server config values and their defaults are described.

By a "scope" in the documentation we can group some values together.

Unfortunately the scopes are not a mandatory attribute inside the java annotations, and when missing, the name of the scope is automatically generated and is often wrong.

Examples:
`scheduler`, `schedule`, `s` ... for same type of configurations

### No headlines for scopes
Inside the headline of asciidoc output, we have only 

- general config
- scheduler definitions

But there are some many settings, that it is very cumbersome to find them inside the document

### Scheduler definitions is a misleading name
"Scheduler definition" was chosen as a name because it describes configurations for the `@Schedule` annotation from Spring Boot. But this is misleading because in SecHub we talk about `Scheduler` when it comes to job processing...

### Scheduler definition has duplicated entries for auto cleanup
For every domain own scope is defined, so the same setup (they have the same key) appears multiple
times! This is confusing

## Wanted
The upper mentioned problems shall be fixed

### Wrong and duplicated scopes
- [x] Introduce constants for groups (scopes)
- [x] Change annotation for `@MustBeDocumented` and make scope mandatory without any default
- [x] Remove the scope auto value generation inside asciidoc generator
- [x] Set all scopes in parts annotated with `@MustBeDocumented` by the introduced constant

### No headlines for scopes
- [ ] Change asciidoc generator: Every table for a given scope shall have also a headline of level 4, means listed in TOC and easier to find

### Rename "Scheduler definitions"
- [ ] Rename to "Trigger setup"

### Duplicated "autocleanup" entries
- [ ] Use "auto cleanup" as scope for all domains -> the setting shall only be generated once

### LINKED ISSUE #3916: Support Spring ConfigurationProperties for auto generated documentation
## Situation
We had situations like #3915 were documentation was outdated or missing.

The situation only appeared, because the documentation for configuration properties is not auto generated like the documentation for spring values.

## Wanted
Spring configuration properties shall be auto documented as well

## Solution
- Enhance generators

### LINKED ISSUE #3927: AsciidocGeneratorManualTest shall trigger all generators, even on failures
## Situation
We have introduced `AsciidocGeneratorManualTest` which is only executed when system property `sechub.manual.test.by.developer` is set to `true`.

The test is intended for developers to start generation while developing on the different sub generators.

Unfortunately some generates rely on generated data which comes from integration tests and other locations and are not available after a `gradlew clean` which will block generation.

## Wanted
It shall be possible to start the generation even when some generators fail.
After all has been generated and at least one failure has happened, there shall be an exception giving an overview which generators did fail and why.

The exceptions shall always be logged with full stacktrace.

Inside documentation gradle task the generation shall fail directly - as before.

de-jcup:
This PR
- closes #3916
- closes #3537
- closes #3927

### DIFF
--- a/deprecated-sechub-scan-product-nessus/src/main/java/com/mercedesbenz/sechub/domain/scan/product/nessus/NessusConstants.java
+++ b/deprecated-sechub-scan-product-nessus/src/main/java/com/mercedesbenz/sechub/domain/scan/product/nessus/NessusConstants.java
@@ -1,0 +1,8 @@
+// SPDX-License-Identifier: MIT
+package com.mercedesbenz.sechub.domain.scan.product.nessus;
+
+public class NessusConstants {
+
+    public static final String SCOPE_NESSUS = "Nessus";
+
+}
--- a/deprecated-sechub-scan-product-nessus/src/main/java/com/mercedesbenz/sechub/domain/scan/product/nessus/NessusInstallSetupImpl.java
+++ b/deprecated-sechub-scan-product-nessus/src/main/java/com/mercedesbenz/sechub/domain/scan/product/nessus/NessusInstallSetupImpl.java
@@ -1,46 +1,48 @@
 // SPDX-License-Identifier: MIT
 package com.mercedesbenz.sechub.domain.scan.product.nessus;
 
+import static com.mercedesbenz.sechub.domain.scan.product.nessus.NessusConstants.*;
+
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 
 import com.mercedesbenz.sechub.adapter.AbstractAdapterConfigBuilder;
-import com.mercedesbenz.sechub.sharedkernel.MustBeDocumented;
+import com.mercedesbenz.sechub.commons.core.doc.MustBeDocumented;
 
 @Component
 @Deprecated // will be removed in future
 public class NessusInstallSetupImpl implements NessusInstallSetup {
 
     @Value("${sechub.adapter.nessus.intranet.baseurl:deprecated}")
-    @MustBeDocumented("Base url of nessus used for intranet scans")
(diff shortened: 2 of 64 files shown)


sven-dmlr (changes requested):
Huge work - well done! :+1: 
Please check my comments.
and:
The generated documentation contains examples about how to set environment variables. (section: 4.1. Mandatory configuration)
```
export SECHUB_SECURITY_SERVER_OAUTH2_OPAQUE-TOKEN_CLIENT-ID=value
export SECHUB_SECURITY_SERVER_OAUTH2_OPAQUE-TOKEN_CLIENT-SECRET=value
export SECHUB_SECURITY_SERVER_OAUTH2_OPAQUE-TOKEN_DEFAULT-TOKEN-EXPIRES-IN=value
export SECHUB_SECURITY_SERVER_OAUTH2_OPAQUE-TOKEN_INTROSPECTION-URI=value
export SECHUB_SECURITY_SERVER_OAUTH2_OPAQUE-TOKEN_MAX-CACHE-DURATION=value
```
The '-' in the variable names must be removed.
E.g.:
`SECHUB_SECURITY_SERVER_OAUTH2_OPAQUE-TOKEN_MAX-CACHE-DURATION` ->
`SECHUB_SECURITY_SERVER_OAUTH2_OPAQUETOKEN_MAXCACHEDURATION`

Also see Spring docs: https://docs.spring.io/spring-boot/reference/features/external-config.html#features.external-config.typesafe-configuration-properties.relaxed-binding.environment-variables

sven-dmlr (approved):
:+1:

PR #3939, opened 2025-02-27
Feature 3923 impl logout section for webscans

### LINKED ISSUE #3923: Introduce logout section for sechub webscan config to avoid logouts by the ajax crawler during scans
### Siutation
Currently we do not notice when the webscan with ZAP performs a logout during the ajax crawler scan.
The regular spider and the active scan can be handled by defaults or adding excluded URLs to the webscan configuration.

### Wanted
For the ajax crawler we want a specific section which specifies the logout html element.

### Solution
Introduce a logout section inside the sechub webscan config. The following example might change before the implementation:
```json
{
  "apiVersion": "1.0",
  "webScan": {
    "url": "https://example.org",
    "login": {
      "url": "https://example.org/login"
    },
    "logout": {
      "xpath": "//*[@id=\"loginButton\"]",
      "htmlElement": "button"
    }
  }
}
```

#### Important
- the `logout` section is optional, but when a `logout` section is defined `xpath` and  `htmlElement` MUST be present

### LINKED ISSUE #3924: Use the logout section of the sechub webscan config inside the ZAP wrapper
### ToDo
After #3923 is implemented use the newly introduced`logout` section inside the ZAP wrapper and configure the ajax crawler logout exclude for the scan.

winzj:
- closes #3923 
- closes #3924


lorriborri (approved):
Well done! Looks good to me! :star: 🦕
I just found a minor issue belonging to java doc, but it's approved.

You might have some merge conflicts now, because of the changes from  https://github.com/mercedes-benz/sechub/pull/3958 and I'm not sure if they require a re-review.

PR #3952, opened 2025-03-03
ZAP wrapper login script improvements

### LINKED ISSUE #3944: Provide ZAP wrapper application argument to disable headless mode
## Situation
Currently the wrapper application uses always the headless variants of the webdriver browsers.

For development / testing it is sometimes necessary to see whats going on (e.g. developing custom groovy login scripts).
To enable this we must currently change code temporary - which is cumbersome and potentially error prone.

## Wanted
As a developer I want to start the ZAP wrapper application in non-headless mode as well.
In non-headless mode, the browser used by ZAP shall be "non-headless" and also the selenium parts of custom login scripts shall be "non-headless" in this case.

## Solution
- introduce command line parameter `--headless` which default is `true`
- switch used browser (name) by Zap by this parameter
- switch webderiver setup for script by this parameter as well

### LINKED ISSUE #3945: Provide possibiility to test custom ZAP login scripts by a manual test
## Situation
With 
- #3710 

the ZAP wrapper application can use a groovy script to handle automated logins.

But when a script is being customized/maintained it is not easy to test it.

## Wanted
It shall be very easy to start a login from the IDE

## Solution
Introduce `ZapGoovyLoginScriptManualTest` inside the `sechub-wrapper-owasp-zap` submodule.

- The test class shall implement `ManualTest` interface from `sechub-testframework`
- The test shall use parameter `--headless=true` introduced in #3944
- load a script from a system property path (mandatory)

### LINKED ISSUE #3953: ZAP wrapper shall give all user variables to login script
## Situation
Currently only `username` and `password` are provided by ZAP wrapper application for custom groovy login scripts.

Also the names of the variables are fix and will be checked.

## Wanted
- It shall be possible to use custom variables without deployment
- The variables `username` and `password` shall no longer be fix or mandatory

### LINKED ISSUE #3960: ZAP wrapper application shall execute login script resilient
## Situation
Currently the ZAP wrapper application does try to login one time.
If the login fails, the scan will be stopped and a failure returned to the user.

But if the login mechanism itself has sometimes problems - e.g. javascript parts have some timing issues etc. - this would always lead to a failing scan.

Means: The user experience would lack here because the scan fails, but the problem could be on the login mask of the login provider.

## Wanted
If a login does not work /has failed, the login shall be retried.
The amount of retries shall be configurable by PDS parameter at runtime.

## Solution
- introduce PDS parameter `wrapper.loginscript.failure.resilience.retries`
- executor will be called given amount if login script fails
- each retry (and final "give up" will be logged

de-jcup:
This PR

- [x] closes #3945 
- [x] closes #3944 
- [x] closes #3953
- [x] closes #3960


winzj (approved):
Looking all good to me :rocket: 

Just found one constants value that might have been missed on refactoring.

PR #3963, opened 2025-03-10
Count amount of logins during scans

### LINKED ISSUE #3961: ZAP wrapper returns User-Info with amount of re-logins during scans
## Situation
The wrapper can check regularly if a session is authenticated or not, when a user has defined a login verification. If the session is not authenticated for some reason, the wrapper can perform a re-login.

## Wanted
The user should be informed about the amount of re-logins with Message.
The amount of re-logins should be logged as warnings, if more than one login was performed.

## Solution
Implement a counter to count the amount of re-logins.
LOG the amount of re-logins as warning if > 1
Use the ZapProductMessageHelper to return message to the user.

lorriborri:
- closes #3961

### DIFF
--- a/sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/zapwrapper/scan/ZapScanner.java
+++ b/sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/zapwrapper/scan/ZapScanner.java
@@ -106,8 +106,8 @@ public class ZapScanner implements ZapScan {
 
             /* After scan */
             generateZapReport();
-            informUserAboutAmountOfLogins();
             cleanUp();
+            informUserAboutAmountOfLogins();
         } catch (ClientApiException | ZapWrapperRuntimeException e) {
             informUserAboutAmountOfLogins();
             cleanUp();
--- a/sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/zapwrapper/scan/ZapScanner.java
+++ b/sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/zapwrapper/scan/ZapScanner.java
@@ -336,7 +336,7 @@ public class ZapScanner implements ZapScan {
                 runAndWaitActiveScan();
             }
         }
-        informUserAboutReLogins();
+        informUserAboutAmountOfLogins();
     }
 
     /**
@@ -746,7 +746,7 @@ public class ZapScanner implements ZapScan {
         LOG.info("For scan {}: Active scan completed.", scanContext.getContextName());
     }
 
-    private void informUserAboutReLogins() {
+    private void informUserAboutAmountOfLogins() {
         int count = loginCounter.getCount();
         if (count == 0) {
             return;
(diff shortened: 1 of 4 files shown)


sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/zapwrapper/scan/ZapScanner.java
```diff
@@ -738,6 +746,30 @@ void waitForActiveScanResults(int scanId) throws ClientApiException {
...
+            LOG.info("For scan: {}. The user was logged in once before the scan.", scanContext.getContextName());
+            message = new SecHubMessage(SecHubMessageType.INFO, "The user was logged in once before the scan.");
+        } else {
+            /* at least one re-login was performed */
+            count--; // the first login is not counted as re-login
+            LOG.warn("For scan: {}. The user was re-logged in {} times during the scan. This could be a sign of a misconfiguration.",
+                    scanContext.getContextName(), count);
+            message = new SecHubMessage(SecHubMessageType.WARNING,
+                    "The configured user was re-logged in " + count + " times during the scan. This could be a sign of a misconfiguration.");
+        }
```
winzj, line 767:
```suggestion
        if (count == 1) {
            /* no re-login was performed and user was logged in once */
            LOG.info("For scan: {}. The user was logged in once before the scan.", scanContext.getContextName());
            message = new SecHubMessage(SecHubMessageType.INFO, "The user was logged in once before the scan.");
        } else {
            /* at least one re-login was performed */
            LOG.warn("For scan: {}. The user was logged in {} times during the scan.",
                    scanContext.getContextName(), count);
            message = new SecHubMessage(SecHubMessageType.WARNING,
                    "The configured user was logged in " + loginCounter + " times during the scan. If the number of login attempts is large, consider configuring excludes and/or a logout section in the SecHub configuration file to avoid any logout.");
        }
```

sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/zapwrapper/scan/ZapScanner.java
```diff
@@ -368,6 +374,8 @@ UserInformation setupLoginInsideZapContext() throws ClientApiException {
             return null;
         }
 
+        loginCounter.increment();
```
winzj, line 377: Move the increment before the return that we count the successful login attempts only.

winzj (changes requested):
Looking good :croissant: 

I just have a small improvement, please see the comments as discussed.

sechub-wrapper-owasp-zap/src/main/java/com/mercedesbenz/sechub/zapwrapper/scan/ZapScanner.java
```diff
@@ -101,8 +106,10 @@ public void scan() throws ZapWrapperRuntimeException {
 
             /* After scan */
             generateZapReport();
+            informUserAboutAmountOfLogins();
             cleanUp();
```
winzj, line 110: I would change the order inside the try block so if the first cleanUp fails for some reason the informUserAboutAmountOfLogins() is only called once inside the catch block. The catch block is fine. This way we ensure that the message is only sent once.
```suggestion
            cleanUp();
            informUserAboutAmountOfLogins();
```

winzj (approved):
Looking good :rocket: 

I would just change the call order in the try block, please see my comment.

PR #3990, opened 2025-03-14
limit cache writes #3987

### LINKED ISSUE #3987: Tweak actions cache
## Situation
The cache for GH Actions is nearly filled up (10GB).
Almost everything is occupied from the `gradle-setup` cache entries.
On each run a new entry is generated.

## Wanted
Do not write cache entries on every run. Instead re-use existing cache entries.
See https://github.com/gradle/actions/blob/main/docs/setup-gradle.md#caching-build-state-between-jobs

## ToDo
Use `cache-read-only` parameter and write cache only on main (`master`) branch.
This should reduce the cache usage of `gradle-setup` significantly.

sven-dmlr:
...in order to keep the cache small
- closes #3987

### DIFF
--- a/.github/workflows/documentation-build.yml
+++ b/.github/workflows/documentation-build.yml
@@ -45,7 +45,10 @@ jobs:
       - name: Set up Gradle
         uses: gradle/actions/setup-gradle@94baf225fe0a508e581a564467443d0e2379123b
         with:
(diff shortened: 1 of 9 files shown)


haerter-tss (approved):
Looks good! 🍭

PR #4004, opened 2025-03-24
PDS wrapper prepare improvements

### LINKED ISSUE #4003: Improve prepare wrapper return code and validate azure tokens
## Situation
The prepare wrapper for docker can currently not validate azure API tokens and further returns always exit code 0 (not failed)

## Wanted
Azure tokens should also be validated and the return code should be 1 when something failed (it is necessary for monitoring)

## Solution
- adjust the validation regex pattern for docker credentials to allow base64 characters
- do a system exit 1, when the wrapper failed

### LINKED ISSUE #4005: Default image name for PDS Xray
## Situation
When a user uploads a docker .tar archive without a tag, the image name results in an empty name and the script crashes.

## Wanted
If the tag is empty, a default mage name should be added. A short checksum would be suitable e.g. MD5 (we can use it here, because its not a safety critical action). If the image is already uploaded, the script crashes.

## Solution
Adjust the xray.sh script and check if the image name is empty. If the name is empty create the MD5 checksum and give the default image name "checksum-<md5-checksum>:latest

lorriborri:
- closes #4003
- closes #4005


sven-dmlr (approved):
Looks good to me :+1:

PR #4008, opened 2025-03-24
Project owner is now able to change ownership itself to other user #4007

### LINKED ISSUE #4007: Project owner shall be able to transform ownership to other user
## Situation
Currently only an administrator is able to change project owner ship.

## Wanted
As a project owner I want to transform the owner ship to another user.
The transfer shall be done immediately without any approval.

## Solution
- [x] We rename the existing use case `UC_ADMIN_CHANGES_PROJECT_OWNER` to `UC_ADMIN_OR_OWNER_CHANGES_PROJECT_OWNER`
- [x] We provide a new endpoint `/api/management/project/{projectId}/owner/{userId}` wich calls the existing service, accessible by administrators and owners
- [x] the existing services will also become accessible by owners but inside the service there will be an assertion if change is allowed or not (only admins or owner of the probject)
- [x] The old api endpoint `/api/admin/project/{projectId}/owner/{userId}` will be kept for compatiblity reasons, but REST method is marked as deprecated (and no longer documented) - still accessible only for administrators

### LINKED ISSUE #4011: Frontend test setup script does not work for integration test variant
## Situation
The test setup script `test-setups/integrationtest-setups/setup.sh` does not work
because the relative paths are not correct in some lines. But inside the docker test setup it is correct.

Also there is much redundant code.

## Wanted
- This shall be fixed
- No redundant script code
- Maintaining or creating new test setups shall be easier
- It shall not matter from which location the scripts are executed: must always work

## Solution
- Introduce common-setup.sh and provide an init method and introduce common constants.
- identify the script source location and calculate all paths from this location

### LINKED ISSUE #4016: WebUI shall show user ids for owner and assigned users
## Situation
Currently the web ui shows for projects the email addresses for owner and assigned users .

## Wanted
The UI shall show the user ids. The email information shall be still available but not so prominent

## Solution
- change api json result of project detail data (server change)
  - introduce ProjectUserData and use it inside ProjectData as "owner" and "assignedUsers"
  - ProjectUserData will provide additional "userId" and "emailAddress"
  - ProjectUserData will be comparable (sortable) and has to implement hashcode and equals
 
  
- change display to user id
  - show "ownerUserId" instead "ownerEmailAddress". 
  - provide a tooltip over the name which contains the mail address of the user

### LINKED ISSUE #4021: Introduce WebUI project settings dialog with possiblity to change project owner
## Wanted
The Web UI shall provide a project settings dialog and it shall be possible to 
change the project owner of a project.

### Todos

- [x] Project details shall be updated, new owner must appear after success
- [x] Settings dialog shall be used later for other parts as well, so we need a close button but also a button to change the owner
- [x] Changing a owner inside the settings dialog but not changing it on server side, may not update the project details

### Out of scope in this issue:
The settings button and the functions there are not restricted by UI, but only at Server side at the moment. This will be done later with https://github.com/mercedes-benz/sechub/issues/4022

## Solution
We provide a modal dialog which can be opened via a button at the project page.

de-jcup:
This PR

- [x] closes #4007
- [x] closes #4011
- [x] closes #4016
- [x] closes #4021


lorriborri (changes requested):
Good Work and well done!
I like the solution you chose for the server implementation  :star: 🦕
I found some issues on the front-end part, please have a look at my suggestions.

sechub-web-ui/src/components/ProjectJobList.vue
```diff
@@ -277,6 +317,56 @@
...
+      function openSettingsDialog () {
+        // setup settings variables to current state of project
+        settingsOwnerFieldIcon.value = 'mdi-check'
+        settingsOwnerUserId.value = projectData.value.owner.userId
+
+        // show dialog
+        settingsDialog.value = true
+      }
+
+      async function onSettingsOwnerFieldChanged () {
```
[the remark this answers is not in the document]
de-jcup: I introduced a new event "onProjectOwnerChanged". I was able to to the operation there and to update the projectData instance from store. But... unfortunately we have not all information here: The email address of the new owner is missing.

Because of this I was forced to reload the projects cache here and update afterwards the projectData.value.owner parts

sechub-web-ui/src/components/ProjectJobList.vue
```diff
@@ -277,6 +292,26 @@
         console.error(errMsg, err)
       }
 
+      function openSettingsDialog () {
+        settingsDialog.value = true
+      }
+
+      async function onProjectOwnerChanged (newOwnerUserId: String) {
```
lorriborri, line 299: Instead of rewriting this method, I would prefer a composable which does the fetch and the store job:
see: https://vuejs.org/guide/reusability/composables 

/comoposables/useProjects.ts
```ts
import { ProjectData } from '@/generated-sources/openapi'
import defaultClient from '@/services/defaultClient'
import { useProjectStore } from '@/stores/projectStore'


export function useFetchProjects(){
    const store = useProjectStore()
    const projects = ref<ProjectData[]>([])
    const error = ref<string | undefined>(undefined)
    const loading = ref(true)

    const fetchProjects = async () => {
    try {
        projects.value = await defaultClient.withProjectApi.getAssignedProjectDataList()
        store.storeProjects(projects.value)
      } catch (err) {
        error.value = 'ProjectAPI error fetching assigned projects.'
        console.error('ProjectAPI error fetching assigned projects:', err)
      } finally {
        loading.value = false
      }
    };

    fetchProjects();
    return { projects, error, loading }
}
```

so you could use it in ProjectList.vue
```ts
    setup () {
      const router = useRouter()

      const { projects, error, loading } = useFetchProjects()

...
```

and in your code
```ts
      async function onProjectOwnerChanged (newOwnerUserId: String) {
        console.log('Project owner for project', projectId.value, 'changed to', newOwnerUserId)

        const { projects, error, loading } = useFetchProjects()

        const newLoadedProject = store.getProjectById(projectId.value)
        if (newLoadedProject !== undefined) {
          projectData.value.owner.userId = newLoadedProject.owner.userId
          projectData.value.owner.emailAddress = newLoadedProject.owner.emailAddress
        }
      }
```

sechub-web-ui/src/components/ProjectJobList.vue
```diff
@@ -277,6 +292,26 @@
         console.error(errMsg, err)
       }
 
+      function openSettingsDialog () {
```
lorriborri, line 295: optional: this function can be replaced by using :)
`            <v-btn v-tooltip="$t('PROJECT_DETAILS_TOOLTIP_SETTINGS')" icon="mdi-pencil" @click="settingsDialog=true" />
`

or better: keep it and also use it when the change was successful

lorriborri (changes requested):
Good work! I like the improvements! 
Unfortunately I found some issues, please have a look at my comments.

lorriborri (approved):
Looks good to me! Very good work! :star:  🦘

PR #4019, opened 2025-03-27
Moved static file location to /static/login

### LINKED ISSUE #4018: Adjust web-ui login static file directory for proxy pass
## Situation
Static files of web-ui login page (delivered by server) are placed under static/ and are therefore not forwarded.

## Wanted
Static files should also be placed under /login

## Solution 
Move files from /static/login
Adjust the AbstractSecurity class and add /login to paths

lorriborri:
- closes #4018


de-jcup (approved):
Good job :+1:

PR #4024, opened 2025-03-31
Add error handling to retrieveBrowserStorage #4023

### LINKED ISSUE #4023: Handle error on storage access correctly
### Situation
Currently the script authentication mechanism uses some JavaScript to check the `sessionStorage` and the `localStorage` for any session related data, like tokens.
If an error happens during the execution of the JavaScript snippet, the execution of the scan fails because the error is not handled correctly.

### Wanted
Handle and log the JavaScript error. The scan job shall **NOT** fail if any browser storage is not accessible.

### Solution
The JavaScript snippet handles the error and passes it to the calling method.

### LINKED ISSUE #4038: Perform authentication before interactions with webscan target
### Situation
Currently it the authetnication steps for webscan can be performed after the first interactions with the scan target, in certain scenarios.

### Solution
Perform the authentication earlier to ensure an authenticated session when interacting with the target URL.

winzj:
- closes #4023 
- closes #4038 4038


de-jcup (approved):
Good work - looks good to me :rocket:

PR #4026, opened 2025-03-31
switch added for enabling internal access #4025

### LINKED ISSUE #4025: Server Helm chart: Add possibility to grant access from internal
## Situation
Currently access to "sechub-server-internal" is restricted to "web-ui" only.
Other Pods from the same k8s namespace currently cannot access.

## Wanted
An switch in the values.yaml so access can be granted.

## ToDo
Add `{{ .Values.security.allowAccessFromSameNamespace }}` which defaults to `false`.

sven-dmlr:
- closes #4025


haerter-tss (approved):
Looks good! 🪭

PR #4040, opened 2025-04-03
Introduce IaC scan type #3287

### LINKED ISSUE #3287: Introduce own scan type for IAC and use it
## Situation
Currently we do `IAC` scanning (e.g. with KICS) via source code scan tpye in configuration.

But we have also the `IAC` findings inside source code scan type in reports
This is often confusing for end users when reading the reports.

Also it becomes difficult for configuration of executors - because currently source scan products which are doing no IAC scanning will also be executed - even when only a IAC scan is wanted.

## Wanted
- In reports and inside configuration it shall be possible to define `IAC` explicit.
- Only for `IAC` scans IAC products shall be called
- It shall NOT be possible to do an IAC scan with binaries!

Out of scope:
- The new scan type will not be supported by GitHub action. For this we have a follow up issue: #4055

### Tasks

- [x] introduce `IaC` scan type
- [x] handle new scan type in system tests
- [x] create `PDS product executor for new type
- [x] change IAC PDS solution pds config file to provide new scan type
- [x] adopt html report build (if necessary)
- [x] write a PDS integration test which returns a Kics result and ensure it is listed as new scan type inside the JSON report and also the HTML report
- [x] ensure IAC can be scanned together with secretScan, codeScan and licenseScan (we have a grouping inside scheduler strategy!)
- [x] Update documentation, describe IAC usage etc.
- [x] enhance SecHubConfigurationFileValidation to deny

### LINKED ISSUE #4052: Sort scan type examples alphabetically inside documentation
## Situation
https://mercedes-benz.github.io/sechub/latest/sechub-client.html
Contains examples for different scan types.

With #4040 we have to add the new scan type as well. The question was: where do we add the the new documentation.

## Wanted
It shall be clear where to find

## Solution
We simply sort the examples alphabetical:

- Code scan
- IaC scan
- Infrastructure scan  
- License scan
- Secret scan
- Web scan

de-jcup:
This PR

- [x] closes #3287
- [x] closes #4052


winzj (approved):
Looking good to me :rocket: 

Just found some minor things like typos. They are optional changes from my side.

PR #4044, opened 2025-04-04
server access only with role sechub-frontend #4043

### LINKED ISSUE #4043: Helm charts: Server access with sechub-frontend role only
## Situation
Recently there was a Helm value defined allowing access from the whole namespace.

## ToDo
This should be changed to a label "role=sechub-frontend" so access can be explicitly granted.

sven-dmlr:
- closes #4043


haerter-tss (approved):
Looks good! 💎

PR #4056, opened 2025-04-08
Add profile IDs for each project to the project data #4053

### LINKED ISSUE #4053: Any user shall be able to see the profiles assigned to the projects the user is part of
### Situation
Currently a normal user cannot see which profiles are assigned to his projects.

### Wanted
A user shall be able to get information about the profiles of each project.

### Solution
With `/api/projects` a user can get information about all projects the user is part of. This REST API shall be extended to return all profile IDs for each project alongside the other project data.

winzj:
- closes #4053

### DIFF
--- a/sechub-administration/src/test/java/com/mercedesbenz/sechub/domain/administration/project/ProjectServiceTest.java
+++ b/sechub-administration/src/test/java/com/mercedesbenz/sechub/domain/administration/project/ProjectServiceTest.java
@@ -3,7 +3,6 @@ package com.mercedesbenz.sechub.domain.administration.project;
 
 import static com.mercedesbenz.sechub.domain.administration.user.TestUserCreationFactory.createProjectUser;
 import static com.mercedesbenz.sechub.sharedkernel.messaging.MessageDataKeys.PROJECT_ASSIGNED_PROFILE_IDS;
-import static com.mercedesbenz.sechub.sharedkernel.messaging.MessageDataKeys.PROJECT_IDS;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.mock;
@@ -59,7 +58,6 @@ class ProjectServiceTest {
             String expectedOwnerEmailAddress) {
         /* prepare */
         DomainMessageSynchronousResult mockedResponse = mock();
-        when(mockedResponse.get(PROJECT_IDS)).thenReturn(List.of("mocked-project"));
         when(mockedResponse.get(PROJECT_ASSIGNED_PROFILE_IDS)).thenReturn(Collections.emptyMap());
         when(eventBus.sendSynchron(any())).thenReturn(mockedResponse);
 
@@ -104,11 +102,14 @@ class ProjectServiceTest {
         assertThat(projects.get(1).getAssignedUsers().size()).isGreaterThan(0);
 
         // verify profile IDs are included correctly
+        boolean assertionDone = false;
         for (ProjectData projectData : projects) {
             if (PROJECT_ID_2.equals(projectData.getProjectId())) {
                 assertThat(projectData.getAssignedProfileIds()).containsAll(profileIdsProject2);
+                assertionDone = true;
             }
         }
+        assertThat(assertionDone).isTrue();
     }
 
     @Test
@@ -116,7 +117,6 @@ class ProjectServiceTest {
         /* prepare */
         String userId = USER_ID_3;
         DomainMessageSynchronousResult mockedResponse = mock();
-        when(mockedResponse.get(PROJECT_IDS)).thenReturn(List.of("mocked-project"));
         when(mockedResponse.get(PROJECT_ASSIGNED_PROFILE_IDS)).thenReturn(Collections.emptyMap());
         when(eventBus.sendSynchron(any())).thenReturn(mockedResponse);
 
@@ -133,7 +133,6 @@ class ProjectServiceTest {
         /* prepare */
         String userId = USER_ID_4;
         DomainMessageSynchronousResult mockedResponse = mock();
-        when(mockedResponse.get(PROJECT_IDS)).thenReturn(List.of("mocked-project"));
         when(mockedResponse.get(PROJECT_ASSIGNED_PROFILE_IDS)).thenReturn(Collections.emptyMap());
         when(eventBus.sendSynchron(any())).thenReturn(mockedResponse);
 
@@ -214,7 +213,6 @@ class ProjectServiceTest {
     private DomainMessageSynchronousResult prepareValidSyncResult(String projectId, List<String> profileIds) {
         DomainMessageSynchronousResult response = new DomainMessageSynchronousResult(MessageID.REQUEST_PROFILE_IDS_FOR_PROJECT);
         response.set(PROJECT_ASSIGNED_PROFILE_IDS, Map.of(projectId, profileIds));
-        response.set(PROJECT_IDS, null);
         return response;
     }
 }
--- a/sechub-administration/src/test/java/com/mercedesbenz/sechub/domain/administration/project/ProjectServiceTest.java
+++ b/sechub-administration/src/test/java/com/mercedesbenz/sechub/domain/administration/project/ProjectServiceTest.java
@@ -2,10 +2,16 @@
 package com.mercedesbenz.sechub.domain.administration.project;
 
 import static com.mercedesbenz.sechub.domain.administration.user.TestUserCreationFactory.createProjectUser;
+import static com.mercedesbenz.sechub.sharedkernel.messaging.MessageDataKeys.PROJECT_ASSIGNED_PROFILE_IDS;
+import static com.mercedesbenz.sechub.sharedkernel.messaging.MessageDataKeys.PROJECT_IDS;
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.mockito.Mockito.*;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
 
+import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.stream.Stream;
 
@@ -20,6 +26,9 @@ import org.mockito.Mockito;
 
 import com.mercedesbenz.sechub.domain.administration.user.User;
 import com.mercedesbenz.sechub.domain.administration.user.UserRepository;
+import com.mercedesbenz.sechub.sharedkernel.messaging.DomainMessageService;
+import com.mercedesbenz.sechub.sharedkernel.messaging.DomainMessageSynchronousResult;
+import com.mercedesbenz.sechub.sharedkernel.messaging.MessageID;
 import com.mercedesbenz.sechub.sharedkernel.validation.UserInputAssertion;
 
 class ProjectServiceTest {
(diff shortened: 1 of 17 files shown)


sechub-administration/src/test/java/com/mercedesbenz/sechub/domain/administration/project/ProjectServiceTest.java
```diff
@@ -82,12 +102,23 @@ void user2_sees_assigned_and_owned_projects_with_users() {
 
         assertThat(projects.get(0).getAssignedUsers().size()).isGreaterThan(0);
         assertThat(projects.get(1).getAssignedUsers().size()).isGreaterThan(0);
+
+        // verify profile IDs are included correctly
+        for (ProjectData projectData : projects) {
+            if (PROJECT_ID_2.equals(projectData.getProjectId())) {
```
de-jcup, line 108: you should add a sanity check here - if the PROJECT_ID_2 would be never inside the results the assertion would never happen...

I would do it the following way:
```java
 boolean assertionDone = false;
 for (ProjectData projectData : projects) {
            if (PROJECT_ID_2.equals(projectData.getProjectId())) {
assertThat(projectData.getAssignedProfileIds()).containsAll(profileIdsProject2);
assertionDone=true;
          }
}
assertThat(assertionDone).isTrue();

File: sechub-administration/src/test/java/com/mercedesbenz/sechub/domain/administration/project/ProjectServiceTest.java [hunk 2]
```diff
@@ -176,4 +211,10 @@ public Stream<? extends Arguments> provideArguments(ExtensionContext context) {
         /* @formatter:on */
     }
 
+    private DomainMessageSynchronousResult prepareValidSyncResult(String projectId, List<String> profileIds) {
+        DomainMessageSynchronousResult response = new DomainMessageSynchronousResult(MessageID.REQUEST_PROFILE_IDS_FOR_PROJECT);
+        response.set(PROJECT_ASSIGNED_PROFILE_IDS, Map.of(projectId, profileIds));
+        response.set(PROJECT_IDS, null);
```
de-jcup, line 217: why do you set this explicit to null? If not set, the field will be null anyway. This line is IMO useless.

sechub-integrationtest/src/test/java/com/mercedesbenz/sechub/integrationtest/scenario3/ProjectAdministrationScenario3IntTest.java
```diff
@@ -91,6 +97,22 @@ void change_project_ownership_by_admin_and_owners() {
...
+        Set<String> userAssignedProfileIds = projectDetailsOfnormalUser.get(0).getAssignedProfileIds();
+        assertThat(userAssignedProfileIds).containsExactly(PROFILE_1.id);
+
+        // owner can view profiles of assigned projects
+        List<ProjectData> projectDetailsOfOwner = as(USER_1).getAssignedProjectDataList();
+        assertThat(projectDetailsOfOwner).hasSize(1);
+        Set<String> ownerAssignedProfileIds = projectDetailsOfOwner.get(0).getAssignedProfileIds();
+        assertThat(ownerAssignedProfileIds).containsExactly(PROFILE_1.id);
+
+        // admin can view profiles of assigned projects
```
de-jcup, line 114: first of all the comment is not correct:
It would be more:
```suggestion
        // admin has no assigned projects
```

But... this is problematic/critical:
All users of the scenario are specific for test and will automatically reset - except the SUPER_ADMIN: this user is always existing, which means: If there is ANY other test executed before, which would assign the super admin to a project, this test wold fail!

If your assumption is here, that an admin is not automatically assigned to projects, then you would need to create here another user and set him to an administrator.
Or you could check here, that the SUPER_ADMIN does not contain PROFILE_1.id

=> for next loop (even when its empty, this works) and check that this does not happen.

Tests are not executed always in same order by gradle or inside IDEs - so I would prefer if we change this to avoid to become this a flaky test in future.

sechub-scan/src/main/java/com/mercedesbenz/sechub/domain/scan/product/config/ProfileMessageHandler.java
```diff
@@ -0,0 +1,50 @@
...
+            List<ProductExecutionProfile> executionProfilesForProject = productExecutionProfileRepository.findExecutionProfilesForProject(id);
+            /* @formatter:off */
+            List<String> profileIds = executionProfilesForProject.stream()
+                                                                .map(ProductExecutionProfile::getId)
+                                                                .collect(Collectors.toList());
+            /* @formatter:on */
+            projectToProfileIds.put(id, profileIds);
+        }
+        response.set(PROJECT_ASSIGNED_PROFILE_IDS, projectToProfileIds);
+        response.set(PROJECT_IDS, null);
```
de-jcup, line 46: why do you set this explicit to null? If not set, the field will be null anyway. This line is IMO useless.

sechub-scan/src/main/java/com/mercedesbenz/sechub/domain/scan/product/config/ProfileMessageHandler.java
```diff
@@ -0,0 +1,50 @@
...
+public class ProfileMessageHandler implements SynchronMessageHandler {
+
+    private final ProductExecutionProfileRepository productExecutionProfileRepository;
+
+    public ProfileMessageHandler(ProductExecutionProfileRepository productExecutionProfileRepository) {
+        this.productExecutionProfileRepository = productExecutionProfileRepository;
+    }
+
+    @IsRecevingSyncMessage(MessageID.REQUEST_PROFILE_IDS_FOR_PROJECT)
+    @Override
```
de-jcup, line 30: Please take a look into `ScanMessageHandler.java`:
I designed the synchronous messaging in a way, 
that we have always two different message ids: one for the question, another one for the answer.

It is interesting that the mechanism works with same message id inside the result.
But when it comes to generartion of diagrams for events, I think this could become problematic.

I would suggest to introduce an additional message id `RESULT_PROFILE_IDS_FOR_PROJECT`
You can of course use the same data type inside the new message Id.

Also you have to markup that your are answering here:
```suggestion
@IsSendingSyncMessageAnswer(value = MessageID.RESULT_PROFILE_IDS_FOR_PROJECT, answeringTo = MessageID.REQUEST_PROFILE_IDS_FOR_PROJECT)
    @Override
```

sechub-scan/src/test/java/com/mercedesbenz/sechub/domain/scan/product/config/ProfileMessageHandlerTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+        /* execute */
+        DomainMessageSynchronousResult response = handlerToTest.receiveSynchronMessage(message);
+
+        /* test */
+        Map<String, List<String>> projectToProfiles = response.get(PROJECT_ASSIGNED_PROFILE_IDS);
+        assertThat(projectToProfiles).containsKey(TEST_PROJECT_ID);
+        assertThat(projectToProfiles.get(TEST_PROJECT_ID)).isEmpty();
+    }
+
```
de-jcup, line 51: With then new answer message id you will need to change/adopt these tests.

sechub-shared-kernel/src/main/java/com/mercedesbenz/sechub/sharedkernel/messaging/MessageID.java
```diff
@@ -258,6 +258,12 @@ public enum MessageID {
      */
     ASSIGN_OWNER_AS_USER_TO_PROJECT(MessageDataKeys.PROJECT_TO_USER_DATA),
 
+    /**
+     * Request all execution profile IDs for each project. This is sent when a user
+     * requests information about all projects the user is part of.
+     */
+    REQUEST_PROFILE_IDS_FOR_PROJECT(MessageDataKeys.PROJECT_ASSIGNED_PROFILE_IDS, MessageDataKeys.PROJECT_IDS),
+
     ;
```
de-jcup, line 267: As mentioned before we need an answer mesage id here 
RESULT_PROFILE_IDS_FOR_PROJECT(...)

sechub-shared-kernel/src/test/java/com/mercedesbenz/sechub/sharedkernel/messaging/ListOfStringsMessageDataProviderTest.java
```diff
@@ -0,0 +1,51 @@
...
+        List<String> profileIds = providerToTest.get(null);
+        String profileIdsAsString = providerToTest.getString(null);
+
+        /* test */
+        assertThat(profileIds).isNull();
+        assertThat(profileIdsAsString).isNull();
+    }
+
+    @Test
+    void empty_values_handled_correctly() {
```
de-jcup, line 27: You check for null and you check for "[]"
what happens for providerToTest.get("") ?

sechub-shared-kernel/src/test/java/com/mercedesbenz/sechub/sharedkernel/messaging/ListOfStringsMessageDataProviderTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.util.Collections;
+import java.util.List;
+
+import org.junit.jupiter.api.Test;
+
+class ListOfStringsMessageDataProviderTest {
+
+    private ListOfStringsMessageDataProvider providerToTest = new ListOfStringsMessageDataProvider();
+
```
de-jcup, line 14: The tests are bit strange to me:

/* execute */ block does contain always 2 calls - which one is the one you test? For me it looks like these are always two tests combined.

And I would have written the tests with one or two ArguementsProvider and made the tests parameterized: provider would just have input and expected output - and maybe an identifier as first parameter, so better to read in outputs.

LocalDateTimeMessageDataProviderTest

sechub-shared-kernel/src/test/java/com/mercedesbenz/sechub/sharedkernel/messaging/MapStringToListOfStringsMessageDataProviderTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+import org.junit.jupiter.api.Test;
+
+class MapStringToListOfStringsMessageDataProviderTest {
+
+    private MapStringToListOfStringsMessageDataProvider providerToTest = new MapStringToListOfStringsMessageDataProvider();
```
de-jcup, line 14: Same as for ListOfStringsMessageDataproviderTest:

please divide tests - we want only ONE executed part per test, so it is clear WHAT is tested

Also `""` is not handled inside the tests as well

de-jcup (changes requested):
Good work, but I found some parts to change.

sechub-scan/src/test/java/com/mercedesbenz/sechub/domain/scan/product/config/ProfileMessageHandlerTest.java
```diff
@@ -0,0 +1,85 @@
...
+        DomainMessage message = prepareValidDomainMessage();
+        List<ProductExecutionProfile> executionProfilesOfProject = createProductExecutionProfilesbyIds(expectedProfileIds);
+        when(productExecutionProfileRepository.findExecutionProfilesForProject(TEST_PROJECT_ID)).thenReturn(executionProfilesOfProject);
+
+        /* execute */
+        DomainMessageSynchronousResult response = handlerToTest.receiveSynchronMessage(message);
+
+        /* test */
+        Map<String, List<String>> projectToProfiles = response.get(PROJECT_ASSIGNED_PROFILE_IDS);
+        assertThat(projectToProfiles).containsKey(TEST_PROJECT_ID);
```
de-jcup, line 65: You could have checked here that the response message id is correct. But this is not so important and can be left. Just for info.

de-jcup (approved):
Looks good to me :+1:  
I just found one small thing, but it's up to you if you want to change it or not - please read my comment.

PR #4064, opened 2025-04-09
Feature 4020 helm customize template

### LINKED ISSUE #4020: Adjust web-ui helm chart for customizable templates
# Situation
In #4006 it was made possible to use custom themes. Files are are put in `/static` (see #4019) will override the default template.

# Wanted
The helm chart needs to be adjusted accordingly.

haerter-tss:
It is now possible to use a custom css and background image. They can be defined as configmaps.
- closes #4020 

I also fixed the indentation inside the sechub-server deployment.yaml


sechub-solution/helm/sechub-server/templates/deployment.yaml
```diff
@@ -152,6 +172,9 @@ spec:
         # Base url for sechub server - e.g. https://sechub.example.com
         - name: SECHUB_SERVER_BASEURL
           value: "{{ .Values.server.baseURL.userAccess }}"
+        # Static resources for login page
+        - name: SPRING_WEB_RESOURCES_STATICLOCATIONS
+          value: "{{ .Values.staticResources.spring.staticLocations }}"
```
sven-dmlr, line 177: The pathes are fixed so the SPRING_WEB_RESOURCES_STATICLOCATIONS can be set dependent on the value of `staticResources.enabled `
```suggestion
{{- if .Values.staticResources.enabled }}
          value: "file:/sechub/web-resources/static,classpath:/static"
{{- else }}
          value: "classpath:/static"
{{- end }}
```

sven-dmlr (approved):
Code looks good for me :+1: 
Please check my suggestions about helping the users to create their custom configmaps.

PR #4069, opened 2025-04-10
Introduced project profiles in web ui #4067

### LINKED ISSUE #4067: WebUI shall show profile ids for project
## Situation
With #4053 users can now check the profiles of their projects by REST API.
But it is not shown at Web UI

## Wanted
At the Web UI the user shall see the profiles for the project

### LINKED ISSUE #4068: User project data shall contain only enabled project profiles
## Situation
With #4053 we introduced information about assigned projects inside project data for user
But this returns also inactive profiles, means information which is not relevant or users (or just confuseses them). 

With #4067 this becomes relevant.

## Wanted
The user shall have only information about active profiles for the project

## Solution
- [x] We rename the field inside project data from "assignedProfiles" to "enabledProfiles"
- [x] We change the logic to return only enabled profiles at the user rest call (no matter which role)
- [x] adopt open api defintion

de-jcup:
This PR

- [ ] closes #4067
- [ ] closes #4068

### DIFF
--- a/sechub-web-ui/src/components/JobReportOverview.vue
+++ b/sechub-web-ui/src/components/JobReportOverview.vue
@@ -107,20 +107,18 @@
       fetchReport()
 
       async function fetchReport () {
-
         // load cached report from store
         const reportFromStore = store.getReportByUUID(jobUUID.value)
         if (reportFromStore) {
           report.value = reportFromStore
-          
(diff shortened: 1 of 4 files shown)


lorriborri (approved):
Looks good to me, well done! :star: 🐊

PR #4077, opened 2025-04-10
update Helm chart version to latest #4047

sven-dmlr:


sven-dmlr:
Closing PR because 2.10.0 comes with PR #4064

PR #4080, opened 2025-04-11
Feature 4079 reduce massive log entries

### LINKED ISSUE #2968: Set default for verbose heartbeat logging of PDS to false
## Situation
`pds.config.heartbeat.verbose.logging.enabled` default is currently true.
This leads to much (unnecessary) console output inside IDEs.

## Wanted
Per default the verbose loggings should be disabled

## Solution
Change constant inside `PDSHeartBeatTriggerService`

### LINKED ISSUE #4079: Overwhelming debug logging
## Situation
Inside K8s environments where the debug logging is enabled, the log is flooded with entries like:

```
{"@timestamp":"2025-04-10T15:17:18.786950091Z","@version":"1","message":"No cookies found in the request","logger_name":"com.mercedesbenz.sechub.spring.security.DynamicBearerTokenResolver","thr 
```
Reason: The is alive check in K8s does checks without cookies - which is correct, but triggers the logging here. This will also happen (IMO) when there are basic auth based REST calls together with oauth2 enabled.

## Wanted
This is just too much and shall not appear even on debug level. 

## Solution 
Change to log level trace

de-jcup:
This PR

- closes #4079 
- closes #2968

### DIFF
--- a/sechub-commons-security-spring/src/main/java/com/mercedesbenz/sechub/spring/security/DynamicBearerTokenResolver.java
+++ b/sechub-commons-security-spring/src/main/java/com/mercedesbenz/sechub/spring/security/DynamicBearerTokenResolver.java
@@ -58,7 +58,7 @@ class DynamicBearerTokenResolver implements BearerTokenResolver {
         Cookie[] cookies = request.getCookies();
 
         if (cookies == null) {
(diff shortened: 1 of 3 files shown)


sven-dmlr (approved):
Looks good :+1:

PR #4091, opened 2025-04-11
Client: verify config for forbidden names #3897

### LINKED ISSUE #3897: Go client shall handle __archive_root__
## Situation
When #3776 has been implemented, it is possible to use  reserved ids `__binaries_archive_root__` and `__sourcecode_archive_root__` as identifiers to root level for the zip or tar archive without having a data section defined inside the configuration.

The SecHub go client does currently not allow any usage of a reference id which is not listed inside the data section. If somebody would use the reserved ids without data section, the client would fail. But if used with data section ,the SecHub server will fail with a validation error (it's forbidden by server to hold a data section with this reserved ids)

## Wanted
The SecHub client shall behave similar to web ui when uploaded zip file contains a root folder - means
it shall be possible to reference accept `__archive_root__` when  in this case.

## Solution

SecHub client will 

- fail when a data section contains `__binaries_archive_root__` or `__sourcecode_archive_root__` definition (will be asserted on
  Server side as well, but we want to fail fast at client side)

- the go client does always create a data section so it does not make sense
  to give here access to `__binaries_archive_root__` or `__sourcecode_archive_root__` 
  so there should be no further special handling

sven-dmlr:
- closes #3897

### DIFF
--- a/sechub-cli/src/mercedes-benz.com/sechub/cli/constants.go
+++ b/sechub-cli/src/mercedes-benz.com/sechub/cli/constants.go
@@ -189,6 +189,12 @@ var actionlist = []string{
 // archiveDataPrefix - Prefix in Zip or Tar archives for files from "data" section
 const archiveDataPrefix = "__data__"
 
+// forbiddenArchiveDataSectionNames - reserved names that are rejected by SecHub server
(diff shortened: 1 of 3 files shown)


haerter-tss (approved):
Looks good! 🕹️

PR #4095, opened 2025-04-14
Changed ordering of expected and values #4094

### LINKED ISSUE #4094: AssertProject mix up expected and value in outputs
## Situation
AssertProject failures message are misleading, because the error message parameters are swapped...

junit has always order: expected, value, but here value,expected was used...

## Solution
Fix ordering

de-jcup:
This PR 
- closes #4094


sven-dmlr (approved):
:+1:

PR #4097, opened 2025-04-14
Pass all environment variables to the sechub client call #4096

### LINKED ISSUE #4096: Pass all env variables to the sechub client calls inside the github actions
### Situation
Currently only specific ENV variables are passed to the sechub client call by the Github scan action.
Because of this custom variables might not be available for the sechub clients templating mechanism. 

### Wanted
Ensure all ENV variables are passed to the sechub client call.

winzj:
- closes #4096


sven-dmlr (approved):
:+1:

PR #4102, opened 2025-04-14
Expired opaque tokens do now lead to 401 which removes cookie #3892

### LINKED ISSUE #3892: SecHub Server send redirect to login if cookie expired
## Situation

Currently there is no way to react to expired auth cookies

## Wanted

SecHub Server should delete expired cookies and redirect to the login page.

## Solution

If the Server detects an expired cookie, it should delete the session cookie and throw a 401.
The web-ui nginx server will then automatically redirect to the login page.

de-jcup:
This PR
- closes #3892


winzj (approved):
Looking good :rocket:

PR #4105, opened 2025-04-16
Feature 4103 provide build as go client version to use

### LINKED ISSUE #3993: gh-action: always print output of client
## Situation
* Per default, the SecHub client's output is hidden - so it's difficult to analyze in case of problems.
   (output can only be seen when debugging the whole workflow at "re-run workflow")
* When setting the "debug" flag in the SecHub GH-Action, no difference is seen compared to not setting the flag.

## Wanted
* Always show the SecHub client's output

Atttention: Please ensure that the environment var `SECHUB_DEBUG=true` behavior for runner debug enabled is kept as is

### LINKED ISSUE #4050: Handle axios mock adapter module not found problem (leading to flaky tests)
## Situation
We have often the situation that npm actions do fail / become flaky because a module cannot be found - e.g.

Image

## Wanted
The problem shall no longer happen and the npm tests shall be stable

### Make builds reproducable
Every npm based build at github actions shall use `npm ci` instead of `npm install`

See https://docs.npmjs.com/cli/v9/commands/npm-ci for details

We must adopt the README files as well.

### Handle import warnings from jest
We have many warnings from jest about import problems and some hints.

To avoid this and to become more es compatible we add following to `tsconfig.json`:
```json
// try to avoid flaky tests /import problems 
    "esModuleInterop": true,
    "typeRoots": [
      "./node_modules/@types",
      "./types"
    ]
```

### LINKED ISSUE #4103: Provide `build` as used go client for github action
## Situation
Sometimes we have odd situations with GitHub workflow runs where the go client (which is used by our GitHub action) does not have access to environment variables or has odd behavior on template treatment etc.

Currently only released clients can be used for scanning - means we have no possibility to test changes on go client in combination with GitHub action before doing a release.

## Wanted
It shall be possible to test also unreleased go clients in combination with GitHub action.
This shall be possible even for GitHub actions running on a repository different than SecHub repository.

### Introduce new client version `build` 
It shall be possible to define not only `latest` or a dedicated version for the run, but also `build`.

In this case there shall be no download of the version, but instead it is assumed that the GitHub workflow which runs the GitHub action has also build a go version locally.

The local build will be used - but will NOT be cached

### Introduce new variable `build_client_target_folder`
Because the build could be done on another location we provide a new variable:
``build_client_target_folder`` which is the directory where the SecHub go client has been build.

For example: `../../sechub-cli/build/go/platform/linux-amd64

Inside this location we expect a file `sechub` which then will be used for execution.

#### Variant A: SecHub integration tests for SecHub GitHub action
Here we have already go client sources available and can build just be calling 
`gradlew buildGo` - afterwards it is available in GitHub Workspace

### Variante B: Testrepository which wants to scan via SecHub

Here we would introduce three steps inside workflow
  
- Step 1: setup go, git etc. (maybe already done - just to mention necessary)
- Step 2: clone SecHub FOSS repo via shell commands to a temporary folder
- Step 3: build go client via `gradlew buildGo`
- Step 4: use github action with `build` and dedicated `build_client_target_folder`

### LINKED ISSUE #4107: GitHub action unit test for project name resolver is flaky
## Situation

The tests inside `projectname-resolver.test.ts` are sometimes failing - means flaky
Image

## Wanted 
Tests shall not be flaky

### Debug logging
In #4105 we added debug possibility for the GitHub action workflow (just restart a action job with debug logging) and it appears following:

Image

Means not the wanted file content is returned but somethingelse.

### Further investigations
Inside `post-scan.test.ts` I found:
```typescript
  const fsMock = require('fs');
  /* prepare */
  const testContext = Object.create(LAUNCHER_CONTEXT_DEFAULTS);
  testContext.reportFormats = ['json', 'html', 'xyz', 'bla'];
  testContext.jobUUID = 1234; // necessary for download

  fsMock.readFileSync = jest.fn(() => '{"test": "test"}'); // Mock an empty JSON report
```

==> a jest mock is here wrong used! The "require" fetches the origin fs function (not a mock!) and the last line in the code before does completely overwrite `readFileSync` -> this will not be reset! 

## Solution
Replace the mock approach with working variant without side effects.

### LINKED ISSUE #4113: GitHub action build shall start also when go client changes are detected
## Situation 
With #4103 we changed also the default client version inside the integration test to `build` and use the current built go client

But changes on go-client do not automatically trigger the integration tests for the github-action (to ensure github-action can work with the new go client)

## Wanted
GitHub action workflow shall also be triggered on changes on go-client

de-jcup:
This PR 

- [x] closes #4103
- [x] closes #4107
- [x] closes #3993
- [x] closes #4113
- [x] closes #4050

### DIFF
--- a/.github/workflows/github-action-scan.yml
+++ b/.github/workflows/github-action-scan.yml
@@ -37,19 +37,32 @@ jobs:
         run: |
           echo "User input for client-version: '${{ inputs.client-version }}'"
           SECHUB_CLIENT_VERSION="${{ inputs.client-version }}"
+          
           if [ "$SECHUB_CLIENT_VERSION" = "" ]; then
-             SECHUB_CLIENT_VERSION=build 
+
              echo "- not defined, set default"
+             
+             # set default value for version:
+             SECHUB_CLIENT_VERSION=build
+             
           else
              echo "- using defined version"
           fi
-          echo "SECHUB_CLIENT_VERSION=$SECHUB_CLIENT_VERSION"
+          
+          echo "- SECHUB_CLIENT_VERSION=$SECHUB_CLIENT_VERSION"
           echo "SECHUB_CLIENT_VERSION=$SECHUB_CLIENT_VERSION" >> $GITHUB_ENV
+          
+          if [ "$SECHUB_CLIENT_VERSION" = "build" ]; then
+             # set client build folder:
+             SECHUB_CLIENT_BUILD_FOLDER=${{ github.workspace }}/sechub-cli/build/go 
+             echo "- SECHUB_CLIENT_BUILD_FOLDER=$SECHUB_CLIENT_BUILD_FOLDER"
+             echo "SECHUB_CLIENT_BUILD_FOLDER=$SECHUB_CLIENT_BUILD_FOLDER" >> $GITHUB_ENV
+          fi
           
       - name: "Show client version"
         working-directory: ~
         run: |
-          echo "client-version in env now: '${{ env.SECHUB_CLIENT_VERSION }}'"
+          echo "env.SECHUB_CLIENT_VERSION='${{ env.SECHUB_CLIENT_VERSION }}'"
       
       - name: Checkout
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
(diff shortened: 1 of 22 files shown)


winzj (approved):
Looking good to me :+1: 

I first thought the build failed but the java build was canceled, because it is not necessary.

PR #4115, opened 2025-04-22
Introduce minimum token validity #4033

### LINKED ISSUE #4033: Provide possibility to define minimumCookieValidity
## Situation
We provide SecHub WebUI login with OAuth2. 

Let's assume that the used IDP provider does have quotas but still short expiration times, or there are configuration problems /changes etc.

This could lead to a bad user experience when always being redirected to login mask after a short time.

Addition: #3892 did check the introspection result and does a logout when expired - this means we must not only set the defaultTokenExpiration (if missing) but also ensure that the minimum is used also when IDP value is to low

## Wanted
It shall be possible to provide an optional `minimumTokenValidity` to handle this.
If the value is not set/equals 0, nothing special shall happen, but if set, after a successful login the oauth2 expiration time shall be never smaller than the `minimumTokenValidity`. 

## Solution
We have already a `sechub.security.server.oauth2.opaque-token.default-token-expires-in` for missing `expiresAt` on oauth2 opaque token handling, but this value is only used for opaque token handling .

The `minimumTokenValidity` shall be used for any type of login were a cookie is used: Means if we implement a expiration time for the classic mode (+cookie) this could be used as well.

At this issue 

- [ ] every setting of the OAUTH2 cookie shall set a value which leads to a cookie which is at least valid until `minimumTokenValidity`
- [ ] this includes the browser cookie setting but also the content of the encrypted OAuth2 token data
- [ ] a change of the IDP vaue to minimum settings shall be logged at debug level
- [ ] set inside opque token inspection result when IDP is too low

winzj:
- closes #4033


de-jcup (changes requested):
Good job, but I have found some things I would like to have changed. Please read me comments.

sechub-commons-security-spring/src/test/java/com/mercedesbenz/sechub/spring/security/LoginOAuth2SuccessHandlerTest.java
```diff
@@ -65,7 +68,10 @@ void on_authentication_success_sends_a_valid_redirect_containing_the_encrypted_a
         Instant now = Instant.now();
         when(oAuth2AccessToken.getIssuedAt()).thenReturn(now);
         /* setting this should make sure that the default expiry (1 hour) is not used */
-        when(oAuth2AccessToken.getExpiresAt()).thenReturn(now.plusSeconds(expiry.toSeconds()));
+        Instant nowPlusOneSecond = now.plusSeconds(expiry.toSeconds());
```
de-jcup, line 71:
```java
Duration expiry = Duration.ofMinutes(1);
Instant nowPlusOneSecond = now.plusSeconds(expiry.toSeconds());
```
==> means:
nowPlusOneSecond = now + oneMinuteInSeconds 

==> naming ... is not a "nowPlusOneMinute" instead?

sechub-commons-security-spring/src/test/java/com/mercedesbenz/sechub/spring/security/OAuth2TokenExpirationCalculatorTest.java
```diff
@@ -0,0 +1,176 @@
...
+    void isExpired_throws_illegal_argument_exception_if_now_is_null() {
+        /* prepare */
+        OAuth2OpaqueTokenIntrospectionResponse response = mock();
+
+        /* execute + test */
+        assertThatThrownBy(() -> calculatorToTest.isExpired(response, null)).isInstanceOf(NullPointerException.class).hasMessageContaining("now");
+    }
+
+    @Test
+    void isExpired_throws_illegal_argument_exception_if_now_is_null_() {
```
de-jcup, line 40: `isExpired_throws_illegal_argument_exception_if_now_is_null_` is not correct
please rename to
`isExpired_throws_illegal_argument_exception_if_response_is_null`

sechub-commons-security-spring/src/test/java/com/mercedesbenz/sechub/spring/security/OAuth2TokenExpirationCalculatorTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+        /* execute */
+        boolean expired = calculatorToTest.isExpired(response, now);
+
+        /* test */
+        assertThat(expired).isEqualTo(shallBeExpired);
+    }
+
+    @Test
+    void now_null_throws_exception() {
```
de-jcup, line 60: Please add a `calculateAccessTokenDuration_` before the test methods which test `calculateAccessTokenDuration` like it was done by `isExpired_` methods for `isExpired`. So it is always clear what is tested here...

sechub-commons-security-spring/src/test/java/com/mercedesbenz/sechub/spring/security/OAuth2TokenExpirationCalculatorTest.java
```diff
@@ -0,0 +1,176 @@
...
+        /* prepare */
+        Duration defaultDuration = mock();
+        OAuth2AccessToken oAuth2AccessToken = mock();
+        Duration minimumTokenValidity = mock();
+
+        /* test */
```
de-jcup, line 66: If possible replace here (and others were test and execution are together) `/* test */` with `/* execute + test */`

sechub-commons-security-spring/src/test/java/com/mercedesbenz/sechub/spring/security/OAuth2TokenExpirationCalculatorTest.java
```diff
@@ -0,0 +1,176 @@
...
+        assertThat(calculatedDuration).isEqualTo(expectedDefaultExpiry);
+    }
+
+    @Test
+    void on_authentication_success_assumes_expires_from_token_when_greater_than_default_and_minimum() {
+        /* prepare */
+        Instant now = Instant.now();
+        Duration minimumTokenValidity = DEFAULT_EXPIRY.plusDays(1);
+
+        Instant expiresAtFromToken = Instant.now().plusSeconds(minimumTokenValidity.plusDays(5).getSeconds());
```
de-jcup, line 145: I would always use the existing `now` variable for later calculations instead of creating a Instant.now() for others. In former test you did it this way.

But just a suggestion.

de-jcup (approved):
Well done :+1: - I found some minor parts but approved the PR anyway. Please read my comments and adopt like you want.

PR #4116, opened 2025-04-22
Add optional email rule ends with domain #2374

### LINKED ISSUE #2374: Provide a mail rule for updates and creation of user accounts
### Situation
For some situations (e.g. in a company) it shall only be allowed to create users with specific mail addresses.
Currently every email can be used to change or create an account.

### Wanted
As an administrator I want to be able to define at deployment time which domains are allowed for email addresses

### Solution
- We introduce a rule (interface) `EmailRule` which is used by the existing mail validation mechanism (this is used for update and also for create workflow)
- Injected by Spring
- Implement RestrictMailDomains rule where we introduce Environment variable `EMAIL_RULE_ALLOWED_DOMAINS`. This entry can contain comma separated list of domains (e.g. "example.com,example.org" ==> user@example.com is valid, user@somewhere.com is not valid...
- handled by normal validation mechanism - no special treatment necessary
- if no email rule is defined (default), every email is allowed (as before), so this feature is only optional (where necessary...)

winzj:
- closes #2374


lorriborri (approved):
Looks good to me! Well done! :star: 🐇

PR #4119, opened 2025-04-23
Limit file size upload on client side 

### LINKED ISSUE #3777: Enhance large file upload web-ui
## Situation
The Web-UI file upload for binaries and source code may need to be revised to ensure that large files can be uploaded and scanned without errors.
In general the Web-Ui should be used for smaller files and for large files, the Go-Client is suitable.

## Wanted
The file upload functionality should be capable of handling larger files without resulting in errors.

## Solution
Revise the Web-UI file upload process to support the scanning of large files, preventing errors like NS_ERROR_NET_RESET or 500 Internal error from occurring.

Details to be defined:
- upload file in chucks
- use Prepare phase to recreate the original file
- adjust headers

### LINKED ISSUE #4118: Web-ui: check file size before upload
## Situation
We do not check the file size before we upload it to the server.

## Wanted
get a warning that the file-size is too large before trying to create a scan

## Solution
1. introduce same ENV as in sechub server: 
SPRING_SERVLET_MULTIPART_MAX_FILE_SIZE 
or SPRING_SERVLET_MULTIPART_MAX_REQUEST_SIZE
SECHUB_UPLOAD_BINARIES_MAXIMUM_BYTES
2. check if file size is over the limit
3. show alert if it is

lorriborri:
- closes #4118 
- closes #3777


winzj (approved):
Looking good :cookie: 

I just found one thing you could look into. But i you need the error message you can ignore this comment.

PR #4123, opened 2025-04-28
Add proxy to client download using http(s)_proxy #4060

### LINKED ISSUE #4060: Axios should respect proxy variables
### Description
It is possible that if you configure a GitHub Runner to operate behind a proxy, Axios may fail to download files from GitHub.com. Although the runner is set up to use the proxy for all HTTP and HTTPS requests, Axios might not utilize these proxy settings, potentially resulting in a 502 error.

### Steps to Reproduce
Configure a GitHub Runner to operate behind a proxy.
Attempt to download a file from GitHub.com using Axios within the runner.
Observe the failure with a 502 status code.

### Expected Behavior
Axios should use the proxy settings configured for the GitHub Runner and successfully download the file from GitHub.com.

### Actual Behavior
Axios does not use the proxy settings, leading to a 502 error when attempting to download the file.

### Possible Solution
It appears that Axios may need explicit proxy configuration. Investigating how to set the proxy settings for Axios within the GitHub Runner environment could resolve this issue.

winzj:
- closes #4060

### DIFF
--- a/github-actions/scan/README.adoc
+++ b/github-actions/scan/README.adoc
@@ -13,6 +13,16 @@ This GitHub action uses the SecHub cli to scan the repository for security issue
 
 To be able to use this action you need a SecHub project. Check the https://mercedes-benz.github.io/sechub/[documentation] on how to set one up.
 
+If the action runs on a custom runner and a proxy is necessary to download the sechub client from https://github.com,
+specify the proxy using the default ENV variables (lowercased like used by many other cli tools) `http_proxy` or `https_proxy`.
+Make sure to use the format, like this:
+- http_proxy="http://user:password@proxy.example.org:1234"
+- https_proxy="https://user:password@proxy.example.org:1234"
+
+or if you do not need credentials:
+- http_proxy="http://proxy.example.org:1234"
+- https_proxy="https://proxy.example.org:1234"
+
 [source,yaml]
 ----
 - uses: mercedes-benz/sechub/github-actions/scan@master
--- a/github-actions/scan/README.adoc
+++ b/github-actions/scan/README.adoc
@@ -15,7 +15,7 @@ To be able to use this action you need a SecHub project. Check the https://merce
 
 If the action runs on a custom runner and a proxy is necessary to download the sechub client from https://github.com,
 specify the proxy using the default ENV variables (lowercased like used by many other cli tools) `http_proxy` or `https_proxy`.
-The ENV variable `https_proxy` will be used if  `http_proxy` and `https_proxy` are set.
+The ENV variable `https_proxy` will be used if `http_proxy` and `https_proxy` are set.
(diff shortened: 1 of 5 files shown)


github-actions/scan/README.adoc
```diff
@@ -13,6 +13,19 @@ This GitHub action uses the SecHub cli to scan the repository for security issue
 
 To be able to use this action you need a SecHub project. Check the https://mercedes-benz.github.io/sechub/[documentation] on how to set one up.
 
+If the action runs on a custom runner and a proxy is necessary to download the sechub client from https://github.com,
+specify the proxy using the default ENV variables (lowercased like used by many other cli tools) `http_proxy` or `https_proxy`.
+The ENV variable `https_proxy` will be used if  `http_proxy` and `https_proxy` are set.
```
de-jcup, line 18:
```suggestion
The ENV variable `https_proxy` will be used if `http_proxy` and `https_proxy` are set.
```

github-actions/scan/src/input-helper.ts
```diff
@@ -16,3 +19,62 @@ export function split(input: string): string[] {
...
+      search: '',
+      searchParams: URLSearchParams {},
+      hash: ''
+    }
+ * @returns configured AxiosProxyConfig or undefined if no proxy was
+ * @throws error if the proxy URL inside the ENV variable is an invalid URL
+ */
+export function resolveProxyConfig(): AxiosProxyConfig | undefined {
+    const httpsProxy = process.env.https_proxy || undefined;
+    const httpProxy = process.env.http_proxy || undefined;
```
de-jcup, line 46: I am asking myself if this wouldn't be just enough:

```suggestion
    const httpsProxy = process.env.https_proxy;
    const httpProxy = process.env.http_proxy;
```

github-actions/scan/src/input-helper.ts
```diff
@@ -16,3 +19,62 @@ export function split(input: string): string[] {
...
+}
+
+function getProtocolDefaultPort(protocol: string) {
+    if (protocol === 'http:') {
+        return 80;
+    } else if (protocol === 'https:') {
+        return 443;
+    } else {
+        throw new Error('Accepted protocols are "http" and "https"');
+    }
```
de-jcup, line 79: Hmm... you rely on the assumption that the protocol will always contain the colon at the end. Maybe this would be easier/more resilient:

```suggestion
    if (!protocol){
     throw new Error('No protocol defined!');
    }
    if (protocol.startsWith('https') {
        return 443;
    } else if (protocol.startsWith('http') {
        return 80;
    } else {
        throw new Error('Accepted protocols are "http" and "https"');
    }
```

de-jcup (approved):
Good job :rocket: 
I found some parts which could be improved, but it's up to you if you change it or not. Please read my comments.

PR #4125, opened 2025-04-28
Handle templating problems between client and github action

### LINKED ISSUE #4124: Client problems with environment variables
## Situation

When a GitHub action workflow is defined and the config file contains a `{{ .PROJECT }}` inside, the value is not filled by template, but tried to be uppercased:

```
NOTICE: Converted project id '{{ .PROJECT }}' to lowercase because it contained uppercase characters, which are not accepted by SecHub server.
```

The Server will stop processing, because '{{ .PROJECT }}' is not a valid project id...

Irritating part: Works local out of the box, but not remote with GitHub action.

## Wanted
- Shall not happen.
- Shall work same for local starts with client and also with remote usage by GitHub action

### LINKED ISSUE #4129: Temp directory for integration tests with sechub client are not automatically cleaned up
## Situation
When an integration test uses the withSecHubClient() method, a temporary folder inside the gradle build folder will be created and used for SecHub client output etc. (e.g. `with-sechub-client-tmp_23717171771`.

But unfortunatately after the tests have been run, the build still contains the temp files.

For local builds or build on Jenkins this can lead to storage memory waste (when the gradle clean task is not called).

## Wanted
Created temp folders by TestUtil shall be automatically deleted recursively when integration test has been done (except when explicit delete is turned off by settings)

de-jcup:
This PR
- closes #4124
- closes #4129


sven-dmlr (approved):
Looks good :+1:

PR #4130, opened 2025-04-30
Feature 4128 client enable zero returncode on red

### LINKED ISSUE #4128: Client: Enable a zero returncode on red findings
## Situation
Currently the client return a non-zero value when there are critical or high findings.

As discussed [here](https://github.com/mercedes-benz/sechub/issues/2841#issuecomment-2275063367) it might make sense to always return zero even if there are red findings.

## ToDo
Introduce an environment variable `SECHUB_FAIL_ON_RED` which is `true` per default.
If set to `false` the client shall return 0 even if there are critical or high findings.

sven-dmlr:
- closes #4128


haerter-tss (approved):
Looks good! 🔮🪭

PR #4140, opened 2025-05-06
Enhanced performance for project data fetching #4136

### LINKED ISSUE #4134: Optimize performance fetching profile IDs of the users projects
### Situation
A user can fetch data for all assigned projects. To fetch the profile IDs of the users projects the database is called, see: https://github.com/mercedes-benz/sechub/blob/13a9d3e5da1874b4d7000eb06b523c317aabc376/sechub-scan/src/main/java/com/mercedesbenz/sechub/domain/scan/product/config/ProfileMessageHandler.java#L39

For a lot of cases this could work just fine, but a user, who is assigned to many projects with many profiles assigned this approach might lead to performance issues.

### Wanted
The database could handle the computation of the profile IDs of the users projects, to avoid multiple database interactions.

### LINKED ISSUE #4136: Increase performance on fetching project information
## Situation
When we have the situation that there are many associations between users and projects, we have a lot of memory consumption. And the response is too slow.

## Wanted
Shall be faster for 
- fetch project ids
- fetch project details about a project

## Solution
- [x] project id fetching is currently done by fetching project entities and mapping their ids... not very efficient... change this to simple DB query which just returns the ids (also order it now )
- [ ] Projects and user entities have a m:n association which is fetched eagerly. This
  behavior must be changed

### LINKED ISSUE #4142: Project owner should get Project Data when not assigned
## Situation
Currently, the API endpoint /api/projects delivers only the ProjectData of the assigned projects.

## Wanted
The endpoint should also return data for owned projects.

## Solution
Modify the REST call to include and return data for owned projects in addition to the assigned projects.

de-jcup:
This PR
- [x] closes #4136 
- [x] closes #4134
- [x] closes #4142

### DIFF
--- a/sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/job/JobCancelService.java
+++ b/sechub-administration/src/main/java/com/mercedesbenz/sechub/domain/administration/job/JobCancelService.java
@@ -9,11 +9,10 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
 
-import com.mercedesbenz.sechub.domain.administration.project.Project;
+import com.mercedesbenz.sechub.domain.administration.project.ProjectRepository;
 import com.mercedesbenz.sechub.domain.administration.user.User;
 import com.mercedesbenz.sechub.domain.administration.user.UserRepository;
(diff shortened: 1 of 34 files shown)


winzj (approved):
Good job looking very good to me :rocket:

PR #4141, opened 2025-05-06
Mark web scan FP in web-ui

### LINKED ISSUE #4135: Web-ui mark false positives for webScans
## Situation
Currently only codeScan findings can be marked as false- positives 

## Wanted
User should be able to mark also webScan false-positives via web-ui

## Solution

- implement FP dialog (like GH security)
- imlement pattern logic

lorriborri:
- closes #4135


de-jcup (approved):
Looks very good to me :rocket: - I found 2 minor parts which could be changed (both are typos). Please check if you want to change them before the merge.

PR #4170, opened 2025-05-19
Feature 4169 fix bug in CheckmarxResilienceConsultant

### LINKED ISSUE #4169: Fix Bug in CheckmarxResilienceConsultant für checkmarx
## Situation
we catch on exceptions of HttpClientException but not HttpServerException, therefore the resiliance proposal for server error 500 will never be returned
```java
 if (rootCause instanceof HttpClientErrorException) {
...
}
```

## Wanted 
- resolve bug to handle server errors 
- improve logging
- add WireMock tests

## Solution
- fix rootcause handling 
```java
if (rootcause instanceof RestClientException){
// catches every rest problem
    if (rootCause instanceof HttpStatusCodeException)
       // do it
       if(statusCode >= 500){
         ...
       }
    }else{
        // log that we cannot handle it...
    }
}

 ```
- improve logging
- add tests

lorriborri:
- closes #4169


de-jcup (changes requested):
Good work - but I found something inside the tests (separation). Please change this.

de-jcup (approved):
Looks good to me :rocket:

PR #4178, opened 2025-05-19
Fix encoding problem #4172

### LINKED ISSUE #4172: Umlaut problems with filenames inside ZIP and TAR files
## Situation
When a file to scan contains umlauts, currently the extraction of the archives by `ArchiveSupport` makes problems/the file is just not contained on recompression (e.g. for Checkmarx).

## Wanted
Extraction shall be able to handle umlauts correctly

## Solution
- analyze problem and find a solution
- check if this only for some security products or a general problem

de-jcup:
This PR
- closes #4172


sven-dmlr (approved):
Looks good to me :+1: 

Maybe you want to rename the folder
"contains-files-with-special-charackters"
to 
"contains-files-with-special-characters"

Same for "contains-files-with-special-charackters.zip"
-> "contains-files-with-special-characters.zip"

PR #4192, opened 2025-05-23
Keycloak test container

### LINKED ISSUE #4191: Test Container for KeyCloak
## Situation
When we configure oauth for the sechub server, we need to set up a IDP server manually (run keycloak docker image manually and configure)

## Wanted
A out-of-the box solution which can be used for integration test 

## Solution

- set up a docker and Dockerfile for keycloack
- add define and add custom realm to the start up
- implement a class (like LocalTestPostgreSQLStarter) to start and stop the container

we do not use https://github.com/dasniko/testcontainers-keycloak, to make the docker solution reusable in helm charts for a local setup.

lorriborri:
- closes #4191

### DIFF
--- a/sechub-developertools/scripts/container/common-containerscript.sh
+++ b/sechub-developertools/scripts/container/common-containerscript.sh
@@ -34,7 +34,7 @@ function defineContainerPort(){
 
 #
 # param 1: exposed port, may not be empty
-function defineEexposedPort(){
+function defineExposedPort(){
     assertNotEmpty "port not defined" $1
     
     exposed_port=$1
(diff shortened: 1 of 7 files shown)


sechub-developertools/scripts/container/keycloak/application-local-test-keycloak-template.yaml
```diff
@@ -0,0 +1,32 @@
+# SPDX-License-Identifier: MIT
+sechub:
+  security:
+    server:
+      modes: oauth2, classic
+      oauth2:
+        mode: opaque-token
```
de-jcup, line 7: must be changed to `jwt` - keycloak does not provide opaque-token

PR #4211, opened 2025-05-28
Introduce PDSSafeProcessEnvironmentAccess #4209

### LINKED ISSUE #4209: PDS Job parameter with unset values lead to exception
## Situation
If we have a PDS job parameter which is defined this way:

```
test.key=env:YOUR_VARIABLE
```
It can happen that the environment variable (on SecHub Server) is not defined/not set.

In this case following exception is thrown:

```java
java.lang.NullPointerException: Cannot invoke "String.indexOf(int)" because "<parameter1>" is null
	at java.base/java.lang.ProcessEnvironment.validateValue(Unknown Source)
	at java.base/java.lang.ProcessEnvironment$Value.valueOf(Unknown Source)
	at java.base/java.lang.ProcessEnvironment$StringEnvironment.put(Unknown Source)
	at java.base/java.lang.ProcessEnvironment$StringEnvironment.put(Unknown Source)
	at com.mercedesbenz.sechub.pds.execution.PDSExecutionEnvironmentService.addJobParamDataWhenAccepted(PDSExecutionEnvironmentService.java:217)
	at com.mercedesbenz.sechub.pds.execution.PDSExecutionEnvironmentService.addPdsExecutorJobParameters(PDSExecutionEnvironmentService.java:129)
	at com.mercedesbenz.sechub.pds.execution.PDSExecutionEnvironmentService.calculateAndSetupEnvironment(PDSExecutionEnvironmentService.java:76)
	at com.mercedesbenz.sechub.pds.execution.PDSExecutionEnvironmentService.initProcessBuilderEnvironmentMap(PDSExecutionEnvironmentService.java:67)
	at com.mercedesbenz.sechub.pds.execution.PDSExecutionCallable.createProcess(PDSExecutionCallable.java:481)
	at com.mercedesbenz.sechub.pds.execution.PDSExecutionCallable.call(PDSExecutionCallable.java:125)
	at com.mercedesbenz.sechub.pds.execution.PDSExecutionCallable.call(PDSExecutionCallable.java:53)
	at java.base/java.util.concurrent.FutureTask.run(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.base/java.lang.Thread.run(Unknown Source)
```

### Analyze
The problem is, that the environment map for process is not able to handle null values (empty strings are okay, but not null).
If a key is not null but the value is null the exception will occur.

## Wanted
- If a key is null it shall be ignored
- if a value is null a warning shall be logged about missing value for the key and it shall be skipped.

de-jcup:
This PR
- closes #4209

### DIFF
--- a/sechub-pds/src/main/java/com/mercedesbenz/sechub/pds/execution/PDSExecutionEnvironmentService.java
+++ b/sechub-pds/src/main/java/com/mercedesbenz/sechub/pds/execution/PDSExecutionEnvironmentService.java
@@ -88,7 +88,7 @@ public class PDSExecutionEnvironmentService {
 
         environmentCleaner.clean(environment, pdsScriptEnvWhitelist);
 
-        PDSSafeProcessEnvironmentAccess result = new PDSSafeProcessEnvironmentAccess(builder.environment());
+        PDSSafeProcessEnvironmentAccess result = new PDSSafeProcessEnvironmentAccess(environment);
 
         return result;
 
--- a/sechub-pds/src/main/java/com/mercedesbenz/sechub/pds/execution/PDSExecutionEnvironmentService.java
+++ b/sechub-pds/src/main/java/com/mercedesbenz/sechub/pds/execution/PDSExecutionEnvironmentService.java
@@ -6,7 +6,6 @@ import static com.mercedesbenz.sechub.commons.pds.PDSLauncherScriptEnvironmentCo
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import java.util.Map.Entry;
 import java.util.Set;
 import java.util.UUID;
 
@@ -70,7 +69,7 @@ public class PDSExecutionEnvironmentService {
 
     private void calculateAndSetupEnvironment(UUID pdsJobUUID, PDSJobConfiguration config, ProcessBuilder builder, PDSProductSetup productSetup) {
         /* now init environment map */
-        Map<String, String> environment = initCleanEnvironment(productSetup, builder);
+        PDSSafeProcessEnvironmentAccess environment = initCleanEnvironment(productSetup, builder);
 
         addPdsJobRelatedVariables(pdsJobUUID, environment);
         addPdsExecutorJobParameters(productSetup, config, environment);
@@ -81,7 +80,7 @@ public class PDSExecutionEnvironmentService {
         LOG.debug("Initialized environment variables for script of pds job: {} with: {}", pdsJobUUID, environment);
     }
 
-    private Map<String, String> initCleanEnvironment(PDSProductSetup productSetup, ProcessBuilder builder) {
+    private PDSSafeProcessEnvironmentAccess initCleanEnvironment(PDSProductSetup productSetup, ProcessBuilder builder) {
         Map<String, String> environment = builder.environment();
 
         Set<String> pdsScriptEnvWhitelist = productSetup.getEnvWhitelist();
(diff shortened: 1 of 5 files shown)


sechub-pds/src/main/java/com/mercedesbenz/sechub/pds/execution/PDSExecutionEnvironmentService.java
```diff
@@ -81,55 +80,57 @@ private void calculateAndSetupEnvironment(UUID pdsJobUUID, PDSJobConfiguration c
...
+    private PDSSafeProcessEnvironmentAccess initCleanEnvironment(PDSProductSetup productSetup, ProcessBuilder builder) {
         Map<String, String> environment = builder.environment();
 
         Set<String> pdsScriptEnvWhitelist = productSetup.getEnvWhitelist();
         LOG.debug("PDS script environment variable white list: '{}'", pdsScriptEnvWhitelist);
 
         environmentCleaner.clean(environment, pdsScriptEnvWhitelist);
 
-        return environment;
+        PDSSafeProcessEnvironmentAccess result = new PDSSafeProcessEnvironmentAccess(builder.environment());
```
lorriborri, line 91: I think the environment should be hand over here? - otherwise the environmentCleaner has no effect.
```suggestion
        PDSSafeProcessEnvironmentAccess result = new PDSSafeProcessEnvironmentAccess(environment);
```

sechub-pds/src/test/java/com/mercedesbenz/sechub/pds/execution/PDSSafeProcessEnvironmentAccessTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+        /* execute */
+        pdsSafeProcessEnvironmentAccess.put(key, value);
+
+        /* test */
+        verify(environment).put(key, value);
+    }
+
+    @Test
+    void get_keys_fetchs_keys_from_origina_map() {
```
lorriborri, line 67: Typo :)
```suggestion
    void get_keys_fetches_keys_from_original_map() {
```

sechub-pds/src/test/java/com/mercedesbenz/sechub/pds/execution/PDSSafeProcessEnvironmentAccessTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+        /* execute */
+        Set<String> keys = pdsSafeProcessEnvironmentAccess.getKeys();
+
+        /* test */
+        assertThat(keys).containsExactlyInAnyOrder("key1", "key2");
+    }
+
+    @Test
+    void get_existing_key__returns_value_from_origin_map() {
```
lorriborri, line 83: Just optional:
sometimes you use original, sometimes origin - I think, both is valid, but I would not recommend to mix it and maybe go for original map (cause you are referring to the underlying map)

lorriborri (approved):
LGTM! :star: 🐕
I might have found one small issue - please see my comments :)

sechub-pds/src/main/java/com/mercedesbenz/sechub/pds/execution/PDSExecutionEnvironmentService.java
de-jcup: Thank you for finding this! You are absolutely right - this was wrong!

PR #4226, opened 2025-06-11
Fix web-ui redirects

### LINKED ISSUE #4225: Fix logout and redirect for web-ui
## Situation 
The logout from the web-ui does not always work correctly, problem is the redirect

## Wanted
The web-ui should follow the redirect from the server

## Solution
use window.location.href = '/logout'
instead of openapi client to follow the redirects correctly

lorriborri:
- closes #4225


hamidonos (approved):
good job

PR #4230, opened 2025-06-11
Feature 4177 enhance cancellation info and logs

### LINKED ISSUE #4176: Cancelling a job shall not produce executor error entries in logs
## Situation
Currently a cancel operation by a user cancels the job but this will lead to execution failures which appears in logs. Those log entries look same as log entries for jobs where product execution really fails.

This is very irritating when monitoring logs are inspected.

## Wanted
It shall be clear inside logs that a canceled operation is not an execution failure

## Solution
We have the information that a job failed because a cancel operation was requested.
In this case we will no longer log an error, but just an information that the job has been canceled.

### LINKED ISSUE #4177: Cancel operation shall produce info message in report instead internall error info
## Situation
Currently when a user cancels a job a downloaded job contains the information
that an internal error has happened. This looks like the execution failed - but it was canceled by user.

This is a bit related to #4176 

## Wanted
It shall be clear that the operation has been canceled and that no internal error happened

## Solution
We change the implementation in a way that
- no internal error message is added when execution was canceled
- an user info message: "Job has been canceled" will be added when a cancel operation was 
  triggered

### LINKED ISSUE #4229: Remove unnecessary inheritance to SecHubModel
## Situation
The class `ReportTransformationResult` does currently inherit from
`SecHubReportModel`. 

The result contains a field that is available via getter and setter - which may never be part of JSON.

If somebody extends this class and forgets to ignore the new fields for JSON, this could have side effects.

Also it produces an unnecessary complexity when we want to replace commons-model parts with generated java classes (by openapi definition).

## Wanted
The class `ReportTransformationResult` shall no longer inherit from `SecHubReportModel`

de-jcup:
This PR

- [ ] closes #4177 
- [ ] closes #4176 
- [ ] closes #4229


winzj (approved):
Looking good to me :rocket:

PR #4231, opened 2025-06-11
Removed NP exception from missing minimumTokenValidity

### LINKED ISSUE #4214: Remove NP exception for minimumTokenValidity
## Situation
If the minimumTokenValidity is not set in the spring properties we receive an NP although it is ok if minimumTokenValidity is null

## Wanted
No exception should be thrown

## Solution
remove: `requireNonNull(minimumTokenValidity, "Parameter 'minimumTokenValidity' may not be null!");` in OAuth2TokenExpirationCalculator

add: Tests in SecHubSecurityPropertiesTest

lorriborri:
- closes #4214


winzj (approved):
Looking good to me :cake:

PR #4242, opened 2025-06-13
Added iacScan to Web-UI

### LINKED ISSUE #4236: Add IaC as possible scan type to web-ui
## Situation
We want to split codescan ind codescan and iacscan (infrastructure as code). The Web-ui does currently not offer IaC as scantype

## Wanted
- user should be able to select also Iac, IaC selection is restricted to source code upload only
- user should be able to see results in web-ui

## Solution
- add IaC as possible scantype
- check if IaC then only source code can be selected
- add creation of IaC config 
- add tests
- adjust the report views

lorriborri:
- closes #4236


hamidonos (approved):
🚀

PR #4244, opened 2025-06-13
Added success logging on retry #4243

### LINKED ISSUE #4243: OAuth2OpaqueTokenClusterCachePersistence shall log success in case of resilience retry
## Situation
In `OAuth2OpaqueTokenClusterCachePersistence` we have a simple resilience which logs warning on retries and does log an error in case that the even the last retry failed.

But it is not clear that the last retry did work

## Wanted
Log on info level when at least one retry was necessary and the execution was successful with a retry.

de-jcup:
This PR
- closes #4243


lorriborri (approved):
Good Job! :star:

PR #4249, opened 2025-06-18
parallelize gradle tasks & tests

### LINKED ISSUE #4212: Use parallelism in Gradle Tests to improve speed
## Situation

Currently all tests run in sequence. While that's necessary for stateful test modules it is not necessary for stateless tests (like ArchUnit tests). This slows the build down.

_Tests should fail fast_ is a best practice in software testing that means:

- When something is wrong, the test should stop as soon as possible and provide immediate feedback.

Why _fail fast_ matters:

- Quick feedback: You find out problems early in the execution cycle.
- Faster debugging: The root cause is easier to identify when fewer things have happened.
- Improved CI/CD: Reduces wasted compute time and speeds up pipelines.

## Quick POC

ArchUnit Test no parallelism:

Image

ArchUnit Test with parallelism:

Image

(Tested on my local machine)

## Wanted

Gradle supports parallelism. We should enable it for modules where it makes sense.

## Solution

- Validate if parallelism is a valid approach
- Enable parallelism globally `org.gradle.parallel=true` in `gradle.settings`
- Use half of the available JVM cores (or 1 for single core JVMs) for parallelism by setting `maxParallelForks= Runtime.runtime.availableProcessors().intdiv(2) ?: 1` for modules where parallelism is suitable.
- Identify stateful modules or modules where sequential tests are necessary. Set `maxParallelForks=1` for those modules.

## Notes

We should not use all available cores!

### Avoid CPU Saturation

Running too many test JVMs in parallel can:

- Overwhelm the CPU, causing heavy context switching.
- Slow down each test process due to contention for CPU, memory, and I/O.
- Lead to diminishing returns or even worse performance.

Using half the cores allows:

- Room for the Gradle daemon, background services, and the OS.
- Better overall stability, especially on developer machines.

### Balance Speed vs. Resource Usage

Most developers are multitasking:

- IDEs (like IntelliJ), browsers, Docker, etc. are all competing for resources.
- Overcommitting CPU to test forks can make the machine unresponsive.

Inspect and adapt:

- Use less cores if half of the cores are too much (e.g. 1/4)

hamidonos:
- closes #4212

### DIFF
--- a/sechub-archunit-test/build.gradle
+++ b/sechub-archunit-test/build.gradle
@@ -11,10 +11,10 @@ projectType.archUnitProjects.each { Project project ->
     evaluationDependsOn(project.path)
 }
 
-/* Define a task dependency to ensure all ArchUnit classes are built before running tests */
+/* Ensure all test classes are built before ArchUnit runs */
 tasks.named('test') {
     projectType.archUnitProjects.each { Project project ->
-        dependsOn "${project.path}:classes"
+        dependsOn "${project.path}:testClasses"
     }
 }
 


lorriborri (changes requested):
Good Work! :star: 
Please have a look at my comments :)

PR #4259, opened 2025-06-30
Feature 4258 web-ui user does not exist message

### LINKED ISSUE #4258: WebUi message if user does not exist
## Situation
if someone wants to add a user to his project, the user must exist in the sechub database.
Currently there is only an error shown, but not dedicated message

## Wanted
A message, that the user need to exist in the sechub context.

## Solution
- add a dedicated error message

lorriborri:
- closes #4258


haerter-tss (approved):
Looks good! 🎷

PR #4268, opened 2025-07-07
Feature 4262 vscode plugin generated client

### LINKED ISSUE #4262: Use generated client in VSCode plugin
## Situation

The current VSCode plugin implementation uses an duplicate impl. of the SeccHub Model

## Wanted

Plugin should run with the generated openapi client (as web-ui does)

## Solution

- make a shared openapi-typescript directory
- generate the client here (and add custom impl.)
- use the shared directory in web-ui and VSCode

lorriborri:
- closes #4262


ide-plugins/vscode/src/utils/sechubUtils.ts
```diff
@@ -0,0 +1,11 @@
+import { SecHubReport } from 'sechub-openapi-ts-client';
+
+export function loadFromFile(location: string): SecHubReport {
+
+    var fs = require('fs');
+    var obj = JSON.parse(fs.readFileSync(location, 'utf8'));
+
+    let report: SecHubReport = obj;
+
+    return report;
```
hamidonos, line 10:
```
const rawReport = fs.readFileSync(location, 'utf8');
return JSON.parse(rawReport) as SecHubReport;
```

I think this is cleaner

hamidonos (changes requested):
Looks good already

Please have a look at my findings

PR #4297, opened 2025-07-15
Feature 4288 GitHub action cancel also cancels sechub job

### LINKED ISSUE #4288: When Github action Workflow is cancelled trigger SIGINT and SIGTERM signal to client
### Situation
Currently the SecHub scan job is not cancelled and keeps running if the Github action workflow is cancelled.

### Wanted
When the Github Action is cancelled, the SecHub job shall be cancelled as well.

### Solution
Since the SecHub client is able to handle SIGINT and SIGTERM, the SecHub Github action must send  SIGINT and SIGTERM to the SecHub client. The SecHub client will then cancel the job.

winzj:
- closes #4288


github-actions/scan/action.yml
```diff
@@ -105,7 +105,7 @@ runs:
           core.exportVariable('ACTIONS_CACHE_URL', process.env['ACTIONS_CACHE_URL'])
 
     - name: Run SecHub Scan
-      run: node "$GITHUB_ACTION_PATH"/dist/index.js
+      run: exec node "$GITHUB_ACTION_PATH"/dist/index.js
```
de-jcup, line 108: Maybe a short comment would be nice (above run) - to explain why exec was necessary.

github-actions/scan/__test__/sechub-cli.test.ts
```diff
@@ -270,4 +300,104 @@ describe('defineFalsePositives', function () {
...
+        // Verify that the signal was forwarded
+        expect(mockChildProcess.kill).toHaveBeenCalledWith('SIGTERM');
+
+        // Simulate child process exit with signal
+        mockChildProcess.on.mock.calls[0][1](null, 'SIGTERM');
+
+        await expect(promise).rejects.toThrow('Process terminated by signal: SIGTERM');
+    });
+
+    it('should reject promise if child process encounters an error', async () => {
```
de-jcup, line 389: You have written tests for
- reject on error
- reject on sigterm
- forward SIGTERM
- forward SIGINT

But what is not tested is, that
- no reject when normal exit

but maybe this is not necessary, because we have the integration tests.

de-jcup (approved):
Very good job! 🚀 
I had only some comments - but it's up to you if you do some additions or not.

PR #4329, opened 2025-07-23
VSCode plugin enhancement and false positive support

### LINKED ISSUE #4327: VSCode: view job messages in plugin
## Situation
Currently job messages are not displayed

## Wanted 
View job info messages and metadata in plugin

## Solution
Add information to report view

### LINKED ISSUE #4328: VSCode: support simple WebScan view
## Situation
Currently webScans can not be imported or viewed in the plugin

## Wanted 
A simple view to see the findings, but do not jump to them

## Solution
- Add new class for the Info tree view
- do not show call hierarchy

### LINKED ISSUE #4340: VSCode: support mark false positives of code findings
## Situation
Currently we do not support false positive marking in the IDE Plugin

## Wanted
User should be able to mark false positives direktly from code

## Solution
Support false positive handling in VSCode Plugin.
Should look similar as web-ui

lorriborri:
- closes #4327 
- closes #4328 
- closes #4340

### DIFF
--- a/ide-plugins/vscode/src/api/defaultClient.ts
+++ b/ide-plugins/vscode/src/api/defaultClient.ts
@@ -136,30 +136,27 @@ export class DefaultClient {
     }
   }
 
-  public async markFalsePositivesForProject(falsePositves: FalsePositives, projectId: string): Promise<boolean> {
+  public async markFalsePositivesForProject(falsePositves: FalsePositives, projectId: string): Promise<void> {
 
     const requestParameter: UserMarkFalsePositivesRequest = {
       projectId: projectId,
       falsePositives: falsePositves
     };
 
-    // catch error is calling method to cache false positives if server could not be reached
     try {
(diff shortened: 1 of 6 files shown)


hamidonos (changes requested):
image

This message pops up always even when I only click the checkbox to mark a false positive


image

When I cancel at the "Enter your custom comment..." step it sends the request to the server and marks the false positive.. is this intended? It feels not right

image

Why not just put the "Mark false positive" action button next to the other action buttons on the top? Feels more intuitive this way

image

I've manually shut down the server and tried to sync the false positives. It tells me to select a valid project but the connection is simply missing. 


- You jump between `const` and `let` a lot. Try to keep one style unless needed otherwise.
- Your function generally miss input validation -> Increases risk for uncontrolled null pointers etc.
- You show an error message "Failed to fetch report" unless the report is in `ENDED` state when user tries to open the report. You could just do nothing if state is not `ENDED` 
- I thinks it's easier to write everything to the cache always and clear the cache when the sync with the server was successful.

PR #4339, opened 2025-07-31
Introduce false positive handling for eclipse plugin

### LINKED ISSUE #4332: Eclipse-Plugin: provide false positive handling
## Situation
Currently we cannot handle false positives directly inside the eclipse plugin

## Wanted 
The eclipse plugin shall have a similar user experience like web UI.

## Solution
1. SecHub report view will have an additional column for false positive (as an icon)
2. The current false positive setup for the current project is synchronized
3. We provide a mark and also an unmark action inside the context menu of the report finding table
4. We provide also a project false positives dialog, where findings can only be removed

Important: The mark/unmark actions are only provided for job based false positives. Generic definitions (like or web scan) are out of scope in this issue.

de-jcup:
This PR
- closes #4332


winzj (approved):
Looking very good :rocket: 

I tested the functionality locally and everything is working and looking good.
Just found some typos in Tool Tips/Error messages.

If it is possible and does not create a lot of changes on other gradle subprojects, maybe you could apply the license headers with the automated script.

PR #4343, opened 2025-08-04
Feature 4326 intellij plugin mark false positive

### LINKED ISSUE #4326: Mark false positives in IntelliJ Plugin
## Situation

Users are able to mark false positives through API calls or by using the WebUI.

## Wanted

We want to offer that functionality also inside the IntellliJ Plugin.

## Solution

- Implement false positive marking in IntelliJ Plugin
- Similar to WebUI look and feel
- DAST is excluded

hamidonos:
- closes #4326


lorriborri (approved):
Now it looks good to me! :star: 🦥
Lets merge!

PR #4374, opened 2025-08-15
Replace 406 with 404 #4372

### LINKED ISSUE #4372: Web-Ui: catch 404 when user is not in SecHub Context available
## Situation
When a project owner assigns new user to the project, the web-ui catches the 406 error, but we have change to back end to return a 404.

## Wanted
Catch 404 instead of 406

## Solution
In the project settings dialog catch 404 instead of the 406

Unfortunately the server message gets lost. so we can not check if maybe the project was not found or if the 404 come really from missing user.

lorriborri:
- closes #4372


haerter-tss (approved):
Looks good! 🪵🐩

PR #4416, opened 2025-09-10
OpaqueToken cluster cache add new primary key

### LINKED ISSUE #4260: Avoid database error logs because of duplicated keys on opaque token cluster caching
## Situation
Inside the database logs we have some error entries about duplicated keys when it comes to opaque token cluster caching.

There is already a resilience integrate in this case and the next retry will just do an update in database, without an error entry.

But ... inside runtime monitoring we have same database error log entries which is very irritating.

## Wanted
- There shall be no database error log entries.
- The solution shall be not database product specific

## Solution
We allow multiple entries for same opaque token. Primary key will be different

### SQL Schema
- Introduce new PK (by flyway)
- Change Entity and use generated PK
- Add index for opaque token id

### Java code
- Fetch last created opaque token info (instead only one)
- Create will be as before: If already available do an update, otherwise create
  If two entries are created, the last created one will be used.

### Cache auto cleanup:
- same as before, no changes necessary

lorriborri:
- closes #4260

### DIFF
--- a/sechub-server/src/main/resources/db/migration/common/V33__opqaue_token_cluster_cache.sql
+++ b/sechub-server/src/main/resources/db/migration/common/V33__opqaue_token_cluster_cache.sql
@@ -1,6 +1,7 @@
 -- SPDX-License-Identifier: MIT
 CREATE TABLE identity_oauth2_opaquetoken_cache
 (
+   id uuid not null,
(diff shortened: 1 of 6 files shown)


hamidonos (approved):
good job 🦫

